PR #8, opened 2019-01-23
Initialize Terraform with CloudFront CDN and S3 origin

### LINKED ISSUE #2: Provision S3 bucket and CloudFront distribution for the app
Use Terraform to provision an S3 bucket and CloudFront distribution to serve the app.

Ultimately, we want to hook the Mastarm build tool into these resources for deployments. For background, see the Mastarm docs: https://github.com/conveyal/mastarm#deploy.

jeancochrane:
## Overview

Perform initial Terraform setup, including a CloudFront CDN and S3 origin for the app. In addition, set up the foundation for the staging site at https://echolocator.azavea.com.

Closes #2.

## Demo

Visit https://echolocator.azavea.com to see a hello world message:

screen shot 2019-01-23 at 1 18 00 pm

## Testing instructions

* Create an `echo-locator` AWS profile using access credentials for the Professional Services account:

```
aws configure --profile echo-locator
```

* Run Terraform plan: 

```
docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra plan
```

* Confirm that the plan comes out clean
* Visit https://echolocator.azavea and confirm:
    * Page loads properly
    * Certificate valid
    * Content is the same as http://echo-locator-staging-site-us-east-1.s3.amazonaws.com/index.html

### DIFF
--- a/deployment/terraform/cdn.tf
+++ b/deployment/terraform/cdn.tf
@@ -5,8 +5,9 @@ resource "aws_cloudfront_distribution" "cdn" {
   }
 
   enabled             = true
+  is_ipv6_enabled     = true
   http_version        = "http2"
-  comment             = "EchoLocator (${var.environment})"
+  comment             = "${var.project} (${var.environment})"
   default_root_object = "index.html"
   retain_on_delete    = true
 
@@ -69,7 +70,12 @@ resource "aws_cloudfront_distribution" "cdn" {
 
   viewer_certificate {
     acm_certificate_arn      = "${module.cert.arn}"
-    minimum_protocol_version = "TLSv1"
+    minimum_protocol_version = "TLSv1.2_2018"
     ssl_support_method       = "sni-only"
+  }
+
+  tags {
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
   }
 }
--- a/deployment/terraform/cdn.tf
+++ b/deployment/terraform/cdn.tf
@@ -1,0 +1,75 @@
+resource "aws_cloudfront_distribution" "cdn" {
+  origin {
+    domain_name = "${module.origin.site_bucket}.s3.amazonaws.com"
+    origin_id   = "originEastId"
+  }
+
+  enabled             = true
+  http_version        = "http2"
+  comment             = "EchoLocator (${var.environment})"
+  default_root_object = "index.html"
+  retain_on_delete    = true
+
+  price_class = "PriceClass_100"
+  aliases     = ["${var.r53_public_hosted_zone_name}"]
+
+  default_cache_behavior {
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "originEastId"
+
+    forwarded_values {
+      query_string = true
+
+      cookies {
+        forward = "all"
+      }
+    }
+
+    compress               = true
+    viewer_protocol_policy = "redirect-to-https"
+
+    # Only applies if the origin adds Cache-Control headers. The
+    # CloudFront default is also 0.
+    min_ttl = 0
+
+    # Five minutes, and only applies when the origin DOES NOT
+    # supply Cache-Control headers.
+    default_ttl = 300
+
+    # One day, but only applies if the origin adds Cache-Control
+    # headers. The CloudFront default is 31536000 (one year).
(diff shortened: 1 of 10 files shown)


deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,75 @@
...
+
+  restrictions {
+    geo_restriction {
+      restriction_type = "none"
+    }
+  }
+
+  viewer_certificate {
+    acm_certificate_arn      = "${module.cert.arn}"
+    minimum_protocol_version = "TLSv1"
```
hectcastro: I think that we can update this to be `TLSv1.2_2018`.

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,75 @@
+resource "aws_cloudfront_distribution" "cdn" {
+  origin {
+    domain_name = "${module.origin.site_bucket}.s3.amazonaws.com"
+    origin_id   = "originEastId"
+  }
+
+  enabled             = true
```
hectcastro: Minor, but when we create CloudFront distributions via the UI, IPv6 is usually enabled. We can enable it here via `is_ipv6_enabled`.

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,75 @@
...
+      restriction_type = "none"
+    }
+  }
+
+  viewer_certificate {
+    acm_certificate_arn      = "${module.cert.arn}"
+    minimum_protocol_version = "TLSv1"
+    ssl_support_method       = "sni-only"
+  }
+}
```
hectcastro: It looks like this resource has a `tags` attribute now too. That would be helpful to add in the event that we need to do accounting based on project tags (if we stay in a shared AWS account like we are now).

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,75 @@
+resource "aws_cloudfront_distribution" "cdn" {
+  origin {
+    domain_name = "${module.origin.site_bucket}.s3.amazonaws.com"
+    origin_id   = "originEastId"
+  }
+
+  enabled             = true
+  http_version        = "http2"
+  comment             = "EchoLocator (${var.environment})"
```
hectcastro: Potentially useless nit, but is it `EchoLocator`? I don't know, but wanted to check.

deployment/terraform/dns.tf
```diff
@@ -0,0 +1,20 @@
...
+  tags {
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_route53_record" "site" {
+  zone_id = "${aws_route53_zone.external.id}"
+  name    = "${var.r53_public_hosted_zone_name}"
+  type    = "A"
```
hectcastro: If we enabled IPv6, we'd have to add an AAAA record here too.

deployment/terraform/storage.tf
```diff
@@ -0,0 +1,15 @@
+module "origin" {
+  source = "github.com/azavea/terraform-aws-s3-origin?ref=0.3.0"
```
hectcastro: Nice module reuse.

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,75 @@
...
+  default_root_object = "index.html"
+  retain_on_delete    = true
+
+  price_class = "PriceClass_100"
```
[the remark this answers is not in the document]
hectcastro: This makes sense to me.

deployment/terraform/cdn.tf
jeancochrane: Potentially unrelated, but I just realized that this should probably be `"${var.project} (${var.environment})"`.

jeancochrane:
Alright, this should be ready for another pass.

deployment/terraform/cdn.tf
ddohler: We've been using `ECHOLocator` or `ECHO Locator`.

jeancochrane: Ooh nice, thanks @ddohler!

hectcastro: Yeah, comment was purely about the spacing and casing. Going with one of the ones Derek suggests sounds good.

jeancochrane: Got it, should be updated now.

PR #10, opened 2019-01-25
Create dev environment and set up CI to deploy the app

### LINKED ISSUE #3: Configure Mastarm to deploy to S3/CloudFront resources
Configure the [Mastarm](https://github.com/conveyal/mastarm) build tool to bundle and push the app to S3, restarting the CloudFront distribution in the process.

This configuration should build on the existing configuration in https://github.com/conveyal/taui/. For more detailed docs on configuring Mastarm, see: https://github.com/conveyal/mastarm#configuration

Depends #2.

### LINKED ISSUE #4: Set up CI to bundle and deploy the app with Mastarm
Configure a CI pipeline to deploy the app using Mastarm, with deployment logic wrapped in an `infra` script.

Since this app is being developed in the open, we expect to use Travis for CI. If Travis proves an unacceptable choice, make a determination for a different build tool to use. Common alternatives include:
* Jenkins
* AWS CodeBuild

Depends #3.
Depends on #7.

jeancochrane:
## Overview

This PR sets up a simple dev environment and CI pipeline for the app.

Closes #3 and closes #4.

## Demo

View a simple deployment of the app at https://echolocator.azavea.com:

screen shot 2019-01-25 at 4 13 22 pm

## Testing instructions

* Follow the instructions in [the README](https://github.com/azavea/echo-locator/blob/feature/jfc/mastarm-deployment/README.md#development) for `Developing` and `Testing`, and confirm that they are accurate
* View the Travis output for the reverted commit that allowed a deployment to run from this branch, and confirm that Mastarm deployed properly: https://travis-ci.com/azavea/echo-locator/builds/98670744#L486

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -1,5 +1,9 @@
 version: '3'
 services:
+  taui:
+    environment:
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
   terraform:
     image: quay.io/azavea/terraform:0.11.11
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,17 +3,7 @@ services:
   taui:
     build: ./taui
     volumes:
-      # Mount app source files separately to avoid clobbering node_modules
       - ./taui/src:/usr/local/src/src
-      - ./taui/configurations:/usr/local/configurations/configurations
-      - ./taui/assets:/usr/local/assets/assets
-      - ./taui/flow-typed:/usr/local/flow-typed/flow-typed
-      - ./taui/package.json:/usr/local/src/package.json
-      - ./taui/yarn.lock:/usr/local/src/yarn.lock
-      - ./taui/config.json:/usr/local/src/config.json
-      - ./taui/__mocks__:/usr/local/src/__mocks__
-      - ./taui/.flowconfig:/usr/local/src/.flowconfig
-      - ./taui/index.html:/usr/local/src/index.html
     working_dir: /usr/local/src
     ports:
       - "9966:9966"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: '3'
+services:
+  taui:
+    build: ./taui
+    volumes:
+      # Mount app source files separately to avoid clobbering node_modules
+      - ./taui/src:/usr/local/src/src
+      - ./taui/configurations:/usr/local/configurations/configurations
+      - ./taui/assets:/usr/local/assets/assets
+      - ./taui/flow-typed:/usr/local/flow-typed/flow-typed
+      - ./taui/package.json:/usr/local/src/package.json
+      - ./taui/yarn.lock:/usr/local/src/yarn.lock
+      - ./taui/config.json:/usr/local/src/config.json
+      - ./taui/__mocks__:/usr/local/src/__mocks__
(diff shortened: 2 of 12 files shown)


docker-compose.ci.yml
```diff
@@ -1,5 +1,9 @@
 version: '3'
 services:
+  taui:
+    environment:
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
```
jeancochrane: Mastarm uses the AWS Node.js SDK under the hood to do deployments. We have to pass these variables into the deployment container in order to give it credentials.

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+      - ./taui/src:/usr/local/src/src
+      - ./taui/configurations:/usr/local/configurations/configurations
+      - ./taui/assets:/usr/local/assets/assets
+      - ./taui/flow-typed:/usr/local/flow-typed/flow-typed
+      - ./taui/package.json:/usr/local/src/package.json
+      - ./taui/yarn.lock:/usr/local/src/yarn.lock
+      - ./taui/config.json:/usr/local/src/config.json
+      - ./taui/__mocks__:/usr/local/src/__mocks__
+      - ./taui/.flowconfig:/usr/local/src/.flowconfig
+      - ./taui/index.html:/usr/local/src/index.html
```
jeancochrane: I tend to prefer keeping `node_modules` in the container image, rather than installing it locally. This wall of volume mounts is necessary to avoid clobbering the directory. If UA has a preferred way of managing this problem I'd be happy to change it to match your pattern.

scripts/cipublish
```diff
@@ -0,0 +1,39 @@
...
+    then
+        usage
+    else
+        TAUI_CONFIG_BUCKET="s3://echo-locator-${ENV}-config-us-east-1"
+        TAUI_CONFIG_DIR="configurations/${ENV}"
+        TAUI_SITE_BUCKET="s3://echo-locator-${ENV}-site-us-east-1"
+
+        # Pull down variables for this environment
+        aws s3 cp "${TAUI_CONFIG_BUCKET}/taui/env.yml" \
+                  "taui/${TAUI_CONFIG_DIR}/env.yml"
```
jeancochrane: `env.yml` is the only Mastarm configuration file that contains secrets. Because of this, I figured the safest way to handle it was to keep versions for each environment in the S3 remote state bucket and pull it down during the CI build.

I think it would make sense to have one of these for development and do a similar thing in an `update` script. I didn't do that in this PR since I didn't really plan to invest too much time in the dev environment.

docker-compose.yml
flibbertigibbet: This looks fragile. I'd rather not have us need to maintain a list of source files. I believe our projects generally sync `node_modules` with the host file system.

scripts/cipublish
flibbertigibbet: Could we add a local `env.yml` example file and instructions to the `README` to copy it over and set secrets for development setup here instead?

jeancochrane: Yup, that works! I added some very basic instructions to the README. You may need to adjust these instructions depending on what winds up being configured in `env.yml`.

docker-compose.yml
jeancochrane: Perfectly fine. I removed the file-specific mounts and just mounted `taui/src/` instead.

PR #12, opened 2019-01-28
Feature/kak/default to boston

flibbertigibbet:
## Overview

Default to center on Boston's City Hall (instead of Japan) and move the R5/analysis URL definitions to `store.yml`. With tile URLs and Mapbox access token set in `env.yml`, allows map to load, and will attempt to load analysis results.

## Demo

On initial load, it appears to override the default location from `store.yml` based on `config.json` contents:
image

With a destination set that has an available route:
image

## Notes

It looks like the analysis files on S3 need to have CORS enabled for the grid interactivity tiles to load, and also it looks like the analysis path results haven't been loaded yet on S3 (requests made for them on changing origin/destination return 404s).

## Testing

 - set up local development Mapbox tiles configuration in `taui/configurations/default/env.yml` (see discussion for values)
 - `./scripts/serve`
 - Map should load tiles
 - Sidebar should indicate network configuration was found (show late night and late afternoon sections)


flibbertigibbet:
Trivial configuration changes, merging

Issue #15, opened 2019-01-28
Deploy minified build to production


flibbertigibbet:
When deploying to production (and staging?), use minified build, not the development build. See [this documentation](https://reactjs.org/docs/optimizing-performance.html#use-the-production-build).

flibbertigibbet:
Minified React builds are for production and not staging.

PR #16, opened 2019-01-28
Pass environment on deploy to mastarm

flibbertigibbet:
Not sure if this will address #15, but `mastarm` uses the environment variable.

### DIFF
--- a/scripts/cipublish
+++ b/scripts/cipublish
@@ -34,6 +34,6 @@ then
         # Deploy CSS and JS with Mastarm
         docker-compose -f docker-compose.yml \
                        -f docker-compose.ci.yml \
-                       run --rm taui run deploy --minify --config "${TAUI_CONFIG_DIR}"
+                       run --rm taui run deploy --minify --config "${TAUI_CONFIG_DIR}" --env "${ENV}"
     fi
 fi


scripts/cipublish
```diff
@@ -34,6 +34,6 @@ then
         # Deploy CSS and JS with Mastarm
         docker-compose -f docker-compose.yml \
                        -f docker-compose.ci.yml \
-                       run --rm taui run deploy --minify --config "${TAUI_CONFIG_DIR}"
+                       run --rm taui run deploy --minify --config "${TAUI_CONFIG_DIR}" --env "${ENV}"
```
jeancochrane: What change do you expect this to make? My understanding was that by passing the environment config directory to `--config` we'd be setting the environment already (since `cipublish` sets `TAUI_CONFIG_DIR` based on the `ENV` variable -- e.g. `TAUI_CONFIG_DIR` is `configurations/staging` in staging). Note that the staging `settings.yml` file also includes the attribute `env: staging`, although I don't know if that's intended to do the same thing as the `--env` flag.

Keep in mind I really don't have a good handle on what either the `--env` flag or the `env` attribute of `settings.yml` is supposed to do in mastarm, so you may have a better sense than I do of the change that this makes.

flibbertigibbet: I'm not sure what passing this explicitly might change either, and wanted to try it on staging to find out. Do you have an idea why the staging deployment isn't minified, although the minify flag is passed here?

jeancochrane: I don't quite understand what the issue is with minification -- the staging JS looks minified to me:

screen shot 2019-01-29 at 9 59 01 am

Is the problem something specific to the React development build? That's what the docs link in #15 seems to indicate to me.

flibbertigibbet: Yes, the React build. https://taui.conveyal.com reports it's using a production build of React, but our staging deployment does not.

jeancochrane: Perhaps it's because `env` is `staging`? The docs for [production React builds in Browserify](https://reactjs.org/docs/optimizing-performance.html#browserify) indicate that `NODE_ENV` needs to be set to `production` during the build, and it seems like [mastarm sets this using `process.env`](https://github.com/conveyal/mastarm/blob/4acc0f174bf52f379099e8d515c91172ba70f9b8/lib/js-transform.js#L12-L27):

```js
module.exports = function transform ({ config, env, instrument }) {
  const envvars = Object.assign(
    {},
    process.env,
    {
      BUILD_TIMESTAMP: (new Date()).getTime(),
      COMMIT_SHA: commit,
      CONFIG_PATH: config.path,
      MESSAGES: JSON.stringify(config.messages),
      NODE_ENV: env,
      REPO_URL: pkg.repository && pkg.repository.url && pkg.repository.url.replace('.git', ''),
      SETTINGS: JSON.stringify(config.settings),
      STORE: JSON.stringify(config.store)
    },
    config.env
  )
```

This might be hard to test in CI since `ENV` also controls the location of the remote state. How about I try running a staging deployment with `--env production` locally and we can see if it works?

flibbertigibbet: Oh, good find! That's not necessary, but it might be helpful to test the production deployment if it's not too much effort.

PR #37, opened 2019-02-12
Pin package versions and update mastarm

flibbertigibbet:
Pin package versions from npm.
Update mastarm fork to fix staging deployment graphql reference.


flibbertigibbet:
Trivial; merging for staging deploy.

flibbertigibbet:
Update to `mastarm` fork to [not mangle when using UglifyJS](https://github.com/graphql/graphql-js/issues/1182) should fix this staging error:

image

PR #38, opened 2019-02-13
Feature/kak/neighborhood data#19

### LINKED ISSUE #19: Add neighborhood data
Extract neighborhood data from spreadsheet, transform it to JSON, and add the 104 which are the zipcodes to use in this iteration of the project (those with `ecc=1` in the spreadsheet?) as a static JSON file.

Depends on #29.

### LINKED ISSUE #30: Overlay zip code areas on map
Using the same [Census ZCTA data](https://www.census.gov/geo/maps-data/data/cbf/cbf_zcta.html) as in #29, add the bounds for the zip codes for this app (those in #19) as a GeoJSON file to the project (either as a static front-end file, or retrieved from S3).

Load the GeoJSON as a map layer.

Depends on #24.
Depends on #19.

flibbertigibbet:
## Overview

Serve and display neighborhood points and bounds on the map.

Adds a script to generate the neighborhood points GeoJSON from the CSV with centroids generated by #36.

## Demo

The 'town' name label displays on hover over the centroid points. The point styling was taken from the unused "points of interest" feature of taui.

image

React console logs with the added data loading steps:

echo_react_neighborhood_loading_logs

## Notes

There are 102 neighborhoods loaded here, rather than the 104 ECC neighborhoods in the source CSV, due to the two zip codes not having their own ZCTA, but belonging to another (see notes on #36).

## Testing

 - `./scripts/server`
 - Go to the [map page](http://localhost:9966/map)
 - ECC neighborhood points and bounds should display properly on the map
 - Point markers should have hover (name label) and click (just a console log for now) interactivity

Closes #19 
Closes #30


taui/src/actions/json-data.js
```diff
@@ -3,11 +3,11 @@ import fetch from '@conveyal/woonerf/fetch'
 
 import cacheURL from '../utils/cache-url'
 
-export function loadPointsOfInterest (url: string) {
+export function loadDataFromJSON (url: string, type: string) {
```
flibbertigibbet: This is an [action creator](https://redux.js.org/basics/actions#action-creators). The data reducer matches on the `type` label.

PR #39, opened 2019-02-14
Persist Amplify resources and configure production environment

### LINKED ISSUE #6: Persist (or configure) Cognito auth resources
Per https://github.com/azavea/echo-locator/pull/1, we'll be using AWS Cognito for user authentication in the app. If app developers wind up configuring these resources using Amplify.js during development, codify them in Terraform configurition; otherwise, use Terraform to provision them from the start.

Auth resources will likely include at minimum:

- S3 bucket for storing user metadata
    - Related policies for restricting access to buckets to the relevant users and admins
- Cognito User Pool
- Cognito Identity Pool

See:
* https://aws-amplify.github.io/docs/js/authentication

jeancochrane:
## Overview

Move Amplify resources into a dedicated deployment directory, `deployment/amplify`, and create a production environment with Amplify CLI.

In addition, update scripts and docs to reflect the altered Amplify deployment workflow.

Closes #6. 

## Notes

* This PR assumes we want to support staging and production Amplify configurations. Potentially, it may be simpler overall to just drop the staging setup altogether, and route everything through the production instance. Curious what others think about this possibility.

## Testing

1. To test the Amplify deployments, you'll need admin credentials in the BHA AWS account. Use the Admin user in LastPass to set yourself up with credentials, and configure those credentials on the command line:

```
$ aws configure --profile echolocator
```

2. Edit a file in the production stack so that you can test changes, e.g.:

```diff
--- a/deployment/amplify/production/amplify/backend/auth/cognitoEcholocatorProduction/parameters.json
+++ b/deployment/amplify/production/amplify/backend/auth/cognitoEcholocatorProduction/parameters.json
@@ -1,6 +1,6 @@
 {
     "identityPoolName": "echolocatorIdentityPoolProduction",
-    "allowUnauthenticatedIdentities": false,
+    "allowUnauthenticatedIdentities": true,
     "thirdPartyAuth": false,
     "lambdaLogPolicy": "cognitoEcholocatorProduction_lambda_log_policy",
     "openIdLambdaRoleName": "cognitoEcholocatorProduction_openid_lambda_role",
``` 
3. Run a container with Amplify CLI:

```
$ docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm amplify
```

4. Change into the production Amplify directory with `cd production`
5. Run `amplify push auth` and confirm that the CloudFormation stack updates with your changes
6. Undo the changes you made in 2) above and rerun `amplify push auth` to clean up your changes


deployment/amplify/production/amplify/.config/project-config.json
```diff
@@ -0,0 +1,20 @@
...
+    "projectName": "src",
+    "projectPath": "/usr/local/src/production/",
+    "defaultEditor": "sublime",
+    "frontendHandler": {
+        "javascript": "/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-frontend-javascript"
+    },
+    "javascript": {
+        "framework": "react",
+        "config": {
+            "SourceDir": ".",
```
jeancochrane: This is a somewhat hacky way to get the `aws-exports.js` file (which the frontend code needs in order to ping AWS resources) to write to the `deployment/amplify/${env}` directory. The Amplify CLI will follow this path to write `aws-exports.js`, whether or not it's the true source directory.

deployment/amplify/staging/amplify/.config/aws-info.json
```diff
@@ -0,0 +1,5 @@
+{
+    "configLevel": "project",
+    "useProfile": true,
+    "profileName": "echolocator"
```
jeancochrane: The Amplify CLI uses this attribute to source the AWS profile that you want to use. Not great, since you'll have to edit this file manually if your profile is named something different (in this case, the staging profile should correspond to the Azavea root AWS account, for instance).

deployment/amplify/staging/amplify/.config/project-config.json
```diff
@@ -1,14 +1,14 @@
 {
```
jeancochrane: As you can see from the filename change, `taui/src/amplify` got moved to `deployment/amplify/staging`, while `deployment/amplify/production` stores the production deployment.

scripts/cipublish
```diff
@@ -31,6 +31,9 @@ then
         # Mastarm doesn't copy over index.html
         aws s3 cp taui/index.html "${TAUI_SITE_BUCKET}/index.html"
 
+        # Move the Amplify config for the proper env into the source dir
+        cp "deployment/amplify/${env}/aws-exports.js" taui/src/aws-exports.js
```
jeancochrane: This is necessary to bundle the Amplify exports into the app for the appropriate environment.

jeancochrane:
Apologies for the extremely gross diff @hectcastro. I tried my best to provide guideposts throughout. If it's too confusing, let me know if you want to meet up for a ~15 min sync and I can try to explain some of the design decisions in more detail.

jeancochrane:
Great questions @hectcastro -- I'll respond to comments on the threads, but I wanted to touch on some of the questions you asked in your review body too:

> What does "route everything through the production instance" imply here? Use all of the production backend resources via the staging site?

Specifically I meant to say that we might consider using production authentication resources on the Staging site -- effectively that we only have one identity provider for both staging and production. The downsides to that choice seem big and obvious, but possibly mitigated if we're planning on tearing down the staging site once this goes live anyway (not sure if that was the plan, I only remember hearing murmurs of it).

> It seems like we're breaking out of the multiple environment process outlined [here](https://aws-amplify.github.io/docs/cli/multienv?sdk=js). [...] Did you come to that approach on your own, or was that influenced by another project using Amplify?

Unfortunately, this was an approach I hacked together on my own. The multi-environment model is only supported in a beta version of the CLI; in addition to being experimental, it's based on a substantially different version of the CLI than the one we used to set up these resources initially (since https://github.com/azavea/echo-locator/pull/34 came in, there has already been a new major version release of Amplify CLI, `1.x`, which introduces breaking changes).

My reasoning for saving two separate stacks in version control was that it seemed like the fastest way to build on top of our existing work, but I think it definitely introduces huge maintainability problems. I wish there had been a good example I could have bootstrapped off of, but resources on how to use Amplify CLI are hard to find, and as far as I can tell they're exclusively tied to the new beta version of the CLI.

deployment/amplify/production/amplify/#current-cloud-backend/amplify-meta.json
```diff
@@ -0,0 +1,51 @@
...
+		"cognitoEcholocatorProduction": {
+			"service": "Cognito",
+			"providerPlugin": "awscloudformation",
+			"providerMetadata": {
+				"s3TemplateURL": "https://s3.amazonaws.com/src-20190214172616-deployment/amplify-cfn-templates/auth/cognitoEcholocatorProduction-cloudformation-template.yml",
+				"logicalId": "authcognitoEcholocatorProduction"
+			},
+			"lastPushTimeStamp": "2019-02-14T18:57:13.661Z",
+			"output": {
+				"AppClientSecret": "begdu9bclerkc860lbs10gfs3b3gufdta028vqoqo6eashh1qs8",
```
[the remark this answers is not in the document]
jeancochrane: This also made me nervous when I first saw it, but as far as @flibbertigibbet and I could tell, it's only used in mobile apps. Relevant discussion: https://github.com/azavea/echo-locator/pull/34#discussion_r255195141

deployment/amplify/production/amplify/#current-cloud-backend/auth/cognitoEcholocatorProduction/cognitoEcholocatorProduction-cloudformation-template.yml
```diff
@@ -0,0 +1,395 @@
...
+              - 'sts:AssumeRole'
+    DependsOn: UserPoolClient
+  UserPoolClientLambda:
+  # Lambda which gets userpool app client config values
+  # Depends on UserPool for id
+  # Depends on UserPoolClientRole for role ARN
+    Type: 'AWS::Lambda::Function'
+    Properties:
+      Code:
+        ZipFile: !Join
```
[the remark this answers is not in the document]
jeancochrane: I know 😅 I decided I wasn't going to even try recreating these resources in Terraform when I saw this.

flibbertigibbet:
The production configuration refuses to allow signing up with a phone number (gives the error "A client attempted to write unauthorized attribute"), and always presents the email verification screen, unlike the staging configuration. I'm not sure how exactly they differ.

deployment/amplify/staging/amplify/.config/aws-info.json
flibbertigibbet: The deployment README says to configure a profile named 'echo-locator', with a hyphen.

deployment/amplify/production/amplify/#current-cloud-backend/amplify-meta.json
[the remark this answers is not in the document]
hectcastro: I looked into the IP address restriction stuff for another project and it isn't straightforward to automate.

For pure S3 it is (you can bake IP whitelists into the bucket policy). For CloudFront, you can either layer in the AWS Web Application Firewall (WAF), or use Lambda@Edge with bits of Node.js functions to drop requests that match remote host IPs.

The WAF approach _could_ be a bit easier if we applied it manually, but if we can disable account creation in Cognito by tweaking settings in the existing CloudFormation YAML, that approach seems best.

hectcastro:
> Specifically I meant to say that we might consider using production authentication resources on the Staging site -- effectively that we only have one identity provider for both staging and production. The downsides to that choice seem big and obvious, but possibly mitigated if we're planning on tearing down the staging site once this goes live anyway (not sure if that was the plan, I only remember hearing murmurs of it).

I would be OK with doing this if we think we can get away with doing so. I think @flibbertigibbet is in a better position to weigh-in on our ability to do so.

flibbertigibbet:
I just pushed a commit for a production deploy I made, in which I removed the `WriteAttributes` setting missing from the staging configuration. This fixes the error with attempting to set a phone number on account creation, but has not fixed the email verification post-login messages.

deployment/README.md
```diff
@@ -51,3 +52,43 @@ output the IDs of both resources. Take these IDs and put them in the
...
+User authentication resources are provisioned using the [AWS Amplify
+CLI](https://aws-amplify.github.io/docs/cli/concept). We don't anticipate
+that these resources will change frequently (if at all), but if you'd like
+to update existing categories or add new categories to the stack, use the
+following instructions.
+
+Run a container with the Amplify CLI installed:
+
+```
+$ docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm amplify
```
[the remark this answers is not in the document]
jeancochrane: Good catch! I'll resurrect `./scripts/amplify-cli` for this purpose.

deployment/amplify/staging/amplify/.config/aws-info.json
jeancochrane: > Does this mean it will be necessary to edit the host AWS credentials to switch accounts/environments?

I'm imagining that if your profile name doesn't match what's in the `aws-info.json` config for the environment you're trying to update, you'll either need to rename your profile or edit the local copy of this config to match your profile name. I assume the latter (editing this config file) will be less obtrusive.

> The deployment README says to configure a profile named 'echo-locator', with a hyphen.

Seems like there's some inconsistency here, since the main repo uses `echolocator` (see https://github.com/azavea/echo-locator/blob/1c3b7da560546e08cb1f05cfc570587c42a04223/README.md#development). Another inconsistency I just remembered is that the staging stack is actually deployed to the Azavea root account. So I think our two options here are either:

1. Redeploy the staging stack in the BHA account and update all `aws-info.js` profile names to `echo-locator` (would remove any existing staging users/data, would be slower, but would be the cleanest setup)
2. Keep the staging stack in the Azavea root account and update this value to be e.g. `azavea`

Leaning towards 2) above, since it's the fastest move.

jeancochrane:
I don't see that commit @flibbertigibbet -- if you push it up I can take a look at pushing a fix to the extraneous email verification messages.

flibbertigibbet:
Regarding sharing staging/production authentication and S3 resources, I think the benefits to keeping them separate make it worth continuing to do so, particularly for the S3 assets. We will want to have separate data and users on staging, to non-destructively test deployed changes.

flibbertigibbet:
I just force-pushed, which dropped @hectcastro's commit.

jeancochrane:
Made updates based on comments. Thanks for the detailed reviews! I wasn't able to find the CloudFormation attribute that controls the email verification screen (it's possible it's a frontend issue, as well), but I did turn off user signup on the production instance. Going to pull this in now.

flibbertigibbet:
As the email verification issue is limited to the production configuration, it's not a frontend issue.

PR #40, opened 2019-02-15
Feature/kak/amplify theme#35

### LINKED ISSUE #23: Implement logout
Implement logout button displayed on site.

Depends on #21.

### LINKED ISSUE #35: Customize Amplify UI theme
Style login page by [customizing the Amplify theme](https://aws-amplify.github.io/docs/js/authentication#customize-ui-theme) or modifying the components.

flibbertigibbet:
## Overview

Customizes sign-in page and adds a customized navbar with a logout button.

## Demo

echo_custom_signin

bha_splash_page

image

## Notes

This will need some styling cleanup (i.e., the existing taui blue probably isn't the BHA blue we want), but sets up the customized sign-in and header bar components alongside a custom Amplify theme that readies the app for further styling work and header bar customization.

The logged-in user name is currently displayed in the header bar; as part of implementing account selection/switching, that should be changed to match the current account, as in the wireframes.

The Amplify theme styling is currently defined near the top of `index.js`, defined as `CustomAuthenticatorTheme`. Ideally, these rules would go in a CSS file instead. I'm not sure how best to handle that.

## Testing

 - `./scripts/serve`
 - Clear local storage for http://localhost:9966
 - Should get customized login page
 - On login, should have customized header and be able to use it to log out

Closes #35 
Closes #23

### DIFF
--- a/taui/src/index.js
+++ b/taui/src/index.js
@@ -7,16 +7,38 @@ import React from 'react'
 import { connect } from 'react-redux'
 import { BrowserRouter, withRouter } from 'react-router-dom'
 import {
+  ConfirmSignIn,
+  ConfirmSignUp,
+  ForgotPassword,
+  Loading,
+  RequireNewPassword,
   SignIn,
-  withAuthenticator
+  TOTPSetup,
+  VerifyContact
 } from 'aws-amplify-react'
 
 import actions from './actions'
 import awsmobile from './aws-exports'
 import Application from './components/application'
 import CustomSignIn from './components/custom-sign-in'
+import withAuthenticator from './components/with-authenticator'
 import reducers from './reducers'
 import * as select from './selectors'
+
+// custom Amplify component styling
+const CustomAuthenticatorTheme = {
+  formContainer: {'margin': '0'},
+  sectionFooter: {'display': 'block'},
+  button: {'backgroundColor': 'var(--blue)', 'width': '100%'},
+  nav: {
+    'display': 'flex',
+    'justifyContent': 'space-between',
+    'backgroundColor': 'lightgrey',
+    'padding': 0},
+  navBar: {'border': 0},
+  navItem: {'lineHeight': '300%'},
+  navButton: {'backgroundColor': 'var(--blue)'}
+}
 
 // Set the title
 document.title = message('Title')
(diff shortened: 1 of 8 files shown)


taui/src/index.js
```diff
@@ -6,14 +6,31 @@ import get from 'lodash/get'
...
 import actions from './actions'
 import awsmobile from './aws-exports'
 import Application from './components/application'
+import CustomSignIn from './components/custom-sign-in'
+import withAuthenticator from './components/with-authenticator'
 import reducers from './reducers'
 import * as select from './selectors'
 
+// custom Amplify component styling
+const CustomAuthenticatorTheme = {
```
ddohler: I dug into this a bit to try and understand the problem better and I don't think I have much to add. You've probably looked at this stuff already, but for @lederer 's benefit (and potential future readers), it looks like Amplify has [rolled their own CSS-in-JS solution](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Components-React.jsx#L10) which they use pretty extensively, and it seems like that will make it awkward to put styling anywhere else as long as we're inheriting from Amplify. The two options I can think of quickly are to either have a single JS file for storing styles, which is what it looks like [Amplify does](https://github.com/aws-amplify/amplify-js/blob/master/packages/aws-amplify-react/src/Amplify-UI/Amplify-UI-Theme-Sample.jsx), or to try to add in some of our own classes that we can use for styling. @flibbertigibbet does that sound right to you? Are there any other options you've thought of?

@lederer do you have any ideas or opinions?

flibbertigibbet: The separate style JS file seems like a good compromise to me.

flibbertigibbet:
Other components apparently get hidden [here](https://github.com/aws-amplify/amplify-js/blob/01eedefe9a406cc19b62e9728fd8279714f48a4e/packages/aws-amplify-react/src/Auth/Authenticator.jsx#L156) when any one of them is overridden; I've modified this to list them all explicitly.

image

taui/src/index.js
lederer: Interesting. 

The _add in our own classes_ approach would fit the design team's usual practice of authoring [componentized SASS](https://github.com/azavea/facade/tree/develop/src/sass) that the build system transforms into a single `main.css` output file. 

Whereas, IIUC, the _JS theme file_ approach would forego all the value we get from SASS and autoprefixer, etc, and require authoring the equivalent of a `main.css` file directly (including prefixing, etc). I'm open to that if this is a small, one-off site. But if we expect it to grow, I'd prefer a scalable SASS-based solution.

Could there be a middle ground where we can transform a `main.css` output file into an Amplify theme file?

flibbertigibbet: Most of the styles are in a `main.css`. What we're discussing here are only the styles that override those that come from Amplify, which are the authentication components we're reusing (the sign-in page and header bar, primarily).

flibbertigibbet: There is a single `main.css` file for most of the styles. This is just to customize aspects of the authentication components (the sign-in and header bar, primarily), not an overall theme. I don't believe attempting to generate the Amplify JS theme overrides from CSS would be practical.

lederer: Ah ok. Whew. 😅 

Customizing a theme file sounds fine for that.

flibbertigibbet: There's a section of the Cognito AWS console page for manually configuring per-domain/per user pool authentication component style overrides. I'm guessing this is related to why they've got JS-defined styles as they do, to support [that feature](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-app-ui-customization.html).

(I don't think remotely-managed styles would be useful for our purposes here, but it's interesting to know this feature exists.)

image

flibbertigibbet:
I moved the theme override to `taui/src/amplify-theme.js`.

PR #47, opened 2019-02-26
Modularize imports

flibbertigibbet:
## Overview

- Modularize Amplify imports to reduce bundle size and allow for updating the react components, by working around automatic import of broken library
- Manually pin amazon-cognito-auth-js transitive dependency version in `yarn.lock`, as 1.3.0 release breaks build
- Update README for instructions to run directly (outside Docker)


flibbertigibbet:
Merging to fix staging deploy.

PR #53, opened 2019-02-27
Stand up new production Amplify stack to fix email verification

### LINKED ISSUE #42: Fix bug where users have to verify email addresses on every login
In the production instance, for some reason users are prompted to verify their email address every time they sign in. Once an address is verified, users shouldn't have to verify it any more.

I spent a total of about an hour messing around with every possible combination of Cognito settings and I couldn't figure out what was going on. Opened up https://github.com/aws-amplify/amplify-js/issues/2730 to seek guidance from the Amplify developers.

jeancochrane:
## Overview

Fix the bug where users have to perpetually re-verify their email address by standing up a new production Amplify stack with the correct parameters selected.

Closes #42.

## Notes

The diff here is pretty huge, since I renamed the production stack. I'll try to call out any relevant portions of it that relate to things we changed, but in general, I don't expect reviewers to pore over the details of the configuration here. Testing is more important.

## Testing instructions

- Move the production `aws-exports.js` config file into the Taui source directory:

```
cp deployment/amplify/production/aws-exports.js taui/src/aws-exports.js
```

- Start up a development server
  - You can use `./scripts/server` to bring up Docker Compose services, or `npm start` if you're using a local Node installation
- Using the admin user credentials supplied in the `Cognito admin user (production)` LastPass entry, log into http://localhost:9966 and confirm that you don't have to verify your email address


deployment/amplify/production/amplify/#current-cloud-backend/auth/echolocatorCognitoProduction/parameters.json
```diff
@@ -0,0 +1,66 @@
...
+        "email"
+    ],
+    "userpoolClientName": "echolocatorCognitoProduction_app_client",
+    "userpoolClientGenerateSecret": true,
+    "userpoolClientRefreshTokenValidity": "90",
+    "userpoolClientReadAttributes": [
+        "email",
+        "phone_number",
+        "updated_at",
+        "email_verified",
```
jeancochrane: This ended up being the culprit -- Cognito wasn't allowing end users to read the `email_verified` attribute.

flibbertigibbet: Curiously, this was set to only "email" for both of the previous configurations. Maybe something changed with the CloudFormation rules for this service.

deployment/amplify/production/amplify/#current-cloud-backend/auth/echolocatorCognitoProduction/parameters.json
```diff
@@ -0,0 +1,66 @@
...
+    "mfaPassRolePolicy": "echolocatorCognitoProduction_totp_pass_role_policy",
+    "mfaLambdaIAMPolicy": "echolocatorCognitoProduction_totp_lambda_iam_policy",
+    "userpoolClientLambdaRole": "echolocatorCognitoProduction_userpoolclient_lambda_role",
+    "userpoolClientLogPolicy": "echolocatorCognitoProduction_userpoolclient_lambda_log_policy",
+    "userpoolClientLambdaPolicy": "echolocatorCognitoProduction_userpoolclient_lambda_iam_policy",
+    "userpoolClientSetAttributes": true,
+    "useDefault": "manual",
+    "authSelections": "identityPoolAndUserPool",
+    "resourceName": "echolocatorCognitoProduction",
+    "userpoolClientWriteAttributes": [
```
flibbertigibbet: These were defined in the previous production configuration, but undefined in the staging configuration.

jeancochrane: Good catch! Do you think it's worth aligning this with the staging stack?

deployment/amplify/production/amplify/#current-cloud-backend/auth/echolocatorCognitoProduction/parameters.json
jeancochrane: That's a good point. It's possible that we adjusted something out of band without knowing it, too.

deployment/amplify/production/amplify/#current-cloud-backend/auth/echolocatorCognitoProduction/parameters.json
flibbertigibbet: No, staging is working fine, so I think it would be better to leave it. If the issue should appear on staging later, we can revisit it then.

Issue #51, opened 2019-02-26
Limit Cognito S3 bucket access to current user pool


flibbertigibbet:
Currently, the Amplify-generated S3 `Storage` permissions for the [preconfigured access level](https://aws-amplify.github.io/docs/js/storage#file-access-levels) 'public' objects are accessible to any authenticated AWS user, regardless of which account.

For this project phase, logged-in Cognito users should be able to create profiles (stored as S3 files, named for the account) and be able to access any profile on the site. No other access to the S3 bucket should be allowed, apart from the logged-in Cognito user role (and the development/deployment role for the account).

To support this, modify the S3 permissions in the Amplify-generated [CloudFormation configurations](https://github.com/azavea/echo-locator/blob/develop/deployment/amplify/staging/amplify/backend/storage/echolocatorContent/s3-cloudformation-template.json) to restrict access to the `public` directory to authenticated Cognito users in the current user pool (who share an AWS role). Logged-in users of other Cognito user pools or CLI users accessing via other account credentials should be denied access to the S3 Amplify data bucket. (No other AWS roles should have access, as they do now.)

The pre-configured `protected` and `private` access levels for Amplify storage rely on paths scoped to the Cognito user ID, and so would not be as useful for the needs of this project, as we will want to get account profiles by S3 key (file name) stored flat in a shared directory.

Issue #22, opened 2019-02-04
Implement profile page


flibbertigibbet:
Display account profile details fetched from S3 after account selected. Write settings back to S3 on user clicking 'Go' button.

![image](https://user-images.githubusercontent.com/960264/52225613-ce124080-2878-11e9-8d0a-2f8f92fec41b.png)

![image](https://user-images.githubusercontent.com/960264/52225985-c3a47680-2879-11e9-8d2f-4961f6dc2bc2.png)

Depends on #21.

flibbertigibbet:
@lederer what should 'Cancel' do here?

lederer:
> what should 'Cancel' do here?

If the user lands here from the "New Search" page, Cancel should return to that page (equivalent to clicking "New" in the global nav). 

If they land here after clicking Edit in the global nav of the search or detail page, then Cancel should abandon any edits to this form and return to the page they came from.

Issue #31, opened 2019-02-04
Implement 'edit' and 'new' buttons


flibbertigibbet:
Implement 'edit' and 'new' buttons, at top of map view:
![image](https://user-images.githubusercontent.com/960264/52228216-c86c2900-287f-11e9-94f3-51ce28984150.png)

'Edit' should redirect to the profile page; on clicking 'cancel' from profile page, it should redirect back to map page, unchanged. See #22.

'New' should redirect to account search/selection page, and unset the currently selected account.

Depends on #24.

PR #55, opened 2019-02-28
Feature/kak/account search#21

### LINKED ISSUE #21: Implement account search and selection
Implement account search and selection page, encountered after login.

image

Search S3 contents for settings matching a given voucher number and/or name and display list of matching results. Allow creating a new settings store, in case no settings exist yet for the account.

Determine if any validation rules can be added. (Could there be, for example, a Lamba check against a list of valid names/voucher numbers before allowing a new settings file be created?)

On selection or creation, retrieve the settings for the account from S3.

Depends on #20.

flibbertigibbet:
## Overview

Adds an unstyled page at `/select` to search for, select, create, and delete account profiles for voucher holders. Redirect to the page on site load if no profile set. Store profile JSON in browser localStorage and in a file on S3.

Also adds a pull request template and updates the `mastarm` build tools version in the lock file.

## Demo

Search with no parameters returns all:
image

On creating a new profile:
image

Navigates to map page on profile selection:
image

## Notes

After this, #22 should only require implementing the form and adding the page; the data management parts should be in place.

Currently, any authenticated AWS user can access the profile files; #51 is to restrict it to the Cognito-associated IAM role.

The wireframe for the select page doesn't offer a way to create a new profile, so I added a button to the search form for that. There's also not a delete option in the wireframes, so I made that a list item button. We may want to revisit where those buttons go and how the profile selection flow works.

## Testing

 - `cd taui && yarn start`
 - After login, should see 'select account' page
 - Clicking 'search' with empty fields should show all
 - Should be able to create an account (after entering any text into the two fields)
 - Should see an error message on attempting to create an account with either field empty
 - Should be able to search by fields
 - Should be able to delete account via list button
 - Select an account
 - Should redirect to map page
 - Should show selected voucher holder name and number in the header
 - Header info should persist through router page changes and page refresh
 - Logging out and back in with an account already selected should go to map page directly
 - On map page, delete account profile from localStorage (key `echolocator-profile`)
 - Refresh page; should redirect to select account with no account in header info

Closes #21


lederer:
I've been operating under the assumption that the app would only perform individual lookups, but listing all accounts in response to an empty search is interesting. It could be useful for BHA staff, though would need to be protected if the app were ever made available to the public. @jcahail Thoughts on including this functionality?

> The wireframe for the select page doesn't offer a way to create a new profile, so I added a button to the search form for that. There's also not a delete option in the wireframes, so I made that a list item button. We may want to revisit where those buttons go and how the profile selection flow works.

Good point. Given the way we're handling "accounts", I think we should make a slight adjustment to the search flow:

- Search page starts with a single field for "Voucher Number". No name field.
- On submit, if the entered number is found, immediately go to that profile page (instead of listing matching accounts).
- If the entered number is not found, render a second field for "Name of Head of Household" under the "Voucher Number" field, a "Create profile" button, and a "Cancel" button (which resets the page back to a single Voucher Number field).
- Delete button can live on the profile page itself.

I don't know if these are in scope for this PR or should be a follow-up.

flibbertigibbet:
Non-staff users, when we add them, will only be able to access their own profiles, and so will have no need for search.

Do we want to require counselors to enter the full voucher number every time? Partial matching and name search seem useful for that purpose. Also confirming account holder name before selecting the profile, in case the number was entered incorrectly.

jcahail:
From BHA re: voucher numbers: "Depending on when they were created, voucher numbers are either number or alphanumeric. Examples could be 669854 or AG213546. They are the unique identifiers used to lookup a family in our system of record." So it sounds like current practice is for staff to enter the full voucher number.  Given our current budget restraints we need to go with the least  complicated option that allows BHA counselors to look up voucher holders and either enter their profile or go back to look up a new one.

Listing all accounts if the search field is empty would have to be restricted to BHA staffers in a subsequent contract; provided (1) we can add that role/permission in easily at a later stage and (2) adding the list all feature here is very low effort, I'm in favor of adding it.

flibbertigibbet:
Only BHA staffers should have access to the search functionality at all, as I understand it.

Thanks for the example voucher numbers. Are there any rules we could apply for validation? (i.e., if there are characters, they are at the start, and voucher numbers are six or eight characters.) It's hard to judge from the examples.

taui/src/components/select-account.js
```diff
@@ -0,0 +1,208 @@
...
+      console.error('Missing name or voucher')
+      this.setState({errorMessage:
+        'Enter both name and voucher to create account.'})
+      return
+    } else {
+      this.setState({errorMessage: ''})
+    }
+
+    const key = name.toUpperCase() + '_' + voucher.toUpperCase()
+    console.log('Creating account profile for ' + key)
```
ddohler: There are a few `console.log` statements in this file that look like they're mostly for debugging purposes. If you're planning to use them in future work I think it's fine for them to stick around for the moment but I wanted to flag them to make sure we know they exist and will need to be removed eventually.

flibbertigibbet: I made a pass to clear out extraneous console logging outside of this file, but yes, there's some partially-implementing things here I intend to clear of logging with the next task.

lederer:
> Given our current budget restraints we need to go with the least complicated option…

I assume the least complicated option is to proceed with the search functionality that's already built in this PR. If so, we'll need these changes:

- Remove [Create New Account] button from the default state.
- After issuing a search, render a div at the top of the results list (even if it's empty):
`<div>Not found? <button>Create a profile</button> for ${name} (${number})</div>`

The idea here is that every session starts with a single unambiguous action: a search to see if the voucher holder already has a profile. If so, it should appear in the list. If not, there's an affordance to create one.

lederer:
> Adds an unstyled page at `/select`

Can we go with `/search` instead?

> Logging out and back in with an account already selected should go to map page directly

Let's deselect the current profile when logging out. So logging back in lands you on the search page.

flibbertigibbet:
The 'new' button to be implemented in #31 should handle redirecting to switch accounts from a header button.

I'd like to take care of refactoring the form to single-field exact voucher match with creation prompt on no match as @lederer suggests in the next task (#22, the profile page), since that task includes the rest of the profile management form flow on login.

flibbertigibbet:
Created #59 to clear the last selected account from `localStorage` on logout.

PR #65, opened 2019-03-07
Feature/kak/profile page#22

### LINKED ISSUE #22: Implement profile page

### LINKED ISSUE #31: Implement 'edit' and 'new' buttons

### LINKED ISSUE #59: Unset selected profile on logout
When Cognito user logs out, clear selected profile from `localStorage`.

See #55.

flibbertigibbet:
## Overview

Add page with a form to enter or edit user profile. Modify search to only use voucher number and match exactly. (See discussion on #55).

Also adds header links to the search and profile pages, and clears profile on logout.

### Demo

New profile page:
image

Search page:
image

On voucher not found:
image

### Notes

The voucher holder name is now an editable field on the profile page, as it is no longer a search/create key.

## Testing Instructions

 * Should be able to edit profile at `/profile`
 * Should be able to add up to three addresses
 * Address geocoder should work as expected
 * Changes should be saved to `localStorage` and to S3 on clicking 'go'
 * Navigating back to profile page after editing should reflect saved changes
 * Clicking 'cancel' should discard changes and navigate to the previous page (search or map)
 * 'Edit' header button should redirect to profile page
 * 'New' header button should redirect to search page
 * Profile should be unset on logout
 * Search should go to profile page on exact voucher number match
 * Search should prompt to create account if voucher number not found

Closes #22 
Closes #31 
Closes #59

### DIFF
--- a/taui/src/components/edit-profile.js
+++ b/taui/src/components/edit-profile.js
@@ -17,7 +17,10 @@ type Props = {
 }
 
 const firstAddress: AccountAddress = {
-  address: '',
+  location: {
+    label: '',
+    position: null
+  },
   primary: true,
   purpose: DEFAULT_PROFILE_DESTINATION_TYPE
 }
@@ -126,7 +129,10 @@ export default class EditProfile extends PureComponent<Props> {
   addAddress () {
     const destinations = this.state.destinations.slice()
     const newAddress: AccountAddress = {
-      address: '',
+      location: {
+        label: '',
+        position: null
+      },
       primary: !destinations.length,
       purpose: DEFAULT_PROFILE_DESTINATION_TYPE
     }
@@ -205,12 +211,13 @@ export default class EditProfile extends PureComponent<Props> {
           <Geocoder
             className='account-profile__input'
             geocode={geocode}
-            onChange={(result) => editAddress(index,
-              'address',
-              result['place_name'])}
+            onChange={(result) => editAddress(index, 'location', {
+              label: result ? result['place_name'] : '',
+              position: result ? {lat: result['center'][0], lon: result['center'][1]} : {}
+            })}
             placeholder={message('Geocoding.PromptText')}
             reverseGeocode={reverseGeocode}
-            value={{label: destination.address}}
+            value={destination.location}
           />
         </div>
         <div className='account-profile__destination_narrow_field'>
--- a/taui/src/components/edit-profile.js
+++ b/taui/src/components/edit-profile.js
@@ -46,6 +46,7 @@ export default class EditProfile extends PureComponent<Props> {
     this.save = this.save.bind(this)
     this.setGeocodeLocation = this.setGeocodeLocation.bind(this)
     this.setPrimaryAddress = this.setPrimaryAddress.bind(this)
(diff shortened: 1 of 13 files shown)


taui/src/components/select-account.js
```diff
@@ -66,61 +57,46 @@ export default class SelectAccount extends PureComponent<Props> {
...
+    if (!searchVoucher) {
+      this.setState({errorMessage: message('Accounts.SearchError')})
+      return
+    } else {
+      this.setState({errorMessage: ''})
+    }
 
-    const accounts = []
+    var found = false
     Storage.list('')
```
ddohler: Not really related to this PR but something that I thought of while reading this: is there a chance that this filtering approach could get bogged down if S3 gets loaded up with several hundred or thousand profiles? If so, would it be feasible to switch to attempting `Storage.get(searchVoucher)` and then branching based on whether that works or returns a not-found?

I don't think this is something we need to address immediately but if it could become an issue then it would be good to have thought through a solution ahead of time.

taui/src/components/edit-profile.js
```diff
@@ -0,0 +1,415 @@
...
+      return <li
+        key={index}
+        className='account-profile__destination_row'>
+        <div className='account-profile__destination_field'>
+          <Geocoder
+            className='account-profile__input'
+            geocode={geocode}
+            onChange={(result) => editAddress(index, 'location', {
+              label: result ? result['place_name'] : '',
+              position: result ? {lat: result['center'][0], lon: result['center'][1]} : {}
```
ddohler: I think these coordinates are flipped; the coordinates that got saved to Local Storage ended up being ` "position": { "lat": -71.073061, "lon": 42.349847}` which is in Antarctica.

taui/src/components/select-account.js
flibbertigibbet: The initial prototype won't see that level of usage, which is why I went with this simple directory listing search, although I originally expected this page to provide inexact matches (actually search). Since we've decided to limit this to exact matches, we could replace this with a simple `get`.

taui/src/components/edit-profile.js
flibbertigibbet: I added a commit to switch to the logic used by another geocoding handler in the app, for consistency. It looked like the `position` it stored was the same as this for a test address. In any case, the result should have better type checking now.

taui/src/components/select-account.js
ddohler: Got it, that makes sense, thanks! I don't think we need to address this right now but it's good to know that we can.

flibbertigibbet: Created #66. We'll want to distinguish between file not existing and other `get` error states.

PR #68, opened 2019-03-11
Allow using the site anonymously

### LINKED ISSUE #64: Explore anonymous accounts
Test enabling anonymous login via Cognito for anonymous browsing.

(Timeboxed exploration.)

flibbertigibbet:
## Overview

Allow bypassing login to browse anonymously instead. Store profile in `localStorage` without using S3 for anonymous user. (Profile settings will survive a page refresh but cannot be stored to a remote account when browsing the site anonymously.)

Hide voucher number and voucher holder name, and change "sign out" button to "sign in" in the header when browsing anonymously.

### Demo

New anonymous sign-in page option:
image

Anonymous user's profile page and header:
image

### Notes

For our use, it is not necessary to enable [unauthenticated roles](https://docs.aws.amazon.com/cognito/latest/developerguide/identity-pools.html) for the Cognito identity pool, as the anonymous user does not need any access to AWS resources. If the anonymous user did need access to S3 or other AWS resources but with different permissions than those granted to authenticated users, we would need to set up using credentials for the unauthenticated Cognito role. The Amplify JS library does not provide tools for any kind of anonymous/unauthenticated user support.

## Testing Instructions

 * Click to 'continue without an account' from the sign-in page
 * Should be taken directly to the profile page (not search)
 * Attempting to navigate directly to `/search` when anonymous should redirect to `/profile`
 * Voucher number and voucher holder name should be hidden in header and on forms
 * 'Sign in' header button should redirect as expected to log in
 * Switching between being signed and and browsing anonymously should behave as expected

Closes #64.

### DIFF
--- a/taui/src/components/with-authenticator.js
+++ b/taui/src/components/with-authenticator.js
@@ -2,8 +2,8 @@
 import { Component, Fragment } from 'react'
 import { Authenticator } from 'aws-amplify-react/dist/Auth'
 
-import {storeConfig} from '../config'
-import {PROFILE_CONFIG_KEY} from '../constants'
+import {clearLocalStorage, storeConfig} from '../config'
+import {ANONYMOUS_USERNAME, PROFILE_CONFIG_KEY} from '../constants'
 import type {AccountProfile} from '../types'
 
 import CustomHeaderBar from './custom-header-bar'
@@ -38,6 +38,9 @@ export default function withAuthenticator (Comp, includeGreetings = false,
           signUpConfig
         }
       }
+
+      // Load the selected user profile from localStorage, if any
+      this.props.loadProfile()
     }
 
     changeUserProfile (profile: AccountProfile) {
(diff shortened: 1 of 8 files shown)


taui/src/components/with-authenticator.js
```diff
@@ -46,17 +49,38 @@ export default function withAuthenticator (Comp, includeGreetings = false,
...
+        // Handle full page reload when browsing site anonymously
+        this.setState({authState: 'signedIn', authData: {username: ANONYMOUS_USERNAME}})
+      } else {
+        this.setState({authState: state, authData: data})
+      }
+
+      // Clear all local storage on logout
       if (state === 'signedOut') {
-        this.changeUserProfile(null)
+        clearLocalStorage()
```
ddohler: I think that something may be going wrong with this. If I sign in, select a voucher profile, and sign out from the map page, then if I sign in again, I am sent to the map page immediately rather than the voucher search page, and the header bar displays the profile that I selected previously.

Something similar happens when switching between anonymous and signed-in accounts: if I set up a profile anonymously and then click "Sign In" from the map page, when I sign in, I'm directed to the map page instead of the voucher search page. The header bar looks like I'm signed in, but the voucher number is blank: 
Screen Shot 2019-03-12 at 9 23 19 AM 

At that point, if I click "Edit", I receive a blank page and a console error:
`TypeError: profile.destinations is undefined`
Screen Shot 2019-03-12 at 9 26 27 AM

flibbertigibbet: This should be fixed now. I think the issue was setting the profile to an empty object instead of null in `signIn` in the header.

taui/src/components/with-authenticator.js
```diff
@@ -46,17 +49,39 @@ export default function withAuthenticator (Comp, includeGreetings = false,
     }
 
     handleAuthStateChange (state, data) {
-      this.setState({ authState: state, authData: data })
-      // Unset user profile on logout
-      if (state === 'signedOut') {
-        this.changeUserProfile(null)
+      const { userProfile } = this.props.data
+      // Create new emtpy profile to use when browsing anonymously
```
ddohler: Whoops, this looks like it got clobbered by the force-push.

taui/src/components/with-authenticator.js
```diff
@@ -46,17 +49,39 @@ export default function withAuthenticator (Comp, includeGreetings = false,
...
+          destinations: [],
+          hasVehicle: false,
+          headOfHousehold: ANONYMOUS_USERNAME,
+          key: ANONYMOUS_USERNAME,
+          rooms: 0,
+          voucherNumber: ANONYMOUS_USERNAME
+        }
+        this.changeUserProfile(profile)
+        // Tell auth library that the anonymous user is signed in
+        this.setState({authState: 'signedIn', authData: {username: ANONYMOUS_USERNAME}})
```
ddohler: This ends up redirecting to `/map` rather than `/profile` as expected.

flibbertigibbet:
Updated to require at least one destination to go to the map page.

Issue #73, opened 2019-03-18
Neighborhood list cards


flibbertigibbet:
Decide what elements of the neighborhood data should be displayed in the neighborhood summary sidebar  cards on the map page, and modify the layout to show those data elements, based on the wireframes.

![image](https://user-images.githubusercontent.com/960264/54537934-b941c600-4969-11e9-9f8f-31ed01dcb285.png)

Depends on fixing the column names in the provided data to an expected set, and establishing which of them to display.

Issue #72, opened 2019-03-18
Sort neighborhoods by desireablility


flibbertigibbet:
List neighborhoods in sidebar using metrics on the neighborhood, instead of or in addition to travel time.

Issue #71, opened 2019-03-18
Add non-transit modes


flibbertigibbet:
The `taui` project only uses transit results from `r5` static analysis results. The [r5 documentation](https://github.com/conveyal/r5/blob/dev/docs/static-output.md#per-origin-file) for the static analysis results file seem to indicate the `_paths.dat` file (which seems to be the "per-origin file" in the documentation) should have non-transit travel times at the head of the file. However, there are no code examples or documentation for how to read out non-transit times from the binary-encoded rows of integers in the `paths.dat` file. Changing the [index offset here](https://github.com/azavea/echo-locator/blob/develop/taui/src/utils/parse-paths-data.js) may point to driving routes.

Determine whether the current static results contain driving data, and if so, add support for reading driving routes and times, and parsing route legs.

flibbertigibbet:
Handle by estimating drive times using distance as a proxy for ordering. Do not display the estimated drive times to the user. There will be no routes to display on the map for driving results.

PR #74, opened 2019-03-18
Feature/kak/neighborhood routing#25

### LINKED ISSUE #25: Load neighborhood results on map page
Query for travel times using `dat` files from S3, sort neighborhood results, and load the top three results in the sidebar on page load or form input change. Determine a basic starting algorithm for neighborhood sorting/weighting (could be simple travel time).

Display neighborhood details in sidebar, with at least travel time (other metrics in card may change).

Implement mode toggle and time of day toggle on form (but not from/to; will always be 'from').

~Load route path for the first result on the map. Do not yet implement interactivity (clicking neighborhood in sidebar or pagination). Put markers for each neighborhood on map.~ (Do this in #27)

image

Depends on #24.
Depends on #29.

flibbertigibbet:
## Overview

 - Get the three route results for all neighborhoods for the set origin and time of day on map page load or input change and display in sidebar cards the three neighborhoods with the shortest best route.
 - Updates the data using the last provided spreadsheet, from March 13th.
 - Adds a workaround for a Firefox Leaflet rendering issue

### Demo

image

On time of day change:
image

### Notes

The updated spreadsheet provided had a header row of column names that did not match the first spreadsheet we received. Here I renamed a couple of them in adding the CSV (one had a misspelling.) We will need to establish what the format of the neighborhood data should be and what fields to expect before we can read from the neighborhood GeoJSON properties extensively in the app.

Created #71 for adding non-transit travel modes. We need at least driving results support.

Created #72 for establishing an approach for using neighborhood data to rank them by metrics such as affordability and school ratings.

Created #73 for modifying the card layout to show neighborhood data.

The [Firefox rendering issue](https://github.com/Leaflet/Leaflet/issues/4686) surfaced when rendering the full list of neighborhoods in the sidebar, as the current styling will match the map height to the sidebar. 

There are some minor package updates available, which I'm postponing pulling in until [this router issue](https://github.com/ReactTraining/react-router/issues/6630) gets addressed.

I've left in the `Log` sidebar component for now, as it has been helpful for debugging.

## Testing Instructions

 * Create a profile and go to map page
 * Route results for the primary destination should show in sidebar
 * Changing the destination or time of day inputs should update results as expected
 * Page reload and in-app navigation should function as expected (load results for the set inputs)

Closes #25


taui/src/taui.css
```diff
@@ -39,7 +48,7 @@
   width: 414px;
   top: 0;
   left: 0;
-  height: 100%;
+  height: 1000px;
```
flibbertigibbet: Fixing the `Dock` component height for now. This should be addressed in a styling pass.

PR #81, opened 2019-03-21
Feature/kak/weigh neighborhoods#72

### LINKED ISSUE #72: Sort neighborhoods by desireablility

flibbertigibbet:
## Overview

Sort neighborhoods displayed in sidebar by a score weighting both travel time and rent difference to the fair market rate, instead of sorting only by travel time.

Also refactors sort logic from the `Dock` sidebar component to a Redux selector, `neighborhoods-sorted-with-routes.js`.

### Demo

Some neighborhoods for which the rent differences outweighed the travel times (results 4-6):
image

The first three results for that origin had travel times that outweighed the rent differences:
image

### Notes

We will likely revisit how the scores are assigned. This should set up a framework that we can easily modify to add other factors, or adjust the factor weighting. For that purpose, factors for the final score are capped to a range, then normalized to percentages. Each factor has an assigned weight percentage for its influence on the final score; those factor weights should always sum to 1.

I've added some values to the sidebar cards, for debugging weighting; we will probably remove or modify those in #73.

## Testing Instructions

 * Sidebar results should be sorted by both travel time and affordability (rent difference)

Closes #72


taui/src/selectors/neighborhoods-sorted-with-routes.js
```diff
@@ -23,9 +27,33 @@ export default createSelector(
...
-    return sortBy(neighborhoodsWithRoutes, 'time')
+      // Map weighting values to percentages
+
+      // Routable neighborhoods outside the max travel time window will be filtered out.
+      // Smaller travel time is better; larger timeWeight is better (reverse range).
+      const timeWeight = time < MAX_TRAVEL_TIME ? scale(time, 0, MAX_TRAVEL_TIME, 1, 0) : 1
+
+      // Normalize rent differential to fair market value to a range;
+      // any value outside the range will be given the min/max value for the range.
+      const useRentDiff = n.properties.rent_diff > RENT_DIFF_RANGE ? RENT_DIFF_RANGE
```
ddohler: It'll be more verbose, but I think that a multi-line if-else would be preferable here; the nested ternary operators make this a bit tough to follow (at least, it took me longer to read this than I think it would have if it had been written out).

PR #82, opened 2019-03-21
Add LogRocket tracking

### LINKED ISSUE #57: Add LogRocket
- Add LogRocket's [main](https://www.npmjs.com/package/logrocket) and [React](https://www.npmjs.com/package/logrocket-react) libraries as dependencies
 - Configure with separate keys per environment
 - Add calls to initialize it on app start and `identify` the user on log-in with string identifier
 - Review network calls to see if any (i.e., login) require [sanitation](https://docs.logrocket.com/reference#network)
 - Disable [console log capture](https://docs.logrocket.com/reference#console) and/or remove `console` logging of any user data
 - If it's not too difficult, [set the release version](https://docs.logrocket.com/reference#release)

ddohler:
## Overview

Adds the LogRocket library for usage and error tracking.

### Demo

Optional. Screenshots, `curl` examples, etc.
Screen Shot 2019-03-21 at 3 13 57 PM

### Notes

- I ended up disabling network logging entirely. Some of the larger requests for data files were causing LogRocket to struggle, and the non-static requests all potentially need sanitization because they contain info that might be considered sensitive (voucher numbers, BHA user logins, and work / childcare addresses). Between these two factors, writing custom sanitization seemed like overkill and likely to remove most of the interesting requests the app was making, so I turned tracking off completely.
- It's not entirely possible to control LogRocket sessions from the frontend. So even if you call `identify()` with changed parameters, LogRocket often seems to continue the existing tracking session. It might be possible to work around this by clearing cookies or local storage, but it didn't seem worth it at this stage.
- We will need the client to create LogRocket accounts and provide us with the app IDs they want us to use. The lack of LogRocket app IDs shouldn't cause problems.
- I added code that I _think_ will insert the git commit into LogRocket so that we can track errors that occur on different versions of the app. I broke it into a separate commit so that it's easy to revert if it causes problems.

## Testing Instructions

 * First, test out the app immediately after checking out this branch to make sure that it functions properly without a defined `LOGROCKET_APP_ID`. You shouldn't see any console errors.
 * Pull in the changes in `env.yml.tmp` into your local `env.yml`, sign up for LogRocket, and create a LogRocket app. Put the LogRocket App ID into your `env.yml` in the appropriate spot. You should probably restart the dev server at this point to be safe.
 * Load the page and browse anonymously. Open your LogRocket dashboard and make sure you can see an Anonymous user session.
 * Load the page with a private browsing tab and log in. You should see a session with your username in LogRocket.

Resolves #57


taui/package.json
```diff
@@ -50,6 +50,8 @@
     "jsolines": "~2.1.0",
     "leaflet": "~1.3.4",
     "lodash": "~4.17.11",
+    "logrocket": "^0.6.19",
```
flibbertigibbet: Nitpick: let's pin the version numbers here with `~` and specify `3.0.0` for the react package.

PR #93, opened 2019-03-29
Feature/kak/limit bucket#51

### LINKED ISSUE #51: Limit Cognito S3 bucket access to current user pool

flibbertigibbet:
Updates Amplify CLI version and rebuilds production Amplify resources for the new CLI.

Staging resources have not been recreated yet, to preserve existing users and profiles; due to issues with the outdated Amplify version, it may be necessary to rebuild the staging resources 

Closes #51. (Verified it is not an issue currently in either environment.)


flibbertigibbet:
Already deployed; merging.

PR #94, opened 2019-03-29
Style app

lederer:
## Overview

Massive styling PR. Most aspect of the app have been touched.

Note placeholder elements on neighborhood detail screen for descriptions and images.

**Not done: map styling.** This will be done in a follow-up PR.

### Demo

image

image


taui/configurations/default/settings.yml
```diff
@@ -2,7 +2,6 @@ cloudfront: EJKZ46R7IC1BB
 entries:
   - src/index.js:assets/index.js
   - src/img/BHAlogo.png:assets/BHAlogo.png
-  - src/img/trash-alt.svg:assets/trash-alt.svg
```
flibbertigibbet: :wastebasket:

taui/src/components/edit-profile.js
```diff
@@ -268,87 +269,86 @@ export default class EditProfile extends PureComponent<Props> {
...
+            <span className='account-profile__destination-list-heading'>
+              {message('Profile.Address')}
+            </span>
+          </div>
+          <div className='account-profile__destination_field'>
+            <span className='account-profile__destination-list-heading'>
+              {message('Profile.Purpose')}
+            </span>
+          </div>
+          {showAllColumns && <>
```
flibbertigibbet: Interesting, hadn't run into [shortened fragment declaration](https://reactjs.org/docs/fragments.html#short-syntax) before.

taui/src/components/geocoder.js
```diff
@@ -133,12 +133,25 @@ export default class Geocoder extends Component<Props> {
     }
   }
 
+  wrapperStyle = {
```
flibbertigibbet: These select styles could be moved to `constants.js`, so they aren't repeated in multiple files.

taui/src/components/neighborhood-details.js
```diff
@@ -53,86 +57,136 @@ export default class NeighborhoodDetails extends PureComponent<Props> {
...
+            type={isFavorite ? 'star' : 'star-o'}
             onClick={(e) => setFavorite(id, userProfile, changeUserProfile)}
-            style={{cursor: 'pointer'}} />
-          <span>{town} - {id}</span>
+          />
+          {town} &ndash; {id}
+          <Icon className='neighborhood-details__marker' type='map-marker' />
+        </header>
+        <div className='neighborhood-details__trip'>
+          {Math.round(time)}&nbsp;
```
flibbertigibbet: Isn't `time` an integer?

taui/src/components/route-card.js
```diff
@@ -1,7 +1,9 @@
 // @flow
 import Icon from '@conveyal/woonerf/components/icon'
-import message from '@conveyal/woonerf/message'
 import React from 'react'
+import message from '@conveyal/woonerf/message'
```
flibbertigibbet: Apparently the linter thinks this is fine, but did you intend to separate the `@conveyal` imports here?

taui/src/components/route-card.js
```diff
@@ -16,45 +18,74 @@ export default class RouteCard extends React.PureComponent<Props> {
...
       title,
       userProfile
     } = this.props
 
+    const markerClass = `neighborhood-summary__marker ${neighborhood.active ? 'neighborhood-summary__marker--on' : ''}`
+    const { time } = neighborhood
+    const originLabel = origin ? origin.label || '' : ''
+    const currentDestination = userProfile.destinations.find(d => d.location.label === originLabel)
+
+    const modeKey = neighborhood.segments && neighborhood.segments.length
```
flibbertigibbet: Referencing `segments` isn't a reliable way to determine the user profile mode; unreachable places in transit mode have no segments. It would be better to check `hasVehicle` on `userProfile`.

taui/src/components/route-segments.js
```diff
@@ -15,37 +15,31 @@ export default function RouteSegments ({hasVehicle, routeSegments, travelTime})
...
-        </td>
-      </tr>
+    <div className='route-segments'>
+      <div className='route-segments__best-trip'>
+        Take&nbsp;
+        {bestJourney.map((segment, index) => (
+          <Segment key={index} segment={segment} />
+        ))}
+        {travelTime > 120
+          ? <span className='decrease'>inaccessible within 120 minutes</span>
```
flibbertigibbet: I noticed this was in `taui`, too; we should move this user-facing string to `messages.yml` (also "Take" above here).

taui/src/components/edit-profile.js
lederer: Right? It's great.

taui/src/components/neighborhood-details.js
lederer: For some reason when user has a car, it's a float.

flibbertigibbet: Ah, okay. I'll go fix that in another PR later.

lederer:
Addressed all comments except:

- Scrolling / viewport sizing. Currently works fine on Chrome/Mac. I'll follow up with x-browser fixes in another PR.
- Green highlight on card hover. Left this in for the moment. May change it as part of a map styling PR coming soon.

flibbertigibbet:
I've been putting off some map page cleanup work due to the viewport issues on Linux Chrome and Firefox, so it would be helpful for me for those to be addressed.

PR #104, opened 2019-04-02
Feature/kak/extra markers#97

### LINKED ISSUE #69: Restore style linting on CI
Add back `mastarm lint-styles` to the `test` command in `taui/package.json` used by Travis once the style warnings and errors have been addressed.

Removed as part of switching linter for #67.

### LINKED ISSUE #90: End marker should not be draggable
The end marker for a route to a neighborhood should not be draggable (nothing happens, but the UI allows it to be moved.)

### LINKED ISSUE #97: Render map markers for all sidebar neighborhoods
Currently on the map we show a marker for the featured destination (eg, work), a marker for the active neighborhood, and, iff transit, the route between them.

We should also show markers for the other (up to two) neighborhoods currently in the sidebar. They will have no icon and be colored `$brand-blue` `#15369d`

flibbertigibbet:
## Overview

Display differently styled map markers for neighborhoods listed in the sidebar (#97).

Also disables dragging end markers and adds back style linting to `yarn test`.

### Demo

image

### Notes

This required refactoring to move the pagination logic out of the sidebar state to share with the map.

## Testing Instructions

 * Go to map page
 * Should have blue markers with no inset icon for non-active neighborhoods listed in sidebar
 * Clicking a blue marker should go to the details view for that neighborhood
 * Pagination should update markers
 * Toggling between saved and full list and between details and list should still work as expected

Closes #90 
Closes #97 
Closes #69


taui/src/actions/network.js
```diff
@@ -127,6 +128,9 @@ export const loadDataset = (
   dispatch(loadDataFromJSON('assets/neighborhoods.json', 'set neighborhoods'))
   dispatch(loadDataFromJSON('assets/neighborhood_bounds.json', 'set neighborhood bounds'))
 
+  // Start on first page of neighborhoods
+  dispatch(setPage(0))
```
ddohler: Is `dispatch` necessary here since it is called by `setPage`?

flibbertigibbet: I believe it's passing through `dispatch` to `setPage` to invoke (similar to `loadDataset` in `network.js`), so yes.

PR #149, opened 2019-04-17
Add non-ECC zip code data

### LINKED ISSUE #127: Data update

flibbertigibbet:
## Overview

Add a step to the input process to include the added non-ECC zip codes in `non_ecc_max_subsidies.csv`.

### Demo

image

image

image

### Notes

Connects #127 but does not close it, as we should be getting another round of data with images and descriptions for the new zip codes.

## Testing Instructions

 * Should have about twice as many zip codes on the map
 * Missing data for the added zip codes (images, etc.) should not cause any errors

Connects #127.


flibbertigibbet:
We should be receiving a full dataset and remove `add_non_ecc.py` entirely with the next update.

PR #157, opened 2019-04-19
Style search weights and ECC field

lederer:
## Overview

- Style search weight fields on Profile page. Moved car checkbox higher up so it doesn't get lost.
- Render ECC boolean field as another table row in neighborhood summary and detail
- [EDIT:] Unlink thumbnails in search result cards

### Demo

Screen Shot 2019-04-19 at 2 42 00 PM

Screen Shot 2019-04-19 at 2 42 14 PM

Screen Shot 2019-04-19 at 2 42 22 PM

### Notes

I first experimented with rendering the ECC field as a sort of tag/badge, but it was too prominent, and looked too similar to the treatment of transit lines on the neighborhood detail page.

### DIFF
--- a/taui/configurations/default/messages.yml
+++ b/taui/configurations/default/messages.yml
@@ -102,8 +102,9 @@ Profile:
   DeleteProfile: Delete profile
   DeleteProfileError: Failed to delete profile. Please try again.
   Destinations: Frequent destinations
-  ImportanceAccessibility: Accessibility
-  ImportanceSchools: Schools
+  ImportanceAccessibility: Travel time
+  ImportanceHeading: How important are the following?
+  ImportanceSchools: School quality
   ImportanceViolentCrime: Violent crime
   Primary: Primary
   Purpose: Purpose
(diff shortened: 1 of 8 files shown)


lederer:
Unfortunately we can't have a click target that large on a small card that is itself a click target. 

I retrained the attribution in the `title` attribute, so it appears on hover. And the thumbs remain linked on the detail screen. I believe that is sufficient at least for the license Wikipedia uses ("You must give appropriate credit, provide a link to the license, and indicate if changes were made. You may do so in any reasonable manner. …")

flibbertigibbet:
>  ("You must give appropriate credit, provide a link to the license, and indicate if changes were made. You may do so in any reasonable manner. …")

This file page link is the "link to the license" and modeled on the suggested example attribution Wikipedia provides on the file page, which is wrapped with a link. [The license attribution rules here](https://en.wikipedia.org/wiki/Creative_Commons_license) say "it is nice to link" to the author, license, and work title, which are all on the file page.

How about putting a tiny link below the image?

lederer:
> How about putting a tiny link below the image?

This would cramp an already packed space. I think we're fine as is, bc:
1. The attribution is published on that page – via hover, which is reasonable for a tiny thumb.
1. We do link to the wikipedia file page – from the detail screen (which you land on if you click the thumb). 

I believe these satisfy the license requirements ("nice to link", "in any reasonable manner") and are particularly reasonable for a non-profit site targeting a specialized audience.

PR #158, opened 2019-04-22
Feature/kak/toggle ecc#129

### LINKED ISSUE #129: Allow user to turn off use of non-ECC neighborhoods
Add checkbox (to map sidebar controls?) that defaults to checked to control whether to use non-ECC neighborhoods. If unchecked:

 - hide non-ECC neighborhoods on map
 - remove non-ECC neighborhoods from set for routing/sidebar and reload results

~Optional for demo milestone.~

Depends on #127.

flibbertigibbet:
## Overview

Add a toggle button to the map sidebar form to show/hide non-Expanded Choice Community zip codes.

### Demo

Checked:
image

Unchecked:
image

Route, details, and start and end markers continue to display in detail view if box is unchecked while viewing a non-ECC zipcode:

image

### Notes

Moving the neighborhood bounds layer to a layer group of custom vector grids was necessary to make the layer dynamic.

## Testing Instructions

 * Use non-ECC checkbox should default to checked on page reload or navigation to map
 * Unchecking box should hide non-ECC bounds and centroid markers on map
 * Unchecking box should exclude non-ECC neighborhood results from sidebar
 * Unchecking box while in details for non-ECC zipcode should not hide route or details
 * Marker highlighting for sidebar neighborhoods should continue to work as expected

Resolves #129


taui/src/components/form.js
```diff
@@ -49,7 +49,8 @@ export default class Form extends React.PureComponent {
       destination,
       network: networks && networks.length ? {
         label: networks[0].name, value: networks[0].url
-      } : null
+      } : null,
+      useNonECC: true
```
ddohler: I might be missing something that happened during later conversations, but my impression from the initial email on this feature was that this should be unchecked by default so that we nudge users toward the expanded choice communities by only showing those by default. Feel free to disregard if that has changed or I'm misunderstanding.

PR #163, opened 2019-04-24
Update data with April 24th source

flibbertigibbet:
## Overview

Update data source.

### Demo

Jamaica Plain (02130) now has images:
image

### Notes

The population column is missing in the latest update. This is a display column.

## Testing Instructions

 * `cd neighborhood_data && ./update_data.sh`
 * `cd ../taui && yarn start`
 * Show non-ECC zip codes
 * Should now have more images and descriptions


taui/src/components/neighborhood-details.js
```diff
@@ -190,7 +190,7 @@ export default class NeighborhoodDetails extends PureComponent<Props> {
     // Look up the currently selected user profile destination from the origin
     const originLabel = origin ? origin.label || '' : ''
     const currentDestination = userProfile.destinations.find(d => d.location.label === originLabel)
-    const { id, town, wikipedia } = neighborhood.properties
+    const { id, town, town_website_description } = neighborhood.properties
```
ddohler: This is making the linter grumpy, it probably needs to be switched off here and below.

PR #168, opened 2019-04-26
Only calculate active route

### LINKED ISSUE #166: Slow response on navigation back to list
Navigating from neighborhood details back to the full list with the expanded list of zip codes now results in a sometimes visible delay before the map layers reload. Modifying the `draw-neighborhood-routes.js` selector to pull only the active neighborhood should address it. (Portion of Chrome performance log captured during back-to-list navigation below.)

echo_performance_draw_routes

flibbertigibbet:
## Overview

Improve performance on navigation back to list from details by only calculating the active route to draw.

All route paths were being calculated before, but only the active one was styled as visible. As the styling for the active route only changes when all routes are recalculated, and the non-active routes are never displayed, it is not necessary to compute them all at once.

### Demo

0.8 ms was spent in `getStopPositions` here, down from 138 ms in #166 performance log.
image

## Testing Instructions

 * Navigation back from details to list should have no visible map redraw delay
 * Active route display on sidebar card hover or detail select should still function as expected

Closes #166.


flibbertigibbet:
This should be ready for another look. I did a little refactoring, and moved the zoom out from the route component to the map.

taui/src/components/map.js
```diff
@@ -109,6 +110,15 @@ export default class Map extends PureComponent<Props, State> {
     console.error(error)
   }
 
+  componentWillReceiveProps (nextProps) {
+    const needToZoomOut = !nextProps.showDetails && this.props.showDetails
+    const map = this.refs ? get(this.refs, 'map.leafletElement') : null
+
+    if (needToZoomOut && nextProps.neighborhoodBoundsExtent) {
```
ddohler: :+1: For moving this into the map component.

PR #189, opened 2019-05-28
Feature/kak/hide walk leg#186

### LINKED ISSUE #186: Hide final walk leg of route
Do not display final walk leg to neighborhood centroid of transit routes.

flibbertigibbet:
## Overview

Do not draw the final walk leg of transit routes, which goes from the nearest transit stop to the neighborhood centroid. Also do not show marker on neighborhood centroid when displaying route.

### Demo

Start walk displays from user address to boarding stop. End walk from disembarking stop to neighborhood centroid used for routing is not displayed.
image

### Notes

The walk legs were not calculated by the routing engine to follow the street network, but are instead a straight line from the first/last stop to the origin/destination.

## Testing Instructions

 * Open a profile in transit mode
 * Go to neighborhood details
 * Should not have walk leg to neighborhood centroid

Closes #186

### DIFF
--- a/taui/src/selectors/draw-neighborhood-route.js
+++ b/taui/src/selectors/draw-neighborhood-route.js
@@ -27,6 +27,8 @@ export default createSelector(
     const transitStyle = {...TRANSIT_STYLE, ...applyStyle}
     const stopStyle = {...STOP_STYLE, ...applyStyle}
     const segments = get(transitive, 'journeys[0].segments', [])
+    // Remove final walk leg
+    segments.pop()
     return {
       index,
       id: transitive.id,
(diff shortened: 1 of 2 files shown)


taui/src/selectors/draw-neighborhood-route.js
```diff
@@ -27,6 +27,8 @@ export default createSelector(
     const transitStyle = {...TRANSIT_STYLE, ...applyStyle}
     const stopStyle = {...STOP_STYLE, ...applyStyle}
     const segments = get(transitive, 'journeys[0].segments', [])
+    // Remove final walk leg
+    segments.pop()
```
ddohler: 🍿

flibbertigibbet:
Good catch. I could only reproduce that by toggling the network. This should be ready for another look.

PR #196, opened 2019-05-30
Display destination type label in drop-down

### LINKED ISSUE #183: Modify form field labels
For the map page destination drop-down, show the type label (e.g., 'work') alongside the address in the options.

flibbertigibbet:
## Overview

Display destination purpose type before the address in the map page destination drop-down.

### Demo

echo_labeled_addresses_dropdown

## Testing Instructions

 * Map page should load with [destination type]: [address] in the destination drop-down text field
 * Drop-down options should match above format
 * Toggling destination should still behave as expected

Closes #183.

### DIFF
--- a/taui/src/components/form.js
+++ b/taui/src/components/form.js
@@ -100,7 +100,7 @@ export default class Form extends React.PureComponent<Props> {
     }
     const position = destination.location.position
     return (position.lat !== 0 && position.lon !== 0) ? {
-      label: destination.location.label,
+      label: destination.purpose + ': ' + destination.location.label,
       position: position
     } : null
   }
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
@lederer might want to take a look to confirm this is what he had in mind.

lederer:
This is great. Thank you!

Though this is likely out of scope for this PR, would you happen to know if we're able to customize the address format of the menu options? Specifically I'd like the drop "United States" and abbreviate state to 2-letter "MA". This would minimize line-wrapping, without truncating zip code, which may be useful info.

flibbertigibbet:
Created #198 @lederer

PR #203, opened 2019-06-06
Feature/kak/hide school ranks#179

### LINKED ISSUE #179: Hide some school ranks in UI
Based on an added data source column to indicate which neighborhoods should have their school rankings hidden in the UI (for those areas averaged due to school choice), do not show the school ranking value or quintile label in the neighborhood detail if marked to be hidden, but use the school ranking value when sorting.

flibbertigibbet:
## Overview

Implement the meters for school ranking, crime rate, and housing availability from #199. Hide the school ranking meter for school choice zip codes (#179).

### Demo

echo_list_school_choice_hidden_meter

image

### Notes

The data source provides the education percentile, but only the crime rank quintile. The number of houses to display in the five-house icon chart is provided directly by the data source.

## Testing Instructions

 * Meters should display as expected
 * Meter shading (orange-yellow, yellow, or green) should match value range
 * Housing availability shading and count should display as expected
 * School choice zip codes (i.e., Boston-labelled zip codes) should hide school rank meter

Closes #179.

### DIFF
--- a/taui/src/components/neighborhood-list-info.js
+++ b/taui/src/components/neighborhood-list-info.js
@@ -13,16 +13,17 @@ export default function NeighborhoodListInfo ({neighborhood}) {
   const crimeQuintile = props['violentcrime_quintile']
   const edPercentile = props['education_percentile']
   const houses = props['house_number_symbol']
+  const isSchoolChoice = !!props['school_choice']
 
   return (
     <table className='neighborhood-facts'>
       <tbody>
-        <tr>
+        {!isSchoolChoice && <tr>
           <td className='neighborhood-facts__cell'>{message('NeighborhoodInfo.EducationCategory')}</td>
           <td className='neighborhood-facts__cell'>
             <Meter value={edPercentile} />
           </td>
-        </tr>
+        </tr>}
         <tr>
           <td className='neighborhood-facts__cell'>{message('NeighborhoodInfo.ViolentCrime')}</td>
           <td className='neighborhood-facts__cell'>
--- a/taui/src/components/neighborhood-list-info.js
+++ b/taui/src/components/neighborhood-list-info.js
@@ -9,25 +9,30 @@ export default function NeighborhoodListInfo ({neighborhood}) {
     return null
   }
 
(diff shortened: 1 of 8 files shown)


taui/src/components/neighborhood-list-info.js
```diff
@@ -9,25 +9,31 @@ export default function NeighborhoodListInfo ({neighborhood}) {
...
-            <Meter value={86} average={58} tier='high' tooltip='hello' />
+            <Meter value={edPercentile} />
           </td>
-        </tr>
+        </tr>}
         <tr>
           <td className='neighborhood-facts__cell'>{message('NeighborhoodInfo.ViolentCrime')}</td>
           <td className='neighborhood-facts__cell'>
-            <Meter value={72} average={66} tier='med' />
+            <Meter value={crimeQuintile} average={3} max={5} />
```
ddohler: I think this needs to get inverted now that we're calling it "Public Safety" rather than "Crime". If I prioritize safety above all else in the weighting, the first page of results shows zip codes with public safety in the lowest quintile, and shaded red, which implies they're relatively unsafe, but I think it's actually showing their crime rate quintile, meaning they're relatively safe. I Googled a few of the zip codes and it does seem like the ones with the higher value here are relatively less safe than the ones with lower values.

So I think we need to subtract from 5 or find some other way to flip the meaning of the value before displaying it so that this graph (for example) shows up with two big green bars:
Screen Shot 2019-06-06 at 1 06 24 PM

flibbertigibbet: From the spreadsheet header for the column: "used to inform public safety symbol (lower is better). 2 and above green, 3 lighter green, 4 yellow, 5 orange."

flibbertigibbet: The actual range in the source data appears to be 0-5, not 1-5 or 0-4.

flibbertigibbet:
I've pushed a change to invert the range for the crime quintiles, and to force out-of-range values in range by using 1 for neighborhoods that have 0 for their crime quintile (maximum in the source data is 5).

PR #213, opened 2019-06-10
Feature/kak/hover interactivity#161

### LINKED ISSUE #161: Additional mouse in/out interactions on neighborhood bounds
See #160.

### Search screen
 - When hovering a sidebar card, paint the corresponding polygon fill green.
 - When hovering a map poly that has a corresponding sidebar card, make it the active neighborhood (deactivate other neighborhoods; make its sidebar and map markers green; render its route)
 - When hovering a **routable** map poly that does not have a corresponding sidebar card, deactivate any active neighborhood (hide route, change marker to blue, remove green fill, etc).
    - Bonus point 1: Add a green marker to the hovered poly (only while hovered).
    - Bonus point 2: Render a route to that marker (only while hovered). Obviously skip this if non-trivial.
- When hovering a **non-routable** poly:
  - Do not activate it (do not paint it green, do not show a marker).
  - Cursor should remain `grab` rather than `pointer`. 
  - Clicking should do nothing.
  - Show a tooltip with the neighborhood name and "Cannot be reached by transit"

### Detail screen
 - The selected neighborhood's polygon should be filled green and its centroid (or route endpoint) should have a green marker.
 - When hovering a **routable** map poly other than the selected one, paint it green and show its name in a tooltip. Do not render its marker or route. Keep the selected neighborhood polygon green.
- When hovering a **non-routable** poly:
  - Do not activate it (do not paint it green, do not show a marker).
  - Cursor should remain `grab` rather than `pointer`. 
  - Clicking should do nothing.
  - Show a tooltip with the neighborhood name and "Cannot be reached by transit"

### LINKED ISSUE #212: Map marker disappears on list item hover
On hovering over a neighborhood card, the corresponding map marker disappears.

flibbertigibbet:
## Overview

Modify map interactivity:

 - make tooltips dynamic (show message in toolitp if neighborhood is not reachable)
 - make map marker green on list card hover (fixes #212)
 - set active neighborhood on map hover in list view, but not detail view (shows route)
 - style reachable neighborhood bounds as green on hover in both list and detail views

### Demo

On list card or neighborhood map bounds hover:
image

Hovering over a reachable neighborhood in details view:

echo_routable_hover

Hovering over a non-reachable neighborhood in details view (see tooltip at bottom):

echo_nonroutable_hover

### Notes

The end leg and neighborhood centroid end marker were removed from route rendering in #189, as requested by the client. The centroid markers are still present here in list view for the three currently listed neighborhoods, but not in the detail view, and the end leg does not render anywhere.

No modifications are made to map cursor appearance, as Leaflet doesn't support per-feature interactivity (see [this comment](https://github.com/azavea/echo-locator/issues/161#issuecomment-500472444)).

There is a small delay added here to showing the route on hover, as doing so immediately would cause Leaflet event errors as the data updates on the change to the active neighborhood.

## Testing Instructions

 * Map and list card interactions should function as expected
 * Reachable neighborhood bounds should turn green on hover, and in list view, show route
 * Non-reachable neighborhood bounds should only show special tooltip on hover, and not turn green

Closes #161 
Fixes #212

### DIFF
--- a/taui/configurations/default/messages.yml
+++ b/taui/configurations/default/messages.yml
@@ -31,6 +31,7 @@ Map:
     SetStart: Set start
     SetEnd: Set end
     ClearMarkers: Clear markers
+  Unreachable: cannot be reached by transit
 Dock:
   LocationLabel: From
   NetworkLabel: When
--- a/taui/src/components/application.js
+++ b/taui/src/components/application.js
@@ -32,6 +32,7 @@ type MapState = {
 type Props = {
   accessibility: number[][],
   actionLog: LogItems,
+  activeNeighborhoodBounds: any,
   data: {
     grids: string[],
     neighborhoodBounds: any,
--- a/taui/src/components/dock.js
+++ b/taui/src/components/dock.js
@@ -12,6 +12,7 @@ import NeighborhoodDetails from './neighborhood-details'
 import RouteCard from './route-card'
 
 type Props = {
+  activeNeighborhood: string,
   activeNetworkIndex: number,
   detailNeighborhood: any,
   endingOffset: number,
@@ -154,6 +155,7 @@ export default class Dock extends PureComponent<Props> {
   // Render list of neighborhoods
   neighborhoodsList (props) {
     const {
+      activeNeighborhood,
       changeUserProfile,
       neighborhoods,
       setActiveNeighborhood,
@@ -170,6 +172,7 @@ export default class Dock extends PureComponent<Props> {
     return (
       neighborhoods.map((neighborhood, index) =>
         <RouteCard
+          activeNeighborhood={activeNeighborhood}
           goToDetails={(e) => goToDetails(e, neighborhood)}
           index={index}
           isFavorite={userProfile.favorites &&
--- a/taui/src/components/draw-neighborhood-bounds.js
+++ b/taui/src/components/draw-neighborhood-bounds.js
@@ -5,7 +5,7 @@ import {LayerGroup} from 'react-leaflet'
 
 import {
   NEIGHBORHOOD_BOUNDS_STYLE,
-  NEIGHBORHOOD_HOVER_COLOR,
+  NEIGHBORHOOD_ROUTABLE_COLOR,
   NEIGHBORHOOD_NONROUTABLE_COLOR
 } from '../constants'
 
@@ -15,14 +15,6 @@ export default class DrawNeighborhoodBounds extends React.PureComponent {
   _key = 0
   _getKey () { return this._key++ }
 
-  hoverStyle = (feature) => {
-    return Object.assign({}, NEIGHBORHOOD_BOUNDS_STYLE, {
-      fillColor: feature.routable
-        ? NEIGHBORHOOD_HOVER_COLOR
-        : NEIGHBORHOOD_NONROUTABLE_COLOR
-    })
-  }
-
   getTooltip = (feature) => {
     if (!feature || !feature.properties) return ''
     const {routable, town} = feature.properties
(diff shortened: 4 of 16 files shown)


flibbertigibbet:
The flickering is because [react-leaflet GeoJSON layers do not update on data changes](https://react-leaflet.js.org/docs/en/components.html#geojson), and so have to be reloaded.

lederer:
Thanks for doing this! Will review this morning.

flibbertigibbet:
Shortly after we began work on this fork of `taui`, the upstream project [switched from Leaflet to Mapbox GL](https://github.com/conveyal/taui/commit/616aece4622589d42c13d92b7365750148e6344d#diff-b9cfc7f2cdf78a7f4b91a753d10865a2), and so doesn't use the Leaflet Vector Grid plugin that handles the layer loading here.

flibbertigibbet:
The data change in the list view causing the element to re-render is the active route change, that controls which neighborhood is currently selected to have its route drawn.

Regarding the neighborhood point marker at the route end, that was removed along with the last walk leg as a specific request from the client to not display either, done in #189, as described in the notes above. The client wants to obscure the neighborhood endpoint that was used for routing. If you want to revert #189, we should run that by the client.

lederer:
> The data change in the list view causing the element to re-render is the active route change, that controls which neighborhood is currently selected to have its route drawn.

Ok. Couple of ideas/questions:

- Are the neighborhoods and routes in the same GeoJSON layer? If so, is it possible to separate them into different layers and only rerender the route layer when the data changes?
- If that's not an option, could we try rendering routes only to the 3 listed neighborhoods, rather than to all routable neighborhoods. This way, the flicker would only happen when mousing in/out of 3 polys.

> Regarding the neighborhood point marker at the route end, that was removed along with the last walk leg as a specific request from the client to not display either, done in #189, as described in the notes above. 

Ok. I didn't realize they'd asked to remove the marker as well. In that case, I'll remove the green marker from the sidebar card in detail view, to eliminate the asymmetry.

flibbertigibbet:
I was able to work around the issue causing the map flash on route change by only referencing the selected neighborhood when the neighborhood bounds layer first loads; the neighborhood bounds layer cannot know which is the active neighborhood without redrawing whenever it changes. This happens to be fine, since the layer redraws on going to details when the active neighborhood gets set.

It might be worth testing out the app to make sure this change hasn't broken anything else.

flibbertigibbet:
The hover styling stopped updating on mousing out, but only in list view. I'll look into it.

ddohler:
This is a tremendous improvement, nice work! Besides the mouse-out behavior that you mentioned, I noticed one other thing amiss, which is that when you hover over cards in the list view, the polygon for the first card you hover over after loading a new page of results will stay highlighted forever, even after you hover over other cards or polygons.

lederer:
Wow!

Also found that hovering on the sidebar cards no longer changes the sidebar card marker or map marker to green.

flibbertigibbet:
Fixing mouse-out interactions re-introduces flashing/refreshing the layer. So to highlight bounds without flashing the whole bounds layer, I've changed it to instead display a separate layer for the active neighborhood bounds on top of the full bounds layer, at full opacity (otherwise the green blends with the purple of the layer beneath). Consequently, hover interactions in the detail view only show the tooltip, but in list view, highlight the bounds on hover.

image

flibbertigibbet:
After a couple of minutes of hovering around with the added layer, the site crashes, so this seems to have introduced a performance issue.

flibbertigibbet:
Debouncing setting the active neighborhood has reduced the performance load.

lederer:
Thanks for fixing. So much better. 👍 

Found #214 while testing.

PR #217, opened 2019-06-13
Bugfix/kak/fix detail user zoom#216

### LINKED ISSUE #216: Detail view zoom weirdness
On detail view, after zooming (in or out), the map automatically reverts to the default zoom level.

flibbertigibbet:
## Overview

Fix user zooming in detail view.

## Testing Instructions

 * Map should still zoom to fit route on going to details
 * Using the zoom control should behave as expected, particularly in detail view
 * Map should still zoom to fit all neighborhoods on returning to list

Fixes #216.

### DIFF
--- a/taui/package.json
+++ b/taui/package.json
@@ -72,7 +72,7 @@
   "devDependencies": {
     "core-js": "~2.6.9",
     "mastarm": "https://github.com/azavea/mastarm.git#dev",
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
I was able to reproduce the issue on Firefox. This is ready for another look.

PR #223, opened 2019-06-18
Feature/kak/modify weighting#178

### LINKED ISSUE #178: Modify sort weighting for neighborhoods
For all user settings for importance:
 - [ ] Add greater weight to travel time
 - [ ] Reduce weighting difference between moderate and low crime quintiles

flibbertigibbet:
## Overview

Modify neighborhood weighting algorithm to:
 - always give travel time (accessibility) extra weight over the other two variables
 - weight schools differently when schools are given maximum importance
 - treat the safest two quintiles for violent crime equally

### Demo

All profile options weighted equally, top transit results:

image

Default profile options weights (travel time 'important', other two 'somewhat'):

image

Schools 'very important', other two options 'not important':

image

### Notes

Travel time effectively always has a one higher importance rating than whatever it has assigned in the profile with this.

When schools are ranked as 'very important' (and the other two options aren't also 'very important'), this changes the algorithm to weight by percentile binned into quarters; for other school importance rankings, or when all have the same importance, schools are still ranked by quintile.

From the client:
 > IF Very Important strongly prioritize school districts in top 75 education percentile and above. No distinction within 75 and above. [I use percentiles here instead of quintiles to reflect the distribution of education quality in the state, which is concentrated more in the top 25% than in the top quintile]
> If important prioritize schools quintile 2 and above.
> Somewhat important prioritize schools in quintile 3 and above.
> Not important not weighted at all.

## Testing Instructions

 * Try different profile importance rankings for the three variables
 * Result sorting should vary as expected

Closes #178.

### DIFF
--- a/taui/src/selectors/neighborhoods-sorted-with-routes.js
+++ b/taui/src/selectors/neighborhoods-sorted-with-routes.js
@@ -26,6 +26,9 @@ const MAX_QUINTILE = 5
 const DEFAULT_EDUCATION_QUINTILE = 5
 const DEFAULT_CRIME_QUINTILE = 5
 
+// extra constant weighting always given to travel time over the other two factors
+const EXTRA_ACCESS_WEIGHT = 1
+
 export default createSelector(
   selectNeighborhoodRoutes,
   neighborhoodTravelTimes,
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
"Not important" settings are given zero weighting, except for travel time, which always has at least 1, and also except in the case the user sets all options to "not important," in which case crime and schools are weighted 1, and travel time 2.

ddohler:
> "Not important" settings are given zero weighting, except for travel time, which always has at least 1, and also except in the case the user sets all options to "not important," in which case crime and schools are weighted 1, and travel time 2.

That's right, I had forgotten that detail! I think the travel time weighting threw me off and I started thinking that everything had a minimum non-zero weight. Cool. :sunglasses:

lederer:
Rebased the color change branch onto this and fiddled around. I think we're in good shape for now. ✅

PR #224, opened 2019-06-18
Visual design improvements

lederer:
## Overview

A slew of visual improvements…

#### Colors
- New chart colors: More visible against white background. Less similar to other app colors. More spectrum-like.
- Dark gray transit route color in search mode (not detail mode), to minimize color chaos.

#### Map polygons
- Optimize opacity, stroke, fill for distinguishing routable and active neighborhoods without obscuring map data
- Don't render polygons for non-routable neighborhoods

#### Sidebar map markers
- Make search sidebar markers match on-map markers
- Replace detail view marker with polygon icon

#### Tooltips
- Make map tooltips look like sidebar tooltips
- But use a less salient treatment for non-routable neighborhoods
- Reposition map tooltips so they're less in the way

#### Copy
- "Your recommendations" – a more personalized search result header

### Demo

**Chart colors, Your recommendations**

Screen Shot 2019-06-18 at 1 43 22 PM

**Tooltips, Route colors**

Screen Shot 2019-06-18 at 1 43 48 PM

Screen Shot 2019-06-18 at 1 44 55 PM

Screen Shot 2019-06-18 at 1 44 04 PM

** Detail view sidebar neighborhood icon**

Screen Shot 2019-06-18 at 1 43 09 PM

## Testing Instructions

 * Review changes in sidebar and map in search and detail views


taui/src/components/draw-neighborhood-bounds.js
```diff
@@ -18,7 +17,12 @@ export default class DrawNeighborhoodBounds extends React.PureComponent {
   getTooltip = (feature) => {
     if (!feature || !feature.properties) return ''
     const {routable, town} = feature.properties
-    return routable ? town : town + ' ' + message('Map.Unreachable')
+    const tooltipEl = document.createElement('div')
```
flibbertigibbet: Why not use `React.createElement()` or similar react/JSX templating?

taui/src/components/draw-neighborhood-bounds.js
```diff
@@ -29,17 +33,16 @@ export default class DrawNeighborhoodBounds extends React.PureComponent {
         data={p.neighborhoods}
         idField='id'
         tooltip={(feature) => getTooltip(feature)}
+        tooltipClassName='map__tooltip'
```
flibbertigibbet: What is `tooltipClassName`? It doesn't seem to be getting read anywhere. The [property name recognized by react-tooltip](https://github.com/wwayne/react-tooltip/blob/a2e8af550ae750aae6a68d2fb3195e5ebc258190/README.md#options) is `className`. The values assigned to `tooltipClass` above are hard-coded.

lederer:
> Please move the JS-wrapped SVG elements out of the `components` directory to a new directory at the same level (`icons` maybe?)

Donezo.

taui/src/components/draw-neighborhood-bounds.js
lederer: I tried using JSX at first but it didn't work. I believe this is because the code in `getTooltip()` isn't called as part of the React rendering cycle. Its return value needs to be either a [string or an HTMLElement](https://leafletjs.com/reference-1.0.0.html#layer-bindtooltip), whereas `React.createElement()` would return a React element.

taui/src/components/draw-neighborhood-bounds.js
lederer: It's an undocumented prop in `react-leaflet-vectorgrid` that allows us to style the tooltip. The tooltips appear to be Leaflet tooltips rather than react-tooltips.

PR #233, opened 2019-06-20
Misc polish

lederer:
## Overview

- Distinct green/yellow/orange chart colors
- "Show transit times…" form header
- "Your recommended neighborhoods 1–3 of 56"
- Tooltip styles
- "School choice" style
- Copy changes per client

### Demo

Screen Shot 2019-06-20 at 4 15 50 PM

Screen Shot 2019-06-20 at 4 16 02 PM

Screen Shot 2019-06-20 at 4 16 27 PM

Screen Shot 2019-06-20 at 4 16 08 PM

PR #235, opened 2019-06-21
Pass count instead of all neighborhoods

flibbertigibbet:
## Overview

Pass the total count of neighborhoods to the `Dock` component, instead of all neighborhoods.

## Testing Instructions

 * Total record counter should still work

Connects #233.


flibbertigibbet:
Trivial; merging.

PR #237, opened 2019-06-21
Fix multiple tooltips bug

### LINKED ISSUE #230: Multiple tooltips showing at once
Mousing along the sidebar causes multiple tooltips to render at once.

Ideal solution: Triggering a tooltip automatically hides any currently visible tooltips.

If that's not possible, we may be able to optimize the `delayHide` and `delayShow` values to minimize the likelihood of this happening.

Screen Shot 2019-06-20 at 12 45 44 PM

lederer:
## Overview

Dialed down the hide delay and changed to one tooltip per card instead of per meter. 

This appears to have largely solved the problem.

~However it introduces a new bug. In search view, when mousing into the tooltip itself, the last card in the list gets selected, as if you've hovered on that card. I don't see any simple way of stopping the propagation of that hover event. Fortunately, clicking the link in the tooltip does not trigger a click on the last card. I think this is a net win, given how bad the multiple tooltips problem was. I'll open a new issue for this new bug.~

### Notes

I initially tried manually hiding all tooltips via `ReactTooltip.hide()` `onMouseEnter` but it didn't work.

Cosmetic change: Tooltips now render to the right, over the map, so as not to occlude other meters.

## Testing Instructions

 * Hover over many meters on search and detail pages, and mouse into tooltips.
 * Confirm that multiple tooltips rarely, if ever, get rendered at once.

Resolves #230

### DIFF
--- a/taui/src/components/meter.js
+++ b/taui/src/components/meter.js
@@ -16,7 +16,7 @@ export default function Meter ({
   const tier = getTier(percentage)
 
   return (
-    <div className='meter' data-tip={tooltip}>
+    <div className='meter' data-tip={tooltip} data-for={`tooltip-${id}`}>
       <svg
         className='meter__chart'
         viewBox={`0 0 ${width} ${height}`}
--- a/taui/src/components/meter.js
+++ b/taui/src/components/meter.js
@@ -1,7 +1,4 @@
 // @flow
-import ReactTooltip from 'react-tooltip'
-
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
The bug is likely the result of having removed the `id`s from the elements. Each element with a tooltip needs a unique `id`, and the `data-for` needs to reference its matching `id`.

lederer:
You're right, that appears to have fixed it. 🎉

PR #244, opened 2019-06-25
Update data to fix one row

flibbertigibbet:
## Overview

Fix crime percentage for zip code 02025 (Cohasset).

### Demo

image


flibbertigibbet:
Trivial data fix; merging.

PR #274, opened 2019-11-04
Disable input field autocomplete

### LINKED ISSUE #271: Disable browser autofill for voucher number
Do not auto-suggest voucher numbers on the voucher number text field in the search page (or login page, if used for client account login).

flibbertigibbet:
## Overview

Tell browsers to not autofill input form fields to search for a profile or to fill out a profile.

## Testing Instructions

 * Go to account search page
 * Click into voucher number input field to give it focus
 * No previously entered voucher numbers should be suggested by the browser

Closes #271.

### DIFF
--- a/taui/src/components/edit-profile.js
+++ b/taui/src/components/edit-profile.js
@@ -347,6 +347,7 @@ export default class EditProfile extends PureComponent<Props> {
               type='radio'
               onChange={(e) => setPrimaryAddress(index, e)}
               checked={!!destination.primary}
+              autoComplete='off'
             />
           </div>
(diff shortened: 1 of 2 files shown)


taui/src/components/edit-profile.js
```diff
@@ -486,6 +487,7 @@ export default class EditProfile extends PureComponent<Props> {
                 type='text'
                 onChange={(e) => changeField('headOfHousehold', e.currentTarget.value)}
                 defaultValue={headOfHousehold || ''}
+                autoComplete='off'
```
ddohler: Good call changing all of these as well -- counselors won't want to auto-fill these and individuals shouldn't need to.

PR #276, opened 2019-11-19
Style client account creation form

lederer:
## Overview

Style the client account creation form added in #275 

#### Empty form

Screen Shot 2019-11-19 at 1 58 27 PM

#### After submission

Screen Shot 2019-11-19 at 1 58 16 PM

### Notes

Refactored the DOM to move the form into its own section. Changed classnames. Added and edited some messages.

### DIFF
--- a/taui/configurations/default/messages.yml
+++ b/taui/configurations/default/messages.yml
@@ -120,9 +120,11 @@ Profile:
   DeleteAddress: Delete this address
   DeletePrimaryAddressError: Cannot delete primary destination. Set another as the primary first.
   Cancel: Cancel
-  ClientEmail: Client e-mail address
-  ClientEmailError: Please enter a valid e-mail address for the client.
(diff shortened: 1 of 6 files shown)


lederer:
Hold off for a moment. I neglected to move the error message.

lederer:
I moved the page's error message to between the main section and the new account creation section. The bottom of a page is an unusual location for error messages, but this will do until we have a sophisticated error system. 

In particular, this should suffice because its middling position allows it to apply to both the main section above it and the account section below it, and because it will always appear near the button clicked to submit the form. If placed at the top of the page, it can be easily missed because it would appear above the top fold when the form is submitted on relatively short screens.

Screen Shot 2019-11-19 at 2 17 21 PM

lederer:
Ok, feel free to review anytime.

lederer:
Doh. Yeah that should not be there when the account exists. I pushed a fix. Can you confirm it works correctly now?

PR #289, opened 2019-12-11
Feature/kak/profile email registration#286

### LINKED ISSUE #286: Move email to profile and modify account registration
Modify the profile page to make the user email address a field on the profile that is only editable by counselors until the email address has been first set (and so user account has been created).

Remove form for counselors to invite users or resend invite. Instead, show button next to email address in the profile to resend the invite if the user hasn't logged in yet for the first time. When a counselor saves a profile, check if the profile is new or the email has been set where it was previously unset, and if so, create a user account for the voucher number and entered email address.

flibbertigibbet:
## Overview

Make the client email address a profile field and handle client account creation on profile save. Also stores client account state on profile: whether or not a client account exists for the email, and whether on not that client has logged in for the first time.

### Demo

Counselors can edit the email on a profile until a Cognito account has been successfully created for the email. After that, the email cannot be edited.

Once a counselor has saved a profile with a new email, a Cognito account gets created, but is not confirmed until the first time the client logs in. Until the first client login, the counselor can resend the invite from the profile:

image

After the client has logged in for the first time, confirming their Cognito account, counselors can no longer resend the sign-up invitation email with temporary password:
image

Clients get a read-only view of their profile email. If a client deletes their profile, the site will log them out, then the next time the client logs in, create a new profile with defaults with the client's voucher number and email address pre-set:
image

If a counselor tries to create a profile with an email already associated with a Cognito account, but for a different voucher number than that of the new profile, a specific save error message tells the client what voucher number is assigned to the Cognito account for that email:

image

If a counselor creates a profile with an email already associated with a Congito account that has the same voucher number as the new profile, the counselor will be able to successfully save the new profile, and viewing the profile again will reflect whether the client has confirmed their account yet. As new profiles can only be created after searching for the voucher number, this should prevent counselors accidentally overwriting an existing profile.

### Notes

Counselors have no way to delete client Cognito accounts once they have been created, and client Cognito accounts are permanently associated with the voucher number assigned at the time the account was created, so if a counselor mistakenly creates a profile associating an email address with the wrong voucher number, there's no way to fix it. Adding a Cognito account deletion feature would address the issue, but would be non-trivial to implement.

## Testing Instructions

 * Log in as a counselor
 * Test creating, modifying, and deleting user profiles
 * Email field should only be editable until client account has been made
 * Counselors should be able to resend invites only until client first logs in
 * Appropriate error should show on attempting to create a profile with an email already associated with a different profile, whether or not the client with that email has logged in for the first time yet
 * Log in as a client
 * Counselor-created profile for client should be found if it exists
 * Email should always be set and be read-only for the client
 * If client does not have a profile (as when deleting their own), a new one should be created on next client log-in with the email and voucher number preset

Closes #286


taui/src/components/edit-profile.js
```diff
@@ -198,80 +243,138 @@ export default class EditProfile extends PureComponent<Props> {
...
+        console.error('API call to create user failed')
+        console.error(error)
+        this.setState({errorMessage: message('Profile.CreateClientAccountError')})
+        reject(error)
+      })
     })
   }
 
-  deleteProfile (key, event) {
+  deleteProfile (key: string, isCounselor: boolean, event) {
```
ddohler: One behavior I noticed (not sure if it's an issue, necessarily) is that once a profile is deleted, a counselor can re-create a profile for the same voucher number, but with a different email address. If the second email address signs in, then the result is two Cognito logins being associated with the same voucher number. It doesn't seem to cause any major harm -- both emails can log in and see the same profile info, which I guess would probably be the desired behavior in this case. However, the email address displayed in the profile will only ever be the most recent email address that was associated with the voucher, so you can log in with the "older" email, and the email displayed in the profile page will be the "newer" one.

I'm not sure whether this is actually a problem, since it seems like a fairly rare case, and both emails can still log in and see the voucher info, so it wouldn't allow someone to get locked out. This also enables counselors to change the email address associated with an account in case they make a mistake, so it might be a useful escape hatch. Do we have enough info from the client to know whether this matches what they wanted?

flibbertigibbet: I modified the `/profiles` endpoint to delete the old login account in the case that a client logs in without a profile, and another profile exists for their voucher number but a different email address.

taui/src/components/edit-profile.js
```diff
@@ -198,80 +243,138 @@ export default class EditProfile extends PureComponent<Props> {
...
+      console.log('Create client user Cognito account for ' + clientEmail)
+      // Also set `response: true` in addition to `body` to get full response,
+      // instead of just data (AWS library uses Axios).
+      API.post(AMPLIFY_API_NAME, '/clients', {
+        body: {
+          email: clientEmail,
+          voucher: key
+        }
+      }).then(response => {
+        if (response.error) {
```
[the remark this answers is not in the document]
flibbertigibbet: I agree (and left a `TODO` for that), but will defer refactoring this file to later, since this PR is blocking styling work at the moment.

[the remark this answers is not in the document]
flibbertigibbet: Opened #290

taui/src/components/with-authenticator.js
```diff
@@ -252,7 +284,12 @@ export default function withAuthenticator (Comp, includeGreetings = false,
       } else if (voucher) {
         // attempt to go to client profile directly
         this.goToClientProfile(voucher, email).then(succeeded => {
-          this.setState({authState: state, authData: data})
+          if (succeeded) {
+            this.setState({authState: state, authData: data})
+          } else {
+            console.warn('Failed to load client profile; logging out')
+            this.logoutEcholocator(this.state.authData)
```
ddohler: When I tested out the "invite->delete->invite different->login with old" workflow, this issued the correct message to the console, but I wasn't redirected to the login screen for some reason. I didn't see any console errors, however.

flibbertigibbet: How about I make a separate issue for that, so we can unblock design?

ddohler: :+1: Works for me, this is not something that's going to come up frequently.

flibbertigibbet: Created #291

PR #293, opened 2019-12-13
Email field, ECC label, Dropdown fix

### LINKED ISSUE #285: "From" dropdown ambiguous behavior
The "from" dropdown has a couple of issues:

- When focused, a blinking insertion point appears in the box, but the content isn't editable. Users are getting the false impression they can type into it.
- Opening the dropdown shows the first option as highlighted, no matter what. If you hover on a different option, the highlight doesn't change. If you select a different option and reopen the the dropdown, the first option remains highlighted.

echolocator

lederer:
## Overview

Fix #285, problematic highlighting behavior in "from" dropdown.

Replace ECC key/value UI with a label beneath neighborhood name.

Style profile email field that was moved into the main profile form in #289 
- Move field to just below name field.
- Clean up copy.
- Remove "invitation sent" message (It appears only fleetingly upon save, since the page immediately changes to the map view, and the presence of the resend button makes it redundant.)
- Move error message to top of form.

### Demo

image

image

Screen Shot 2019-12-13 at 2 18 52 PM

Screen Shot 2019-12-13 at 2 28 17 PM

Screen Shot 2019-12-13 at 2 18 29 PM

Screen Shot 2019-12-13 at 2 21 18 PM

Resolves #285


taui/src/sass/main.scss
```diff
@@ -15,7 +15,6 @@
 '06_components/account-profile.scss',
 '06_components/account-search.scss',
 '06_components/account-list.scss',
-'06_components/convert-to-account',
```
flibbertigibbet: :sparkles:

taui/src/components/neighborhood-details.js
```diff
@@ -241,7 +241,18 @@ export default class NeighborhoodDetails extends PureComponent<Props> {
               type={isFavorite ? 'star' : 'star-o'}
               onClick={(e) => setFavorite(id, userProfile, changeUserProfile)}
             />
-            <div className='neighborhood-details__name'>{town} &ndash; {id}</div>
+            <div className='neighborhood-details__name'>
+              {town} &ndash; {id}
+              {ecc &&
+                <div
+                  className='neighborhood-details__ecc'
+                  data-tip={message('Tooltips.ExpandedChoice')}
```
flibbertigibbet: These tooltips are still showing off the right end of the card, although the label is now left-aligned. Was that intentional?

lederer: Fixed!

lederer:
> would it make sense/not be too much effort to scroll to top of page on failed save now, so it's clear that it didn't actually save? Or alternatively, reload the page on save, so it's reset to the top of page where the error message is?

Good idea. I'll take a look.

lederer:
Profile now scrolls to top when there's an error on save. More specifically, it scrolls to top when there's an error that doesn't match the previous state's error. One implication is that if there is an error, and the user attempts to save the profile with the same error, it won't scroll to the top. I think we should consider this an edge case. Fixing it would require more complexity.

lederer:
Hold off on approving. Turns out `window.scroll()` isn’t fully supported. I will fix on Monday.

lederer:
Ok force pushed a change that should be compatible with supported browsers. Those that support the `scroll-behavior` CSS property will scroll smoothly to the top on error. Others will jump to top.

taui/src/components/edit-profile.js
```diff
@@ -79,6 +79,12 @@ export default class EditProfile extends PureComponent<Props> {
     }
   }
 
+  componentDidUpdate (prevProps, prevState) {
+    if (this.state.errorMessage !== prevState.errorMessage) {
```
flibbertigibbet: This will also scroll to the top on the first character change when the user goes to edit the problem field after getting the error. Changing this to `if (this.state.errorMessage) {` I think should do what we want.

lederer: Ah, I see. That seems to work.

PR #298, opened 2020-01-14
Feature/kak/modify rankings

### LINKED ISSUE #297: Modify rankings
Modify neighborhood rankings as described below (from client), except for the travel time cutoffs.

## Travel time
IF Very Important should act as a more severe constraint on neighborhood selections. ~~I.e. Very important should really only be showing all commutes within 45 minutes first (subject to other constraints, below).~~

~~If Important show commutes within 60 minutes first.~~

~~Somewhat important – show travel times within 90 minutes first.~~

~~Not important – travel time not weighted.~~

## Schools
If Very Important strongly prioritize school districts in top 75 education percentile and above. No distinction within 75 and above. [I use percentiles here instead of quintiles to reflect the distribution of education quality in the state, which is concentrated more in the top 25% than in the top quintile]

If important prioritize schools quintile 2 and above.

Somewhat important prioritize schools in quintile 3 and above. 

Not important not weighted at all. 

## Public Safety
In general: no distinction between those communities of violent crime quintiles 1-2 – all weighted equally positively. Quintile 3 communities should be rated almost equally well. This is to reflect the fact that people are much more sensitive to high crime than they are to moderate crime neighborhoods. 

If very important 
All quintile 5 should show up last, overriding all other variables. Quntile 4 should show up second to last.
Quintile 1-2 should show up first all other things being equal, but quintile 3 also can show up just as high if it meets other criteria very well (for example, if it within the commute constraint or school quality).

If important 
Same as very important except no distinction between quintiles 1-3 (all treated equally positively).

Somewhat important 
Same as very important except no distinction between quintiles 1-4 (all treated equally positively).

Not important 
Not weighted.

flibbertigibbet:
## Overview

Modify neighborhood ranking logic to:
 - increase weight for transit time
 - weight equally category buckets for school and crime when certain criteria are met

### Notes

Transit times have been weighted more heavily rather than implementing the cutoff categories requested, as they don't make much sense with a weighted algorithm. The safety and school ranking changes are implemented as described; the changes were essentially to group ranking values depending on the combination of value and assigned importance to the category.

## Testing Instructions

 * View results for `VERIFY1` and `VERIFY2` test profiles
 * Results should be ordered as expected

Closes #297

### DIFF
--- a/taui/src/selectors/neighborhoods-sorted-with-routes.js
+++ b/taui/src/selectors/neighborhoods-sorted-with-routes.js
@@ -37,6 +37,7 @@ const DEFAULT_CRIME_QUINTILE = 5
 // extra constant weighting always given to travel time over the other two factors
 const EXTRA_ACCESS_WEIGHT = 2
 
+/* eslint complexity: 0 */
 export default createSelector(
   selectNeighborhoodRoutes,
   neighborhoodTravelTimes,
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
`MAX_TRAVEL_TIME` is an absolute cutoff; anything outside that gets omitted from the results list completely. The requested change was rather to group travel time windows into buckets for sorting results (not filtering them out). That isn't really compatible with the concept of a weighted algorithm, which allows factors to interact with each other; what was requested is to always sort items within a travel time bucket above a second travel time bucket, irrespective of the other factors (schools and safety).

I instead thought to go with adding extra weight to travel time in the case the profile is set to "very important" for accessibility, to keep using weighting while being in line with the spirit of the first sentence of the change request description: "If Very Important should act as a more severe constraint on neighborhood selections."

I'd made a mistake in adding that extra weight, though: good catch! I'd missed applying an offset, and had put the added weight when the profile was set to "important" instead of "very important." I've added an extra constant to aid readability and to help avoid that issue in future.

The weight vs profile setting values are now:
6 - 4 profile
4 - 3
3 - 2
2 - 1

(The weight settings are zero-based while the profile settings start at 1; a constant +2 is added to all accessibility weights, plus an extra 1 when "very important".)

We could modify the travel time values based on the buckets given to always be the same within a bucket, but since there's only two buckets specified in the document per profile setting, that would lose a significant amount of detail (i.e., everything within 45 minutes would be treated as equally accessible when accessibility is "very important") so I'm holding off on making such a change for now, as I doubt that is what was intended.

PR #316, opened 2020-03-02
Feature/kak/update data#288

### LINKED ISSUE #288: Update data source
Update data in `neighborhoods.csv` and `neighborhood_descriptions.csv`, verify provided data matches expected format, and rerun processing scripts.

~~Blocked awaiting data.~~

flibbertigibbet:
## Overview

Update neighborhood data using source spreadsheet dated February 18, 2020.

## Notes

I modified the provided spreadsheet to conform to the expected format and to fix an off-by-one error in one row caused by a semicolon, then ran the `update_data.sh` script, which generated the changes to the other CSV files and the JSON files here.

## Testing Instructions

 * Site should still function as expected
 * Data should match what was provided in the spreadsheet

Closes #288


flibbertigibbet:
Good find! It looks like Abington also returns no results on staging, and that it is due to results for that area missing from S3 (there are console errors for the requests). We should let the client know they're missing travel time coverage for that area.

Whether a neighborhood shows in the results list (which uses the JSON modified here) is unrelated to whether that neighborhood finds results when the origin address is in the neighborhood (which comes from the pre-generated travel time files on S3).

flibbertigibbet:
I've updated this PR to add education data for Boston zip codes, which came from a later spreadsheet. @ddohler could you please take a look to see if it still looks okay? (Education data for Boston zip codes is still ignored in the sorting logic; that will be changed in a later task.)

PR #322, opened 2020-03-20
Fix overweighting crime

### LINKED ISSUE #321: Safety overweighted when very important
Neighborhood scoring appears to be off when safety is 'very important'.

flibbertigibbet:
## Overview

Fixes a neighborhood weighting/sorting issue when crime is marked "very important."

Also makes some minor package updates.

## Demo

Before:

echolocator_sorting_before

After:

echolocator_sorting_after

## Testing Instructions

 * Add profile address 1050 Adams Street, Dorchester Center (or use profile `22222222`)
 * Sorting should behave differently than on staging
 
Fixes #321


flibbertigibbet:
I do see variation with settings, such as setting schools to 'somewhat' and leaving the other two at 'very' important. I see variation in results 3-6 between 'somewhat important' and 'very important' transit when schools are 'important' and safety 'very important'. I think the variation in commute time settings is more subtle now due to the added weighting put on commute time in the last batch of changes.

flibbertigibbet:
Good catch, I've modified the multiplier to 0.9.

Issue #359, opened 2021-10-13
Allow public signups with manual verification


ddohler:
BHA wants to allow users to sign up for ECHO without having to be added by an administrator, but they want to still require administrator approval for any new users. So this is a multi-part task:

1. Enable self-service signup to the public. I don't exactly know where to do this but BHA has asked us to enable / disable this feature a few times in the past so I believe it's relatively easy and there should be pull requests showing how we've done it in the past.
2. Require administrator approval for new signups. This will require adjusting [Cognito settings](https://docs.aws.amazon.com/cognito/latest/developerguide/signing-up-users-in-your-app.html#signing-up-users-in-your-app-and-confirming-them-as-admin). This can be done manually; the second phase of the project will migrate away from Cognito. This should be done for the staging Cognito pool _only_ at this point.
3. Edit #361 to include instructions on how to make the necessary changes to the production Cognito pool.
4. There is some additional logic that happens on login to associate the new user with a voucher number. That should be investigated to confirm that it still works.
5. Check and see whether there is a way to forcibly convert usernames to lowercase on signup. This is known to be incorrect behavior (because email addresses may, technically, be case sensitive), but it's a stopgap to address a certain class of problems encountered by users until we can add Django-based authentication.

ddohler:
I added a timebox label to this -- I ran into some issues while testing the staging deploy that make me think that the staging user authentication machinery may not be set up entirely correctly, which might make it impossible to test part 4 without first fixing the authentication more generally. If that happens, spend the rest of the timebox on the card gathering as much information as possible and then we'll regroup and make another card to finish it off (hopefully).

ddohler:
After discussing with the client, we're going to punt this until we do the Django authentication implementation. There was some investigation that had to be done to get to that decision, so I'm going to close this and open a new card when we get to Phase II.

PR #365, opened 2021-10-13
Add info about demo site to deployment README

### LINKED ISSUE #350: Roll default branch back to most recent Azavea-contributed commit
We need to preserve existing work so that we can consult it but we don't necessarily want to build directly off of all of it, so we're going to rewind to the most recent Azavea commit and then cherry-pick or copy as necessary back to the default branch.

Tasks:
- Check out a new branch with a descriptive name with the current HEAD of the default branch (`develop`) and push this branch to GitHub.
- Revert `develop` to the most recent Azavea-contributed commit that precedes all non-Azavea commits. This could either be done by reverting the non-Azavea merge commits, or by forcing a reset to the appropriate Azavea commit. I think that reverting the non-Azavea merge commits would be preferable, but doing a force reset back to an older Azavea commit is acceptable as well if reverting is difficult for some reason.

Notes:
- This is likely to be intertwined with #351 because we don't want an alteration to the `develop` branch to accidentally trigger a CI build that might clobber the demo site. So we should be confident that CI won't clobber the demo / old staging site before proceeding with this.

### LINKED ISSUE #351: Resuscitate staging environment
The staging environment hasn't been rebuilt for quite a while, and we know that some of the AWS resources it used to rely on were erroneously deleted. The goal of this issue is to get CI functional once again, _without_ impacting the functionality of [the demo site](https://github.com/azavea/echo-locator/issues/349). This will likely require spinning up a new set of AWS resources so that the staging site can be deployed without overwriting the demo site.

Acceptance criteria:
- The contents of [the demo site](https://github.com/azavea/echo-locator/issues/349) are not affected by CI.
  - This may require targeting the staging build to output static assets to a different location
- Accessing staging.echosearch.org works and you can successfully log in and set up your profile (you will likely need to use the AWS Cognito console to create yourself a user).
- Changes merged to `develop` trigger a CI build and are deployed to staging when it is successful.

Resources:
- The existing deployment infrastructure uses Terraform and AWS Amplify: https://github.com/azavea/echo-locator/blob/develop/deployment/README.md
- The existing CI uses Travis: 
  - https://github.com/azavea/echo-locator/blob/develop/.travis.yml
  - https://app.travis-ci.com/github/azavea/echo-locator

Notes:
- The client is hoping to switch to echosearch.org as the root domain. We haven't committed to this, but if it's equivalently complex to continuing to use echolocator.org, then we should switch it. If it's hard to switch the root domain for some reason we can use `staging.echolocator.org` as a fallback.

ddohler:
## Overview

Rolls the `develop` branch back to approximately where we left off a few years ago, and moves changes that have happened since then over onto a `demo` branch for future reference.

### Notes

- This PR is mostly symbolic; the actual work is already done, but I did include a commit to add some explanatory text about the new demo site.
- Creating and logging in with a counselor account works just fine, but I haven't managed to get things working with a non-counselor (i.e. voucher) account yet, and it is failing in a way that makes me suspect that the staging authentication machinery may not have been reconstituted completely. However, a counselor account will be sufficient for most of the changes we need to make at this point, so I'm leaving further investigation of the login issue for #359 and if we can't solve it then we can at least decide on an approach.

## Testing Instructions

 * Log in to https://demo.echosearch.org using the "BHA ECHO Staging" credentials. Search for a voucher (`11111111` is one that exists). Confirm that the site seems to work.
 * Perform the same steps on https://staging.echolocator.org -- the site should also work, but should be noticeably different from `demo.echosearch.org` in terms of features.
 * Verify that this branch is built by Travis and that it's successful.
 * Inspect the logs for [the most recent `develop` build on Travis](https://app.travis-ci.com/github/azavea/echo-locator/builds/239763436) and confirm that the deploy succeeded.
 * Use the GitHub interface to inspect the commit history for the `demo` and `develop` branches and confirm that `develop` has only a limited amount of non-Azavea code (which was necessary to get logins working), while `demo` has quite a bit.

Resolves #351 , Resolves #350

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -4,6 +4,7 @@
 * [Terraform](#terraform)
 * [Taui](#taui)
 * [Amplify](#amplify)
+* [Demo site](#demo-site)
 
 ## AWS Credentials
 
@@ -100,3 +101,14 @@ $ cp deployment/amplify/staging/aws-exports.js taui/src/aws-exports.js
 
 (Note that CI will properly bundle the appropriate Amplify configuration file
 during deployment depending on the environment.)
+
+## Demo site
+There is a demo site deployed to demo.echosearch.org.
+
+Its infrastructure was created using the Terraform and Amplify methods
+described above, but it is not connected to CI; the functionality of the site
+was copied over from the staging site before returning to active CI use.
+
+It uses the same Cognito assets as the staging site, so this infrastructure
+should be left in place even after the staging and production sites are shifted
+to a Django authentication backend.


KlaasH (approved):
:+1: Looks good!
All the testing items check out.
"quite a bit" is right--[the comparison view](https://github.com/azavea/echo-locator/compare/demo) says "301,934 additions and 2,031 deletions."

PR #367, opened 2021-10-28
Custom signin error messages

### LINKED ISSUE #360: Adjust signin error text
The application's usernames are currently case-sensitive. Until we switch to Django for authentication, we can't make them case insensitive, so instead we should update the login failure message to explain to users that usernames are case sensitive.

ddohler:
## Overview

Adds custom error messages to the sign-in page so that users are reminded that usernames are case-sensitive.

### Demo

Screenshot_2021-10-28_12-16-43

### Notes

This relies on the error message from the Cognito API exactly matching the keys in the mapping of overridden error messages. The Amplify source code from which this was adapted uses regular expressions in an apparent attempt to be more robust to API changes, but at the cost of significant additional complexity (it turns the one-step lookup here into a loop plus a lookup because you have to check for a match against all the regexes, and then associate a key with each regex to map to the message that you want to display). The additional complexity of adding a regex didn't seem worth it to me because without guarantees about what the format of the message will be, there's only so flexible you can be without losing the ability to distinguish between different error messages, so it'll always be possible for API changes to cause problems. So this keeps the logic simple (and brittle).

## Testing Instructions

 * Make a small change to the `UserNotFound` or `BadUsernameOrPassword` messages in `messages.yml` 
 * Fire up `./scripts/server` (you may need to restart it if it was already running when you changed `messages.yml`).
 * Try to log in with bad credentials and confirm you see your changes reflected in the error message.

Resolves #360

### DIFF
--- a/taui/configurations/default/messages.yml
+++ b/taui/configurations/default/messages.yml
@@ -99,6 +99,8 @@ SignIn:
   Anonymous: Continue without an account
   AnonymousExplanation: Don't have an account?
   Greeting: Improving Housing Search for Voucher Families
+  UserNotFound: Could not find that username as entered. Usernames are case-sensitive.
+  BadUsernameOrPassword: Username or password was incorrect. Both usernames and passwords are case-sensitive.
 Header:
   New: New search
   Edit: Edit profile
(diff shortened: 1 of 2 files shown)


taui/configurations/default/messages.yml
```diff
@@ -99,6 +99,8 @@ SignIn:
   Anonymous: Continue without an account
   AnonymousExplanation: Don't have an account?
   Greeting: Improving Housing Search for Voucher Families
+  UserNotFound: Could not find that username as entered. Usernames are case-sensitive.
+  BadUsernameOrPassword: Username or password was incorrect. Both usernames and passwords are case-sensitive.
```
ddohler, line 103: @stephwall could you run this copy by the client and ask them if they'd like it to say something different?

KlaasH (approved):
:+1: Looks good to me. Confirmed the basic message on `develop`, the better message on the branch, and that it changed when I modified `messages.yml`.

PR #370, opened 2021-11-02
Replace old logo with new one

### LINKED ISSUE #353: Change logo
Switch to the new ECHO logo.

@stephwall I recall that BHA said they'd send us their logo files for us to choose the appropriate one, have they done that?

ktohalloran:
## Overview

The log in page and header were attempting to display the old logo. This PR adds the new logo and deletes the old one. 

### Demo
Screen Shot 2021-11-02 at 3 34 00 PM

Screen Shot 2021-11-02 at 3 33 51 PM

### Notes

Prior to this PR, the old logo source path was `'taui/assets/BHAlogo.png'`. I changed the path to direct to where the old logo actually was in `'taui/src/img/'` as the repo ignores `'assets'` and I therefore didn't understand why that would be a good place to store images. I thought maybe `assets` related to deployment and therefore my not being able to see the old logo at the given path could have been related to the fact that I wasn't connected to AWS, so I set up the AWS profile but am still not seeing the original logo. So I went with changing the path, but please let me know if I'm just missing something about Taui, AWS, or the purpose of `assets`.

This PR does not do any styling around the logos, just changes out the old for the new. It also just re-uses the names given to the new logo files rather than naming them to something less wordy. In both cases, I wasn't quite sure what the ideal was so just went with what was already there.

## Testing Instructions

 * Run `./scripts/server`
 * Navigate to the log in page; you should see the new logo at the top of the screen
 * Log in; you should see the new logo in the top left corner

Resolves #353


taui/src/components/custom-sign-in.js
```diff
@@ -26,7 +26,7 @@ class SignInHeader extends React.Component {
     return (
       <header className='auth-screen__header auth-header'>
         <h2 className='auth-header__agency'>
-          <img className='auth-header__logo' src='assets/BHAlogo.png' alt='' />
+          <img className='auth-header__logo' src='src/img/echo_combined_logo_STACKED_fullcolor.svg' alt='' />
```
ddohler:
```suggestion
          <img className='auth-header__logo' src='assets/echo_combined_logo_STACKED_fullcolor.svg' alt='' />
```
And this file will also need to be added to settings.yml, similar to the one above.

taui/yarn.lock
```diff
@@ -3888,14 +3888,14 @@ debug@4, debug@^4.0.0, debug@^4.0.1, debug@^4.1.0, debug@^4.1.1, debug@~4.1.1:
   dependencies:
     ms "^2.1.1"
 
-debug@^3.1.0, debug@^3.2.6:
+debug@^3.1.0:
```
ddohler, line 3891: I'm not entirely sure why the lockfile is getting changed, but I'm seeing the same behavior as you, so it's a good step to commit this so that it doesn't keep changing on us. However, this should be a separate commit from the other files because it's unrelated to the logo change.

taui/src/components/custom-header-bar.js
```diff
@@ -58,7 +58,7 @@ export default class CustomHeaderBar extends Greetings {
     return (
       <header className='app-header'>
         <div className='app-header__brand'>
-          <img className='app-header__logo' src='assets/BHAlogo.png' alt='' />
+          <img className='app-header__logo' src='src/img/echo_combined_logo_fullcolor.svg' alt='' />
```
ddohler: Okay! So I had to do some digging to figure out what was happening here, and the short version is that this does indeed need to remain `assets/echo_combined_logo_fullcolor.svg`, but the project doesn't seem to be configured exactly right as things stand. Buckle up for the long version:

Ordinarily, when you encounter an app where the path at which images (aka static assets) are accessed in production (`taui/assets/` in this case) is different from where they live in the source tree (`taui/src/img/` in this case), that's an indication that the project **bundler** (usually we use Webpack, but in this case it's Mastarm, which is a custom project) is processing the static files somehow. That's also why the `/assets` folder is ignored by Git because it's not exactly part of the source of project -- all the files in `/assets` are generated at build time from other files in the source tree.

So the way I would (normally) expect to change an image under such circumstances is to

1. Figure out where the images are stored in the source tree, and drop in the new file;
2. Update the `src` attribute in any relevant `img` tags, but keep them pointing at the static assets folder; and
3. Proceed on your way and let the bundler handle copying from the source tree to the static assets folder.

Unfortunately, that doesn't seem to be working with this project, even on `develop`. I checked out `develop`, deleted the `assets/` folder, and ran `./scripts/server`. I expected `assets/` to be recreated with all the same files in it, and it was, mostly, but the logo was missing. I searched the project directory for the phrase "assets" and noticed that we have set up a separate command to handle copying from `/src/img`  in [./scripts/cipublish](https://github.com/azavea/echo-locator/blob/develop/scripts/cipublish#L45) for when we're deploying. So that implies to me that `assets/` is indeed the correct path, because it's how things are structured for deployment, but that image copying on this project isn't quite working the way I would expect at first glance because it doesn't seem to be copying properly in development.

I did some further digging into Mastarm itself and it looks like it [uses a settings.yml file](https://github.com/azavea/mastarm#build) to determine which files to copy into `assets/` when bundling. I searched for all files named `settings.yml` in the project directory [and found that the settings file doesn't specify including the logo](https://github.com/azavea/echo-locator/blob/develop/taui/configurations/default/settings.yml), but does specify the other files that are showing up in `assets/`. This is a little surprising. Normally, when using a bundler, you create some kind of general rule for handling all images when setting up the project, and once you've got that, you can add, remove, or modify images however you want, but I tried adding wildcard rules to `settings.yml` to cover all SVGs, and Mastarm got mad at me. However, I was able to add a specific rule for this one SVG file, and that did cause it to get copied into `assets/`. That's not ideal because it means any time we want to add or remove an image we need to do it in two different places, but I don't see any other viable route to get things working aside from manually copying the logo file into `assets/`.

This does raise an interesting question though--I had a copy of `BHAlogo.png` in my `assets/` folder from past development on this project, but then I wasn't able to get it to reappear once I deleted the `assets/` folder -- did the logo appear for you at all the first time you set up the project? If it did, that may imply that there's another step somewhere that makes things work on development that I've missed.

Anyway, as to what needs to happen here, I think there are two things to do to get this working for the time being, and a third for later:
1. Switch this path to `assets/echo_combined_logo_fullcolor.svg` because that's what will be expected once this is deployed. If we don't do this, then things will work in local development but probably not on staging or production.
2. Add `src/img/echo_combined_logo_fullcolor.svg` to [settings.yml](https://github.com/azavea/echo-locator/blob/develop/taui/configurations/default/settings.yml) following the pattern for the other files that are already there. (You may notice that there are other versions of `settings.yml` for staging and production. I don't think those need to be changed because `scripts/cipublish` should be handling images correctly in those cases).

And then later

3. Switch away from Mastarm so that we can use something with less idiosyncratic behavior and better documentation. Unfortunately, switching bundlers is unlikely to be straightforward, so I think we'll want to hold off until later in the project to make sure we've got the budget available for it.

I'll open an issue for switching bundlers--sorry this turned out to be a lot less straightforward than anticipated, but I hope this is helpful. Let me know if you have any questions!

ddohler (changes requested):
Thanks for calling out the strangeness with the `assets` folder! There does indeed seem to be something a bit unusual happening -- I've written up my findings in a comment. Luckily it seems like there's a relatively easy solution in the short term.

taui/src/components/custom-header-bar.js
ktohalloran: Thanks for looking into this! I will work on applying these changes you suggested. In answer to your question about whether the logo appeared on set up: no. I initially fully thought this was just a path issue until I looked at your PR for guidance on making my own and realized you could see the logo in your screenshots. This makes a lot of sense now as to why you could see it but both Scott and I could not.

taui/src/components/custom-header-bar.js
```diff
@@ -58,7 +58,7 @@ export default class CustomHeaderBar extends Greetings {
     return (
       <header className='app-header'>
         <div className='app-header__brand'>
-          <img className='app-header__logo' src='assets/BHAlogo.png' alt='' />
+          <img className='app-header__logo' src='assets/echo_combined_logo_fullcolor.svg' alt='' />
```
ddohler: While we're in here, this would be a good opportunity to fill out the `alt` property for these images, which will help with accessibility for screenreaders.

ddohler (approved):
Works great! I had one small comment relating to accessibility but this is good to go after that, no need for a re-review.

Issue #49, opened 2019-02-26
Docker development server maxes CPU on macOS


flibbertigibbet:
As a workaround, `yarn start` can be run directly on the host from the `taui` directory.

PR #371, opened 2021-11-04
Update README with account info and yarn command

ktohalloran:
## Overview

Originally, `README.md` had the wrong yarn command in its section on 'Running directly' and did not provide any information on how to log in once spun up. This PR changes `yarn add` to `yarn install` and adds where to find log in credentials, as well as how to create a Client ID following log in.

## Testing Instructions

 * Try to run `yarn add` on host; should respond with error `Missing list of packages to add to your project.` Try running `yarn install` instead.
 * Log in using the credentials referred to in the new version of the README. Follow instructions to create new client ID.


README.md
```diff
@@ -53,12 +53,18 @@ $ ./scripts/server
 ### Running directly
 
 * `cd taui`
-* Install packages: `yarn add`
+* Install packages: `yarn install`
```
ddohler, line 63: Interesting, I didn't remember that this was part of the README. I suspect this is here because of https://github.com/azavea/echo-locator/issues/49 , but that's an old issue, so I think it would be good to check and see whether it's still relevant -- does running using the Docker setup steps work for you on your Mac without causing your laptop to sound like a plane preparing for takeoff (you can also check your CPU usage with the Activity Monitor app)? If it does, then I think we should remove the `### Running directly` header and change this command to use Docker:
```
Install packages: docker-compose run --rm taui install
```

If Docker is still causing your CPU to hit 100% then we can leave these as-is, but I think it would be good to add clearer instructions that Docker is the preferred development environment and that these steps are for Mac users due to #49 . And then probably open a technical-debt issue to add Vagrant to work around the problem.

ddohler (commented):
Thanks for updating the README! I had forgotten that there are two different sets of setup instructions here, so I think there may be some additional clarity we can add to help people better choose which to use, depending on what the development situation is on Macs currently.

README.md
ktohalloran: So I tried both of these; Docker works but takes a _long_ time plus it does make my Mac sound like a jet engine (Activity Monitor is reporting Docker as using 300%+ CPU which I looked into because 😲 . Looks like that means I'm using 3 of my 4 cores). However, it's better than running on host because that didn't work at all--it failed while trying to build node-sass. While helping me spin up, Klaas mentioned that the lack of versioning on node and yarn could cause potential consistency problems with trying to run on host and that Docker was the lesser of the two evils. So I've been doing all of my work while on Docker and didn't try to pursue the build fail on host.

ktohalloran: Ok, I asked Rachele if she was able to run it on host and she is so something is happening with my computer. I'm going to try to look at it with Klaas during pairing tomorrow morning to figure out what it's getting caught up on and see if I can get running on host to work for me.

ktohalloran:
Ok, after meeting with Klaas, it turns out the issue was with my version of Node and possibly Yarn so he suggested adding `.yvmrc` and `.nvmrc` files and instructions into the README on using them when running on host. I also added more specific instructions on why there are two separate development environments available. So I believe this is ready for re-review!

taui/.nvmrc
```diff
@@ -0,0 +1 @@
+v12.22.5
```
ddohler, line 1: It doesn't look to me like these match the versions inside the Docker container; here's what I'm seeing in the container:
```
ddohler@lima:~/Development/projects/echo-locator$ docker-compose run --rm --entrypoint bash taui
root@65f554bdb4f6:/usr/local/src# node --version
v12.22.7
root@65f554bdb4f6:/usr/local/src# yarn --version
1.22.15
```

Does that match what you see if you follow those steps? If so, I'd update this and `.yvmrc` so that the Docker environment and on-host environment match as closely as possible. If this already matches the versions that you have inside of Docker locally then it's more likely a problem on my end because I may have older images lying around, so this is fine as-is in that case.

README.md
```diff
@@ -52,13 +55,21 @@ $ ./scripts/server
 
 ### Running directly
 
+* Make sure you have `nvm` and `yarn` installed (see links in Requirements)
```
ddohler:
```suggestion
* Make sure you have `nvm` and `yvm` installed (see links in Requirements)
```

README.md
```diff
@@ -52,13 +55,21 @@ $ ./scripts/server
...
 * `cd taui`
-* Install packages: `yarn add`
+* Specify Yarn and Node versions: Run `nvm use` and `yvm use`
```
ddohler:
```suggestion
* Specify Yarn and Node versions: Run `nvm use` and `yvm use` (you may need to run `nvm install` or `yvm install` first if you see error messages that the appropriate versions are not installed).
```
I ran into an issue the first time trying to use this because I didn't have the correct Node version already installed.

ddohler (approved):
This is working well, thanks for making these updates! I found a few minor areas for cleanup but this is good to merge once those changes are made.

taui/.nvmrc
ktohalloran: Yes, the values I added to the README match the ones used in my Docker container.

PR #374, opened 2021-11-08
Display BHA/Realtor.com listings

### LINKED ISSUE #356: Implement buttons to display BHA / Realtor.com rental units on map
[The demo site](https://demo.echosearch.org) has functionality which allows users to display markers for apartment listings for their chosen neighborhoods on the map. We should re-implement this functionality. This issue is just to show / hide the map markers when the appropriate button is clicked. The interactivity for the markers will be separate issues.

## Screenshots
Screen Shot 2021-10-13 at 9 26 54 AM

## Notes
- There are API Gateway endpoints that are in use on [the demo site](https://demo.echosearch.org) in order to pull down markers and display them. We should reuse these endpoints.
- Other than the endpoints, feel free to reuse as much or as little of the `demo` branch code as is helpful.
- There is no need for the functionality to match perfectly; if you see opportunities to improve either the code or the interface, feel free to take them.

rachelekm:
## Overview

This PR implements buttons to fetch and display BHA and Realtor.com listings as markers on the map. The buttons toggle between "show"/"hide" actions and the map markers should be appropriately visible/hidden when each button is clicked. Map marker interactivity will be handled separately, but button functionality should behave similar to [the demo site](https://demo.echosearch.org/map).

### Demo

Screen Shot 2021-11-08 at 2 59 58 PM
On Select Show BHA:
Screen Shot 2021-11-08 at 3 00 38 PM
On Select Show Realtor:
Screen Shot 2021-11-08 at 3 00 56 PM

### Notes

PR-specific notes: 
1. Loading spinner, button colors, map marker colors and top bar styling were set to match demo site
2. The Realtor.com API response includes 100+ listings for a neighborhood. While the demo site displays all listings returned from the request, we may eventually include a limit or pagination.
3. The BHA API response does not include `lat`, `lon` values and the address must be forward geocoded. Mapbox API does have an endpoint for batch geocoding, however it involves reaching out a sales rep and paying extra. Right now, there are only a handful of BHA listings per neighborhood so making forward geocode requests individually isn't burdensome, but might be necessary with an increase in BHA listings.

Larger project-specific notes:
1. This project implements `redux-actions` which is no longer actively maintained. This PR continues to use `redux-actions` to handle state but it would be worth eventually looking into more stable redux utility packages.
2. This project handles async logic in action dispatches through custom middleware: `fetch`. By default, if no `error` parameter is present then, on `error`, `fetchError` is dispatched. We currently don't have any reducer to handle a `fetchError` dispatch and we have multiple action creators that depend on this action for error handling. This branch includes custom error handling, however we should create a separate issue that handles `fetchError` more broadly.

## Testing Instructions

 * Run `./scripts/server`
 * Log in and select a neighborhood (select 'Boston - Allston – 02134' neighborhood for BHA listings)
 * Confirm top bar and buttons appear above map
 * Select 'Show BHA' and confirm listings appear, select 'Hide BHA' and confirm listings disappear
 * Select 'Show Realtor' and confirm listings appear, select 'Hide Realtor' and confirm listings disappear
 * Select a new neighborhood and confirm listings clear
 * Simulate an API error (either by dropping a letter in the url or blocking request) and confirm that an error message appears next to buttons

Resolves #356


taui/src/actions/listings.js
```diff
@@ -0,0 +1,91 @@
+// @flow
+import fetch from '@conveyal/woonerf/fetch'
+
+import {fwdGeocodeBatch} from '../utils/fwd-geocode'
+
+import {addActionLogItem} from './log'
+
+const REALTOR_BASE_URL = 'https://akk8p5k8o0.execute-api.us-east-1.amazonaws.com/staging/get-realtor-listings'
+const BHA_BASE_URL = 'https://akk8p5k8o0.execute-api.us-east-1.amazonaws.com/staging/-get-bha-listings?'
+const REALTOR_TYPE = 'set Realtor listings'
```
ddohler:
```suggestion
const REALTOR_ACTION_TYPE = 'set Realtor listings'
```
I think this (and `BHA_TYPE`) would be more self-explanatory if they call out that they're an action type; when I first read this I wasn't 100% sure what type of type these were.

taui/src/actions/listings.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }))
+}
+
+export const setBHAListings = () => (dispatch: Dispatch, getState: any) => {
+  const current = getState()
+  const zipcode = current.data.activeNeighborhood
+  const budget = current.data.userProfile.budget
+  const rooms = current.data.userProfile.rooms
+  const urlWithQuery = BHA_BASE_URL + 'zipcode=' + zipcode + '&rooms=' + rooms + '&budget=' + budget
```
ddohler: I usually prefer to implement this type of pattern (constructing a string from a series of components) using template literals by using backticks, something like:
``const urlWithQuery = `${BHA_BASE_URL}?zipcode=${zipcode}...` ``

I find doing it with backticks makes the string easier to understand because the template literal matches the intended string format pretty closely and keeps all the components in one place.

taui/src/actions/listings.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+    next: (error, response) => {
+      // handle error here instead of automatically dispatching fetch's fetchError
+      if (error) {
+        console.error('Error fetching Realtor listings.')
+        console.error(error)
+        return {type: REALTOR_TYPE, payload: {error: error}}
+      }
+      try {
+        if (typeof response.value.body === 'string') {
```
ddohler: Do you know whether `response.value.body` is ever not a string (or conversely, ever is)? It would be surprising to me if it changed frequently, so unless you know that both of these branches are necessary, it would simplify the logic to assume that `body` has a certain format and handle it appropriately. If you want to handle the case where the format isn't what we expect, I think that raising a helpful error message would be sufficient (and potentially more robust because it'll let us know that things have changed rather than silently doing something different that might not be correct under changed conditions).

taui/src/actions/listings.js
```diff
@@ -0,0 +1,91 @@
...
+    next: async (error, response) => {
+      // handle error here instead of automatically dispatching fetch's fetchError
+      if (error) {
+        console.error('Error fetching BHA listings.')
+        console.error(error)
+        return {type: BHA_TYPE, payload: {error: error}}
+      }
+      let listings = []
+      try {
+        if (typeof response.value.body === 'string') {
```
ddohler: Same comment here about potentially removing this typecheck or converting it to a "happy path + notification if we detect deviation from the happy path".

taui/src/actions/listings.js
```diff
@@ -0,0 +1,91 @@
...
+          listings = JSON.parse(response.value.body)
+        } else {
+          listings = response.value.body
+        }
+      } catch (error) {
+        console.error('Error parsing BHA response.')
```
ddohler: In general it's good to have fine-grained errors to help future debuggers, but that's often in tension with keeping the code readable, which can assist future travelers as well. In this case I think it would be fine to roll both of these try / catch statements into one bigger one -- there aren't that many places where something would be expected to go wrong in this process so localizing the error to the `next` function as a whole is good enough and I think it would simplify the logic a good deal, especially in combination with simplification to the type-checking on `response.value.body`.

taui/src/actions/neighborhood.js
```diff
@@ -16,6 +16,11 @@ export const setPage = (page) => (dispatch, getState) => {
   dispatch({type: 'set page', payload: page})
 }
 
+export const setShowBHAListings = (show) => (dispatch, getState) => {
+  addActionLogItem(`Set show BHA listing to ${show}`)
+  dispatch({type: 'set show BHA listing', payload: !!show})
```
ddohler: As I'm looking at this, I don't think that `!!` is necessary here (it's usually used defensively to ensure that a function requiring a boolean actually gets a boolean, and we have type-checking so that should never happen). But I see that it's following the pattern from the rest of the file and it doesn't hurt anything, so it's fine to leave it as-is.

taui/src/components/map.js
```diff
@@ -282,6 +296,26 @@ export default class Map extends PureComponent<Props, State> {
             </Popup>
           </Marker>}
 
+        {p.showRealtorListings && p.realtorListings.data && p.realtorListings.data.map((item, key) => {
+          return <Marker
```
ddohler: Since you're just returning this `<Marker` without doing anything else you can omit the curly braces and explicit `return` statement here (though it does no harm to keep them).

taui/src/components/top-bar.js
```diff
@@ -0,0 +1,174 @@
...
+
+  render () {
+    const p = this.props
+    const {
+      showBHAListings,
+      showRealtorListings,
+      bhaListings,
+      realtorListings
+    } = p
+    const ShowListingsButton = this.showListingsButton
```
ddohler: This multi-component framework is functional, so I don't think any changes are necessary, but here's an alternative way to structure it for your consideration: Rather than having three components for each of the three button states (off, pending, on) it might make things easier to manage if you moved all three into a single `ToggleListingsButton` component. The interface to that might look something like:

```
<ToggleListingsButton
  onShow={() => { this.props.setBHAListings(); this.props.setShowBHAListings(true) }
  onHide={() => { this.props.setShowBHAListings(false); }
  pendingWhen={bhaListings.pending}
  showWhen={showBHAListings}
  showMessage={...}
  hideMessage={...}
/>
```
(I haven't thought through the interface thoroughly so there might be some wrinkles that I'm missing).

I think the benefit of doing that would be that you'd be able to reuse not only the functions for showing / hiding, but also the logic for how the button changes state depending on its inputs. As things stand, it's possible for the two listing buttons to behave differently if someone changes the logic for one but not the other, and it's a little complex to figure out the effect that each state variable has. By combining into a single component you'd be guaranteed that both would behave identically, and you'd also set up a clear expectation for the effect of each state variable by creating an interface for them.

That said, this would be a larger refactor, and this is functional as-is, so I don't think it's crucial to take this on unless you prefer this version!

taui/src/types.js
```diff
@@ -289,3 +289,5 @@ export type MapEvent = {
     _zoom: number
   }
 }
+
+export type Listing = { pending: true } | { data: Array } | { error: String }
```
ddohler, line 293: Nice, this is great.

taui/src/utils/fwd-geocode.js
```diff
@@ -0,0 +1,19 @@
...
+      return {lat: response.data.features[0].center[1], lon: response.data.features[0].center[0]}
+    })
+    .catch(err => {
+      return (err)
+    })
+}
+
+export async function fwdGeocodeBatch (listings: Array): Promise<void> {
+  const promises = listings.map((item, key) => {
+    return fwdGeocode(item.geoAddress)
```
ddohler: Same comment here as earlier, this can just be `listings.map((item) => fwdGeocode(item.geoAddress))`.

In fact, you could potentially collapse this whole function down to 
```
  return Promise.all(listings.map((item) => fwdGeocode(item.geoAddress)))
```

taui/src/selectors/bha-listings.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import get from 'lodash/get'
+import {createSelector} from 'reselect'
+
+export default createSelector(
```
ddohler, line 5: Is this actually being used? I'm having trouble spotting it but I remember from the past that it hasn't always been obvious to me how selectors get used.

ddohler (changes requested):
Nice work putting this all together!

I made some comments for potential cleanups, but most of them are not urgent. However, I'm spotting two things that almost certainly need to be addressed:
- All of the map layers disappear while a request for markers is pending. I'm not exactly sure what's causing this, but it's likely that the other map layers are being impacted by the pending status somehow, probably via a state variable. I suspect the fix will require either reworking some of the conditionals by which the markers are displayed, or adding a new state variable for tracking the marker layer pending state separately from the other layers.
- The Realtor.com markers don't seem to be showing up for me. The request/response seems to work, but the response is always empty. I'm hoping this is because a parameter got forgotten in the process of putting together the PR, but if the realtors API has suddenly started failing at the least opportune moment then we may need to look into this more deeply.
Screen Shot 2021-11-09 at 2 23 27 PM

rachelekm:
@ddohler thank you for these edits, will go ahead and implement with special attention to the first two issues (map layers and realtor listings). However, I'm unable to replicate the empty Realtor response error. I ran git pull and then cleared my browser cache to be sure, but I'm able to populate the map with listings. Can you share the options body sent with post request?

taui/src/selectors/bha-listings.js
rachelekm: This selector returns a default value on app start before the `'set BHA listings'` action gets dispatched and then returns data based on state once updated by `'set BHA listings'`. Aside from returning a default value it's just a direct lookup of state, but still seems to be recommended in the redux docs and keeps in line with how the rest of the project derives data from state using mapStateToProps.

taui/src/components/top-bar.js
```diff
@@ -71,14 +75,35 @@ export default class TopBar extends PureComponent<Props, State> {
     )
   }
 
+  loadingListingsButton (props) {
+    const { message } = props
+    return (
+      <div style={{ display: 'inline-block', position: 'relative' }}>
```
lederer, line 120: Haven't tried it locally but I'm pretty sure you can drop the wrapping `div` and just add `position: relative` to `&__button-loading` in `_top-bar.scss`

ddohler:
@rachelekm I just tried viewing Realtor.com listings again without changing anything (well, other than accessing from the office rather than from home) and now it works fine. :shrug: If it happens again we can make an issue to look into it, but it doesn't seem likely to be a problem with these changes!

taui/src/selectors/bha-listings.js
ddohler: Got it! That seems fine then. :+1:

taui/src/components/main-page.js
```diff
@@ -165,14 +180,21 @@ export default class MainPage extends React.PureComponent<Props> {
...
+            showRealtorListings={p.data.showRealtorListings}
             showRoutes={this._showNeighborhoodRoutes()}
             setActiveNeighborhood={p.setActiveNeighborhood}
             setPage={p.setPage}
             setEndPosition={p.updateEndPosition}
+            setShowBHAListings={p.setShowBHAListings}
             setShowDetails={p.setShowDetails}
+            setShowRealtorListings={p.setShowRealtorListings}
             setStartPosition={p.updateStartPosition}
+            showBHAListings={p.data.showBHAListings}
```
ddohler: I missed this on my pass, but running `docker-compose run --rm taui lint-js` informs me that this prop is duplicated here and on line 185, and `showRealtorListings` is also duplicated.

rachelekm:
@ddohler , I implemented these changes with the addition of a refined error message in the TopBar component. The missing layers/components during listings pending was due to `isLoading` state being updated on `INCREMENT_FETCH`. From what I can tell, `INCREMENT_FETCH` updates the `fetch` state which then updates the `isLoading` state, derived through the `loading.js ` selector. The `fetch` state is used to load all networks when the `loadDataset` action creator is called, and `isLoading` is used as a conditional for a variety of components until that data has completed loading. For now, I added a conditional in the `INCREMENT_FETCH` reducer to ignore fetches containing our listings' urls since it’s not relevant to the intended function of `fetch` state and allows us to bypass the update to `isLoading`. However, I'm sure there's a better way to fix this that creates a better foundation for future fetches—maybe by creating separate `fetch` and `networkFetch` states, with `isLoading` state derived from the latter? Let me know what you think!

lederer (approved):
Thanks for doing this. I don't think this one needs any urgent UI polish. I will include it in a general polish pass later.

taui/src/components/top-bar.js
```diff
@@ -0,0 +1,165 @@
...
+  props: Props;
+
+  constructor (props) {
+    super(props)
+    this.error = this.error.bind(this)
+  }
+
+  componentDidUpdate (prevProps) {
+    // Checking against previous props prevents looping error handling
+    if (((this.props.bhaListings.error && prevProps.bhaListings !== this.props.bhaListings) || (this.props.realtorListings.error && prevProps.realtorListings !== this.props.realtorListings))) {
```
ddohler: I'm surprised that the default formatting on this project doesn't attempt to keep the line length shorter; hopefully we'll have time to switch to a new bundler with a more mainstream formatting solution, but for the time being, this line would benefit from some linebreaks so that it's not quite so long.

taui/src/components/top-bar.js
```diff
@@ -0,0 +1,165 @@
...
+        onClick={onShow}>{ show }
+      </button>
+    ))
+  }
+
+  error () {
+    return (
+      <div style={{ display: 'inline-block', position: 'relative' }}>
+        <Icon type='exclamation-triangle' />
+        There was an error fetching your listings
```
ddohler: This should be moved into the messages file and wrapped in a `t()` call here.

ddohler (approved):
Works great, nice job! 👏  I made two minor cleanup comments but this is good to go once those are addressed, no need for re-review.

PR #377, opened 2021-11-11
Add translation buttons

### LINKED ISSUE #355: Add Spanish and Chinese translations and language switcher
This exists on [the demo site](https://demo.echosearch.org) and our implementation should match the functionality there.

## Notes
- The functionality match doesn't need to be perfect. If you see easy opportunities for improvement, feel free to take them.
- Feel free to reuse as much, or as little, of the code on the `demo` branch as is helpful.

ddohler:
## Overview

Adds a language switcher and translation machinery to enable the app to appear in multiple languages.

### Demo

ECHOSpanish

### Notes

- The application used to use a custom-built library called [`woonerf`](https://github.com/conveyal/woonerf/) to handle message strings, but it was only a partial solution for multiple languages, and it is no longer maintained. After some discussion with others, I decided to attempt to switch the app to use `react-i18next`, which is a full-featured, well-supported React translation library. This turned out to be a great choice -- `woonerf` and `react-i18next` use very similar patterns for looking up message strings and handling string interpolation, so `react-i18next` was largely a drop-in replacement.
- Most of the changes in this PR are related to converting components from using `woonerf` to use `react-i18next`. This generally followed a pattern something like this:
  - Wrap component in `withTranslation()` higher-ordered-component. This makes `t` and `i18n` objects available in the component's props.
  - Destructure props to pull out `t`
  - Replace calls to `message()` with calls to `t()`.
- However, there are a lot of places where we've added what essentially amount to function components, but inside of existing class-based components as class methods. I'm not sure what the motivation was for this pattern, but it means that the most straightforward path to injecting the `t` parameter into those places varied depending on the details of the implementation of the sub-component. In general, I tried to prefer using the `useTranslation()` hook in these cases because much of the time they didn't access `this` and thus were essentially just function components anyway, but in some cases they were more like class methods because they were destructuring `this.props` from the parent, so in those cases I just pulled out the `t` that was already there.
- The Amplify code isn't translated because it apparently uses its own, separate translation machinery. This primarily affects the login flow, and the translation switcher isn't available on the login page anyway, so I didn't try to find a solution. In any event, we won't be able to use Amplify's login components once we switch to Django authentication, so eventually the login components will be migrated to `react-i18next` too.
- There is one spot with numbers that I didn't spend too much time on -- plurals can be tricky to translate, but to the best of my knowledge, all of our target languages have relatively simple number handling, so I left it alone. If it turns out to be necessary, `react-i18next` has support for more complex numbering schemes.
- @lederer the current language picker pretty closely matches the demo but it definitely needs some improvement -- let me know what changes I should make.
- The `messages-fr.yml` file is a holdover from when this project was forked; it could have been deleted at any time, but now seemed like a good time.

## Testing Instructions

 * Click around on all views (profile, map, neighborhood detail) and make sure that you see no console errors or blank pages.
 * In general not much is translated yet, but the Profile page has pretty good coverage in Spanish, so head there and confirm that the translation switcher works. You could also add a translation for some other part of the app and confirm that it shows up when you switch languages elsewhere.

Resolves #355


lederer:
I added a couple small commits to space out the picker buttons, and turn the enabled one black, but it needs some of your help… 

Can you add some logic that toggles an `app-header__button--on` class on the 3 language buttons? So the one that's active should have that class and the others should not.

lederer:
Eek. My editor broke the file formatting. I will fix momentarily.

taui/src/index.js
```diff
@@ -43,6 +48,25 @@ API.configure({endpoints: awsmobile['aws_cloud_logic_custom']})
 Auth.configure(awsmobile)
 Storage.configure(awsmobile)
 
+console.log('process.env.environment', process.env.LEAFLET_TILE_URL)
```
KlaasH: This looks unrelated. Did it sneak in from something else?

KlaasH (commented):
This looks good.  Big yay for switching to a mature and actively maintained package.

I have one question, which I started to write as a line note than realized it's consistent throughout:
There are a bunch of places where `export default class ClassName` was changed to `export class ClassName` with `export default withTranslation()(ClassName)` added at the end of the file.  Is the `export` in the new version of the class declaration necessary?  I would have thought it should go away along with the `default`.  It seems like exporting it wouldn't provide anything useful, and would make the bug that would arise from accidentally importing with `import { ClassName } from './components/class-name'` (rather than `import ClassName from './components/class-name'`) more difficult to diagnose.

taui/src/index.js
ddohler: Oops, good catch!

ddohler:
Hmm, that's a good point about the exports vs. default exports; we often export everything in a file and I fell into following that pattern, but you're right that in this case, there's no use for the non-default thing, and in fact it would most likely cause problems if you tried importing it by accident. I'll go ahead and remove those exports entirely.

ddohler:
Okay, updates pushed!
@KlaasH I cleaned up the exports and removed that stray `console.log`.
@lederer I added the styling for the active language:

https://user-images.githubusercontent.com/447977/141852397-4fba0f62-62e8-4084-89a4-a7f58f0ece3d.mp4

lederer (approved):
Thanks for the follow up!

KlaasH (approved):
:+1: Looks good!

PR #385, opened 2021-11-24
Display listing details on Marker

### LINKED ISSUE #358: Display apartment detail on markers
The [demo site](https://echosearch.org) allows users to hover over a map marker in order to see detailed information about an apartment, including a listing photo, rental cost, and a link to further details. We should re-implement this functionality.

## Screenshots
Screen Shot 2021-10-13 at 10 12 03 AM

## Notes
- @lederer The current popup-on-hover functionality is kind of frustrating to use because it recenters the map, so if you hover over a marker near the edge accidentally the map will move suddenly and unexpectedly. I think we could stop it from recentering while still displaying the popups on hover, or we could switch the popups to display only on a click but maintain the recentering behavior. Do you have any thoughts?
- We can reuse as much or as little of the code on the `demo` branch as is helpful when re-implementing this, and the popups do not need to match the demo site exactly.

rachelekm:
## Overview

Allows users to click on or hover over a map marker to see detailed information about an apartment, including listing photos, rental cost, number of bedrooms, address, and a link to further details. Markers and their corresponding pop-ups do not recenter the map in this PR and pop-ups display on both hover and click functionality for now, as discussed in the #358 conversation. Other than those changes, the content and styling of each popup currently matches the `demo` branch as best as possible while meeting the `develop` branch’s current styling.

### Demo
Screen Shot 2021-11-24 at 4 42 47 PM

### Notes

Pop-up’s thumbnail carousel implements [Nuka](https://github.com/FormidableLabs/nuka-carousel). Library was chosen as it fits the basic carousel needs, allows for dynamic image height, and includes built-in type definitions. 

## Testing Instructions

 * Run `./scripts/server`
 * Log in and select a neighborhood (select 'Boston - Allston – 02134' neighborhood to test BHA listings)
 * Select 'Show BHA' or 'Show Realtor' to display markers
 * Confirm marker pop-up opens on mouse hover and closes on mouse exit
 * On marker click, confirm pop-up continues to display on mouse exit and closes when close button is selected
 * Confirm map does not recenter when markers/pop-ups are near edge of map bounds
 * Confirm all pop-up content is present and link and image carousel work as expected

Resolves #358

### DIFF
--- a/taui/src/utils/standardize-listings-data.js
+++ b/taui/src/utils/standardize-listings-data.js
@@ -1,0 +1,29 @@
+// @flow
+
+export default function standardizeListingsData (fn: Function): Function {
+  return function (...args: [any, 'BHA' | 'Realtor']) {
+    const [data, type] = args
+    switch (type) {
+      case 'BHA':
+        return fn({
+          photos: data.photos,
+          rent: data.Rent,
+          beds: data['Bedroom Type'] === 'Studio' ? data['Bedroom Type'] : `${data['Bedroom Type']} Bed`,
+          address: data['Apartment Number'] ? `${data.address.line} #${data['Apartment Number']}` : data.address,
+          url: data.rdc_web_url,
+          lat: data.lat,
+          lon: data.lon
+        })
+      case 'Realtor':
+        return fn({
+          photos: data.photos,
+          rent: data.price,
+          beds: `${data.beds} Bed`,
+          address: data.address.line,
+          url: data.rdc_web_url,
+          lat: data.address.lat,
+          lon: data.address.lon
+        })
+    }
+  }
+}
--- a/taui/package.json
+++ b/taui/package.json
@@ -55,6 +55,7 @@
     "lodash": "~4.17.15",
     "logrocket": "^1.0.6",
     "logrocket-react": "~4.0.1",
+    "nuka-carousel": "^4.8.3",
     "react": "~16.13.0",
     "react-dom": "~16.13.0",
     "react-ga": "~2.7.0",
--- a/taui/package.json
+++ b/taui/package.json
@@ -55,7 +55,6 @@
     "lodash": "~4.17.15",
     "logrocket": "^1.0.6",
     "logrocket-react": "~4.0.1",
-    "nuka-carousel": "^4.8.3",
     "react": "~16.13.0",
     "react-dom": "~16.13.0",
     "react-ga": "~2.7.0",
--- a/taui/src/components/map.js
+++ b/taui/src/components/map.js
@@ -21,6 +21,7 @@ import type {
   MapEvent,
   Listing
 } from '../types'
+import standardizeData from '../utils/standardize-listings-data'
 
 import DrawNeighborhoodBounds from './draw-neighborhood-bounds'
 import DrawRoute from './draw-route'
@@ -159,29 +160,6 @@ export default class Map extends PureComponent<Props, State> {
     }
   }
 
-  listingsDetailPopup = ({photos, rent, beds, address, url}) => {
-    return <div className='map__popup'>
-      <Carousel heightMode='current' defaultControlsConfig={{
-        nextButtonText: '>',
-        prevButtonText: '<'
-      }}
-      >
-        {photos.map((item, key) =>
-          <img className='map__popup__image' src={item.href} key={`listings-image-${this._getKey()}`} />
-        )}
-      </Carousel>
-      <div className='map__popup-contents'>
-        <h1>{`Price: $${rent}/month`}</h1>
-        <h2>{beds}</h2>
-        <div className='map__popup__line' />
-        <p>{address}</p>
-        <div className='map__popup__url-wrapper'>
-          <a className='map__popup__url' href={url} target='_blank'>Go to Apartment</a>
-        </div>
-      </div>
-    </div>
-  }
-
   /**
    * Reset state
    */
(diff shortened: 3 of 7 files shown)


taui/src/utils/standardize-listings-data.js
```diff
@@ -0,0 +1,33 @@
...
+          rent: data.price,
+          beds: `${data.beds} Bed`,
+          address: data.address.line,
+          url: data.rdc_web_url,
+          lat: data.address.lat,
+          lon: data.address.lon
+        }
+        break
+    }
+    return fn(newData)
```
KlaasH: This structure of declaring an empty `newData` up front, setting it in the `switch`, then returning it at the end took me a moment to parse out, and I think it could be a little more self-contained.  Here's one possibility--keeping the switch returning from inside each case.  Or it could be done by assigning `newData` with a ternary chain (then still calling `fn()` on it at the end).

```suggestion
    switch (type) {
      case 'BHA':
        return fn({
          photos: data.photos,
          rent: data.Rent,
          beds: data['Bedroom Type'] === 'Studio' ? data['Bedroom Type'] : `${data['Bedroom Type']} Bed`,
          address: data['Apartment Number'] ? `${data.address.line} #${data['Apartment Number']}` : data.address,
          url: data.rdc_web_url,
          lat: data.lat,
          lon: data.lon
        })
      case 'Realtor':
        return fn({
          photos: data.photos,
          rent: data.price,
          beds: `${data.beds} Bed`,
          address: data.address.line,
          url: data.rdc_web_url,
          lat: data.address.lat,
          lon: data.address.lon
        })
    }
```
Actually this isn't quite the same, since if `type` doesn't match either case, this won't return anything, whereas the original will return `fn(undefined)`.  Not sure if that's an important distinction, or what the failure case of this function would ideally be.

KlaasH (changes requested):
This looks good and works well over all.  I had one code style thought below, and a few notes on how it's working in the front-end:
1. The way the pop-ups always grow isn't ideal. Obviously we could just pick a height and not let anything be taller than that. Or if we want to keep showing the whole image, we could set a default height that's somewhere in the middle of the typical range, so the resizing wouldn't be so dramatic when it happens.

2. The hover functionality is kind of nice for being able to quickly scan through the different markers, but it breaks down if you find one you like and want to click the "Go to apartment" button. As soon as the mouse leaves the marker to move to the link, the pop-up with the link disappears.  Same with the image carousel--there's no way to interact with it unless you click the marker to make the pop-up stick around.

   I could see doing slightly different variations on the pop-up depending on how it's opened. I.e. hover gets you the picture and info but no link or carousel, clicking gets you the full thing.  That would preserve the advantage of being able to use hover to quickly scan through the listings, but of course it would add complexity to the implementation.

3. The carousel arrows are working for me, but the dots aren't.  That is, there's a black dot sitting in the middle of the image that looks like it's probably supposed to be carousel dots, but there's only ever one of them and clicking it doesn't do anything.

lederer:
### Hover/Click

Thanks for trying out the current hover/click behavior. This really helps clarify our solution options. As Klaas mentioned, the current hover behavior effectively invites mousing into the popup to click its contents, yet also prevents doing so. From here I think there are three options:

1. **Allow mousing into the hover-triggered popup.** This requires keeping the popup visible for a short period of time after mousing out of the marker (typically `~250–400ms`), so there's time to then mouse into the popup. If the map library offers this as a built-in option, let's try it. If not, I recommend against it. While it's easy enough to implement the basic logic of it, there are a number of edge cases and nuances that make it feel weird unless there's a very sophisticated (read: complex implementation) logic operating behind the scenes (including detecting mouse trajectories, etc).

2. **Hover for a simple popup, click for a more complex one.** The hover popup could be smaller (eg, just address, beds, price; no photo), or it could be full size (with photo). In either case, it shouldn't contain any interactive elements (no close button, no carousel, no "Go to…"). Then on marker click, the popup persists and becomes interactive (close button, carousel, "Go to…").

3. **Hover for a simple popup, click to visit the detail page.** The popup would never have any interactive elements. This means we could only show one photo; no carousel. But on the plus side, it would be easier to reach the detail pages (a single click instead of click -> move to button -> click).

I think the choice depends in part on implementation complexity. Curious for your thoughts about which might be easier to implement and which feels more right to you.


### Popup size

Agree with Klaas about popup height. I recommend two static heights (with vs without photo). Let's try `160` without photo and `320` with photo. And a static width of, say, `240` in all cases. Hopefully we can easily change those numbers if needed.

That will require a static size for the carousel as well. If the carousel library doesn't naturally fit the photos to the static carousel size, we may be able to finesse it with CSS (eg, `object-fit: cover`).


### Empty carousel

Some listings show the carousel, but with no photos. Not sure what's happening there. (I'm not convinced they have no photos, because there are other listing with no photos AND no carousel.)

For example this listing in `Boston - Dorchester – 02121`…
image

I think the carousel logic should be:
- If no photos, don't show carousel.
- If one photo, show photo but without carousel.
- If > 1 photo, show photos in carousel.


### Tooltip

The neighborhood name tooltip that sticks around as you hover the map is kinda pesky when hovering a map marker. Can that be hidden while hovering any marker or popup?


### Polish

JFYI I will take care of popup content and styling in a polish pass coming up soon.

rachelekm:
Hey @KlaasH  and @lederer thank you so much for these reviews and suggestions, I went ahead and implemented these edits to the best of my ability and here are my notes on the changes:

- **Hover/Click Behavior**: As far as I can tell, neither react-leaflet or the Leaflet library have a built-in ability to set a timeout on popup open. I decided to move forward with trying to set up a simple popup during hover and a more complex one on click since I already had a state variable `clickCheck` to differentiate between click and hover actions that I could use as a conditional. However, I ran into lots of funny side effects with this strategy. Namely, when including photos in the click popup jsx content, whether carousel or just a single image, it broke down how/which popups open, if they opened at all. At first I thought it could be a rendering issue (since the popup height changes on click to accommodate the carousel), so I added a photos `div` to the hover popup as a placeholder. However, even with this change, when the popup did open it was for a different marker. Somehow, when the popup content updates on click the marker event target is lost and there doesn’t seem to be an order to which popup opens in its place. As I am able to create a popup with dynamic content (including a `div` for the url doesn’t seem to be an issue), it stands that adding in photos/carousel on click is _possible_ but past my ability to implement*. If anyone has any ideas for why this might be happening/how to implement a fix I think this is still a plausible option, but for now I’ve moved forward with the third option: hover for a simple popup, click to visit the detail page. Since the only interactive element present on the click popup was the url button, it made sense to drop the popup altogether and make click the redirect, as suggested. Let me know what you think. It might make sense to include some type of “Click marker to see more…” text within the hover popup or in the bar above so that action is more intuitive? 

- I also can’t get rid of the neighborhood tooltip when hovering over a marker – I can’t seem to get rid of it without deleting the tooltip present when generally hovering the whole neighborhood.

- @KlaasH re: the `standardize-listings-data` fn, I’ve edited it to the structure you’ve suggested and then defined specific `‘BHA’ | ‘Realtor’` type to the `type` variable so that it will throw a type error if it doesn’t match either case.

*There is another approach which is to use Leaflet Popup’s `.setContent()` method to reset the content html on the appropriate action. This works, but requires a String or HTMLElement to be passed in and using template literals fell short/were going to require a lot of adjustments to accommodate dynamic content and CSS styling previously set by class names. Some clever solutions I found were to use React Portals or `ReactDOM.renderToString`, but this quickly got complex so I opted for a simpler option. However, if we feel strongly about keeping the url button and carousel on click, this could be a potential route.

lederer:
Thanks for making these changes. I agree the solution you've implemented is the right direction given the constraints.

A few notes:

- Please do add some "Click for details" text to the bottom of the popup. I'll style it in the forthcoming polish pass.
- Is it possible to hide the close `x` button in the popup corner?
- Try the following CSS on the image. It should ensure that the image takes up all available width, while cropping it vertically to fit within the height.

```css
width: 100%;
height: 160px;
object-fit: cover;
```

KlaasH (approved):
:+1: Looks good!
The hover interaction and click-for-details seems just fine to me--once I know the location and price, I either want to see the details page or I don't. Having a carousel in the little pop-up doesn't add much.  Though speaking of that, this branch now adds the `nuka-carousel` package but doesn't use it, right?  We should remove the import and update `yarn.lock` accordingly.

rachelekm:
Thanks all! I removed the last remnants of `nuka-carousel` and made those styling changes, @lederer let me know if you see anything else I can edit.

lederer (approved):
👏 Thanks for incorporating the feedback. Great job iterating!

Issue #389, opened 2021-12-13
Increase flexibility of grammar in translations


ktohalloran:
Both the Spanish and Chinese translations pose unique challenges in terms of grammar and the way we have structured our translations within components. This issue is about restructuring the places where this grammar is currently incorrect or where our method does not work with the translations we need to add. This issue does not include all dynamic text; a list of affected phrases is available below. One way we got around this in other parts of the app for phrases that are not in this list was to use interpolation, replacing the entire phrase at once and simply handing the translation the dynamic values, such as was used to populate the tooltips. Another possibility is to introduce some form of logic to check for language and then piece together the sentence from there.

**Chinese**
The way some of the phrases are currently built and translated works for both English and Spanish but needs to be restructured to allow for reordering as required by Chinese grammar. For example, the phrase "Take [bus number] in [number] minutes" in the dock is currently structured with "Take", "in", and "minutes" as separate phrases that get translated separately and pieced together into a sentence along with the dynamic values within the component. This order of words does not work in that order in Chinese.

This is the list of impacted phrases; all are from the dock in the neighborhood detail view. Note that for the last two, clarification on the translation may be necessary.
* About [number of min] min via subway/bus from [destination] | 從[destination]方向乘地鐵/巴士約[number of min]分鐘
* Take [route number] in [number of min] minutes | 在[number of min]分鐘內乘坐 (it's unclear where in this translation the [route number] should go)
* Search for [number of br]br with max rent $[dollar amount] | 以最高租金搜索[number of br]臥房單位 (it's unclear where in this translation the [dollar amount] should go)

**Spanish**
The existing Spanish phrases aren't always grammatically correct and require special rules when reusing the same word across multiple parts of the app. For example, the app references the Spanish word "trabajo" in the Profile page destination dropdown menu and in the phrase "[mode of transportation] de trabajo" in the dock. While the word "trabajo" without an article makes sense for the dropdown menu, that word should become "el trabajo" when used in a sentence and, if the word before it is "de", should become "del trabajo."

This is the list of impacted phrases:
* [mode of transportation] from [destination] (dock, viewing recommendations)
* About [number of min] min via subway/bus from [destination] (dock, neighborhood detail view)

You should branch off of `release/2022-03-02` for this work and make the PR back to that same branch.

ddohler:
[...]

PR #390, opened 2021-12-13
Add translation strings

### LINKED ISSUE #384: Add translation strings
The application has translation machinery in place, but most strings are not yet translated. We have more complete translations available; this task is to put those translations into the application. The translations are attached.

This task will largely involve copy-pasting from the translation files into the appropriate [translation file](https://github.com/azavea/echo-locator/tree/develop/taui/src/locales) file in the project source. However, there are some places where a bit more care and/or effort will be required. See the notes, below.

## Notes
- Some strings require inserting other values, for example "This neighborhood has a **low** rate of crime compared to others". The translators provided guidance on where and how to do this, so it will be necessary to consult the [react-i18next documentation](https://react.i18next.com/) (or maybe simply the regular [i18next documentation](https://www.i18next.com/) in order to figure out how to accomplish what is needed.
- The translation documents say that we should include both English AND another language on the page when the application is translated: "Leave all English text on the page. Only add translation text. Specifically, add the translated to the right and under its English version". However, we double checked with the client and this is NOT necessary. The way the app is currently translated is sufficient, only the strings need to be updated.

## Attachments
[ECHO translation files.zip](https://github.com/azavea/echo-locator/files/7563507/ECHO.translation.files.zip)

ktohalloran:
## Overview

This PR adds translation strings that switch when the translation buttons are clicked. 

### Demo

Screen Shot 2021-12-13 at 11 16 46 AM

Translated into Spanish

Screen Shot 2021-12-13 at 11 16 55 AM

Translated in Chinese

### Notes

While some of the strings were very straight forward, a lot of them were not. Here is an overview of the problems faced and the decisions made. 

**Missing Translations**
* Missing translations in **Spanish**: "Peak", "Off Peak" (in dock when commuter trains are selected in profile), "Site by" (dock)
* Missing translations in **Chinese**: "New Search" (nav bar), "when" (dock), "Apartments" (nav bar), "Resend invitation email" (edit profile), "Primary" (edit profile)
* Missing translations in **both languages**: "Peak No Express", "Off Peak No Express" (in dock when commuter trains are _not_ selected in profile), "of" (in the dock, e.g. 1-3 of 142), all neighborhood descriptions

**Amplify Code**

There were also elements that had translations but were not incorporated for similar reasons as laid out in [PR 377](https://github.com/azavea/echo-locator/pull/377). As quoted from that PR, the log in components, especially "Sign Out", "Username", "Password", "Forgot your password?", and "Reset password" are not translated "because [Amplify code] apparently uses its own, separate translation machinery. This primarily affects the login flow, and the translation switcher isn't available on the login page anyway, so I didn't try to find a solution. In any event, we won't be able to use Amplify's login components once we switch to Django authentication, so eventually the login components will be migrated to `react-i18next` too."

**Grammar in Dynamic Phrases**

One of the issues I ran into with translating dynamic phrases into Chinese was how we structured sentences. For example, the phrase "Take [bus number] in [number] minutes" in the dock is currently structured with "Take", "in", and "minutes" as separate phrases that get translated separately and pieced together into a sentence along with the dynamic values within the component. There are grammatical issues with some of the existing translations, as well, especially in phrases that have been structured piece-meal. One example is in the "[mode of transportation] from [destination]" phrase in the dock. This particular sentence has grammatical issues at least in Spanish and possibly also in Chinese, again due to how we have chosen to structure that component. These sentences also have an added complexity in that the phrases used to complete the mode of transportation and destination name are connected to other pieces of the app, particularly the Profile page, where one version of the word might make sense but then does not when used in a sentence in another piece of the app. As an example, the app uses the Spanish word "trabajo" in the Profile page destination dropdown menu but that word should become "el trabajo" when used in a sentence and, if the word before it is "de", should become "del trabajo."  A separate task (#389) has been added to explore alternate ways of structuring the problematic phrases to allow for more flexible grammatical rules. 

**Peak/Off Peak/Peak No Express/Off Peak No Express Dropdown Menu**

The Peak/Off Peak/Peak No Express/Off Peak No Express Dropdown Menu has not been translated, as I ran into an issue with how the `Select` element in `form.js` generates the options and sends those options to network state. I tried a few different methods, such as adding the translate method into a mapping function that generates the network options to the Select component and translating within the Select function as the options are generated; however, there is a disconnect between how the option labels need to look to work with `i18next` and the option labels that state expects. Ultimately, for the purpose of this PR, I decided that this should be set aside as its own future task (#391) because we currently do not have translations for any of the options in Spanish and for two of the options in Chinese.

**Maintaining State Across Multiple Languages**

As of right now, the destination label in the "Show Travel Times: from" dropdown menu is always in the language that it was set in from the Profile page and not in the language that is currently active. This hints at a larger issue as to how state understands the same value but in a different language that may be related to the Peak/Off Peak menu issue related above; however, I'm not sure this particular manifestation is a huge issue, as in theory users will not be going back and forth between languages.

## Testing Instructions

 * Either run `./scripts/server` or `cd taui` &#8594; `yarn start`
 * Click on the "Español" button. Check the dock, neighborhood detail view, marker pop ups, and edit profile page to verify that any English has been addressed above.
 * Click on the "中文" button. Check the dock, neighborhood detail view, marker pop ups, and edit profile page to verify that any English has been addressed above.

Resolves #384


taui/src/locales/zh/translations.js
```diff
@@ -109,89 +115,88 @@ export default {
...
-    ViolentCrimeBoston: '<p><strong>{{town}}</strong> has <strong>{{averageRelation}}</strong> public safety, safer than <strong>{{crimePercentile}}</strong> of other zipcodes in the state. (Source: <a href="https://data.boston.gov/" target="_blank">Boston Police Department Incident Reports</a>)</p>',
-    RentalUnits: '<p><strong>{{town}}</strong> has an <strong>{{averageRelation}}</strong> number of rental units for the state. From 2017-2018 there were <strong>{{totalMapc}}</strong> units listed on Craigslist. (Source: <a href="https://www.mapc.org/" target="_blank">MAPC</a>)</p>'
+    AboveAverage: '高於平均水平',
+    Average: '相對平均水平',
+    BelowAverage: '低於平均水平',
+    Education: '<p>在參考過學校的整體質量和低收入家庭學生的表現，（如有低收入家庭學生表現的數據可使用），<strong>{{town}}</strong> 的學校在本州 <strong>{{averageRelation}}</strong> ，其排名高於[ ]的其他學區 <strong>{{edPercentile}}</strong> (資料來源: <a href="http://profiles.doe.mass.edu/" target="_blank">馬薩諸塞州中小學教育部 MA DESE</a>)</p>',
```
KlaasH, line 189: This happened to catch my eye, and it doesn't really make sense that the Chinese translation of "DOE" would be "DESE".  It looks like despite the URL being `doe`, the department is called the Department of Elementary and Secondary Education and the site uses DESE.  The source strings for the Chinese translation must have been pulled at a time when the English text said DESE, then it must have been changed later.
It kind of seems like it should be changed to DESE in all three translations files, though there might be more to the situation since I don't know why it would have been changed to DOE.

KlaasH (approved):
This looks good.  Clearly there are a number of lingering issues around translation, but the approach of focusing on the task at hand (getting the strings we have into the translations file) and spinning out the others makes sense.  

I think the last one you mentioned--the translated word in the "Show Travel Times: from" droplist not changing with the language selection--also deserves an issue (or to be added to #391, since they're right next to each other and might be similar enough that it would make sense to do them together).  There must be a way to reconfigure how that's built to make it respond to the selected language, and I don't think we should leave a case of mixed translation without at least having an issue about it.

There's one line comment below, which is just kind of a weird detail. Not sure if it's even worth making a thing out of, though on the other hand we could also avoid making a thing of it by just changing them all to DESE and calling it good.

So either way on that, and either a new issue or a revision to #391 for the "from" droplist, then this is good to go :+1:

taui/src/locales/zh/translations.js
ktohalloran: Yeah, I also noticed this and thought it was strange so thanks for catching it. I also agree it should be changed to DESE because the app uses DOE to refer to the US Dept of Energy in other places.

PR #392, opened 2021-12-15
Update routing on listing interaction

### LINKED ISSUE #357: Update routing when a user clicks on a displayed property
By default, the routing we display routes to the centroid of a selected neighborhood. When a user clicks on an apartment listing from Realtor.com / BHA, we should adjust the routing to route to the location of that apartment, rather than the neighborhood centroid.

## Notes
- As with #356 , this functionality already exists on [the demo site](https://demo.echosearch.org) -- the functionality should match, but isn't required to be an exact match, and we can reuse as much or as little of the demo code as we prefer. The demo code is on the `demo` branch.

rachelekm:
## Overview

Updates the routing so that a selected apartment listing is the destination, rather than the neighborhood centroid, on hover and on click. Clicking a listing’s marker currently redirects to a listing’s url in a new tab. In order to display the listing route while allowing for the click redirect, this PR displays route and trip info on marker hover with with a 500ms delay or immediately on click for when the user returns to the page. The listings’ routing was built following the same process for neighborhood routing—although simplified since there is only one selected listing and its route represented in state at a time—and successful routing includes: correct route path drawn on map, accurate trip info in Dock, and resetting routing when listing marker is no longer displayed.

### Demo
Route layer should update to apartment listing as destination:
Screen Shot 2021-12-15 at 11 23 06 AM
Trip info should update accordingly:
Screen Shot 2021-12-15 at 11 23 28 AM
If apartment listing is beyond `MAX_TRAVEL_TIME` or the trip is undefined, no route displays and an error message displays in Dock:
Screen Shot 2021-12-15 at 11 25 45 AM

### Notes

- Since there is a link to directions in the side Dock that users should be able to navigate to without losing the route/trip info, the `mouseout` action of a marker does not reset the route to the centroid of a neighborhood if a user has clicked the listing.
- Users have the option of selecting transit or car for transportation preference. There is no road network that can be used to create a routable "by car" journey and building support for that feature isn't a viable option at the time, so the current solution is to display a "drivable route" represented by a straight line from origin to destination. Similar to transit routing, destination is the neighborhood centroid by default and will update to an apartment listing when selected.

## Testing Instructions

- Run ./scripts/server
- Log in and select a neighborhood and select 'Show BHA' or 'Show Realtor' in the top bar
- Confirm that route updates to route to marker on hover and on click
- Confirm that trip info in Dock updates to route to marker on hover and on click
- Navigate to Haverhill – 01832 neighborhood and select Realtor listing that exceeds `MAX_TRAVEL_TIME` and confirm error message in Dock and no route appears
- On 'Hide BHA' or 'Hide Realtor' confirm that route resets to neighborhood centroid
- Navigate to Edit Profile, toggle between 'car' and 'Bus and/or train' to ensure routing accurately updates by transportation type

Resolves #357


lederer (changes requested):
Thanks for putting this together. 

Found one issue, and one requirement I miscommunicated…

When the user has a car, the route that gets generated to a listing is a transit route, but it should be a driving route. I think there were some cases where it might have properly generated a driving route? But I'm not sure and can't reliably recreate.

I called the 500ms timer a "delay" but that was an insufficient description. Essentially what's needed is: 
1. `onmouseover` begin the timer.
2. `onmouseout` clear the timer (canceling out the next step…)
3. When the timer finishes, perform the routing.

This results in the routing only happening if the cursor remains on the marker for the entire timer duration.

rachelekm:
@lederer thank you for catching those issues! I’ve fixed the 500ms timer so that the routing only happens if the cursor remains on the marker long enough or immediately if the user clicks on a listing. 
I’ve also adjusted the driving routes so that it no longer displays the incorrect transit routes, but the direct “driving” route we discussed: if the user is using a car, display a straight line to the neighborhood centroid by default and to the active listing when selected. I’ve also updated the notes to better communicate that change/test.

lederer (changes requested):
Thank you for fixing! Works as described, with the exception of two issues:
1. I found inconsistent behavior when mousing away from a listing marker after hovering on it long enough to select it. Sometimes when I hover away, the listing remains selected (which feels right), while sometimes when I hover away, the listing deselects and the route changes back to the centroid. This seems to be not per-listing but rather all-or-none, meaning that if any listing behaves this way, they all do. Not sure how to track down the conditions that lead to this behavior, but I did notice that when it behaves the "wrong" way, if I **click** a listing, they all then start behaving the "right" way.
2. When a listing is active, the sidebar Directions link attempts to route from Antarctica. I believe this is just due to the lat and long being swapped. It occurs for both car and transit (but only for listings, not for the centroid).

rachelekm:
Thanks @lederer! I had initially misinterpreted how `onmouseout` would clear the timer but after fixing looks like I didn't catch the leftover "click" conditional. That inconsistent behavior should be fixed now as well as the sidebar directions.

lederer (approved):

lederer:
Thanks for fixing so quickly. Works as described! 🎉

taui/src/components/map.js
```diff
@@ -160,6 +167,29 @@ class Map extends PureComponent<Props, State> {
     }
   }
 
+  /* Display listing routing on hover with 500ms delay, clear on mouseout.
```
KlaasH: "clear on mouseout" isn't exactly what's happening. Or at least, as I read it this is saying that the route would be reset to the centroid on mouseout, even if the mouse stayed long enough for it to be displayed. Just changing "clear" to "cancel" would probably suffice to make this description match what's happening (which I think is what's intended to happen).

taui/src/components/neighborhood-details.js
```diff
@@ -49,6 +51,47 @@ class NeighborhoodDetails extends PureComponent<Props> {
...
+    const { hasVehicle, listing, neighborhood, origin, userProfile, t } = props
+    // Look up the currently selected user profile destination from the origin
+    const originLabel = origin ? origin.label || '' : ''
+    const currentDestination = userProfile.destinations.find(d => originLabel.endsWith(d.location.label))
+    const bestJourney = listing ? (
+      listing.segments && listing.segments.length ? listing.segments[0] : null
+    ) : (
+      neighborhood.segments && neighborhood.segments.length ? neighborhood.segments[0] : null
+    )
+    const roundedTripTime = listing ? (Math.round(listing.time / ROUND_TRIP_MINUTES) * ROUND_TRIP_MINUTES) : (Math.round(neighborhood.time / ROUND_TRIP_MINUTES) * ROUND_TRIP_MINUTES)
```
KlaasH: Moving the ternary further in would make this do the same calculation a little more succinctly:
```suggestion
    const roundedTripTime = Math.round((listing ? listing.time : neighborhood.time) / ROUND_TRIP_MINUTES) * ROUND_TRIP_MINUTES
```

taui/src/selectors/detail-listing.js
```diff
@@ -0,0 +1,52 @@
...
+      return null
+    }
+    const useTransit = !profile || !profile.hasVehicle
+
+    const segments = useTransit ? ListingRoute.routeSegments : []
+    const time = useTransit ? travelTime : distanceTime(origin, listing)
+    // Lisitings called from within routable neighborhoods filtered within max travel time window,
+    // so it's assumed listings also within max travel time window.
+    // Smaller travel time is better; larger timeWeight is better (reverse range).
+    const timeWeight = time < MAX_TRAVEL_TIME ? scale(time, 0, MAX_TRAVEL_TIME, 1, 0) : 1
```
KlaasH: I might be reading this wrong, but if larger timeWeight is better, shouldn't the fallback value in the case where a time that's greater than MAX_TRAVEL_TIME somehow slips through be 0?

taui/src/selectors/network-listing-route.js
```diff
@@ -0,0 +1,80 @@
...
+  const foundKeys = {}
+  return (routeSegments || []).reduce((uniqueRoutes, route) => {
+    const key = routeToString(route)
+    if (!foundKeys[key]) {
+      foundKeys[key] = true
+      return [...uniqueRoutes, route]
+    }
+    return uniqueRoutes
+  }, [])
+}
```
KlaasH: It's kind of a bummer to have two copies of this somewhat-complex logic.  Is there somewhere we could at least put this top part, which is fully identical between them, so it could be defined once and shared?

The `createSelector` functions are a bit trickier, since there are differences in both the input object and the return structure.  Ideally we could have a shared `composeRoute` function to do the "check for completeness and build a route object" part, though it's possibly managing the differences in the inputs would be too cumbersome for that to come out well.

KlaasH (commented):
This looks good!  The interactions are very smooth and intuitive (at least everything besides the "we're going to pretend you drive in a straight line through the air" thing is intuitive).

I have a few questions and suggestions below, but none of them is a definite "this needs to change," so I'm marking this review as a comment.

taui/src/selectors/detail-listing.js
rachelekm: Thanks for this catch! You’re right that routes that exceed the `MAX_TRAVEL_TIME` should ultimately be filtered out and `timeWeight` should be 0 if any slip through. Actually, I included this variable to try and keep route time creation/route state consistent between neighborhoods and listings, but looking at this more closely `timeWeight` is really only used to create the `score` in which `neighborhoodsSortedWithRoutes` are ordered. Since we don’t create a recommended order of listings in a neighborhood, I think we can just get rid of the `timeWieght` variable here.

rachelekm:
Thanks for the review @KlaasH! I made the suggested changes. I went ahead and got rid of the `timeWeight` variable in `detail-listing.js` since it’s not needed, but left it as a separate commit if we want to revert and keep the listing route details consistent with neighborhood route details. I created a couple modules to share the repeated logic that you noted, as well as for the redundant `getSegmentPositions` and `getStopPositions` functions in the selectors to draw routes. I agree a shared `composeRoute` function would be ideal in the network route selectors, though I ultimately didn’t include it as it felt like enough conditionals to merit separate functions. Looking at all of the selectors again and trying to reduce repeated code, I think it may be feasible to create one `drawRoute` selector that would merge `draw-listing-route.js `and `draw-neighborhood-route.js` and ultimately simplify the `DisplayRouteLayers` component as well. However, I could see this taking a little bit of time to finagle. What are your thoughts?

KlaasH (approved):
:+1: Looks great!  Very nice consolidation/modularization of the repeated functions.  Stopping there and not trying to merge the mostly-but-not-entirely-repeated ones right now seems reasonable.

PR #397, opened 2022-01-18
UI polish

### LINKED ISSUE #388: Design Pass
Design pass for "Task 1 Priority Changes":

- [x] Logo update
- [x] Copy changes to profile page
- [ ] Signin error message text
- [x] Spanish and Chinese translations (buttons, text, layout implications)
- [x] BHA and Realtor.com unit map display (buttons, markers, popup, routing)
- [x] Listing routing (sidebar routing info, error message)

@ktohalloran and @rachelekm, please add things that I might have missed.

lederer:
## Overview

Refine styles of recent feature updates, including:
- Logo
- Translated strings
- Listings (toggles, markers, popups, routing, errors)

#### Logo

image

image

#### Translations

image

#### Listing toggles and markers

image

#### Listing popup

image

#### No routes

image

### Notes

- I opportunistically changed rent amounts to render with commas.
- In the detail sidebar, I had to split the stats and rent card into two cards to accommodate longer strings from translation. See screenshot above.
- I changed the listing markers to use the brand color from the new logo.
- I replaced the listing toggle buttons with checkboxes, which will require new translations for `ApartmentsToggles` (formerly `Apartments`): "Show apartments from" #398 
  - This also involved deprecating some strings: `HideBHAApartments`, `HideRealtorApartments`, `ShowBHAApartments`, `ShowRealtorApartments`
- I updates some error messages, which will require new translations #398:
  - `ListingsFetchError` "Could not fetch listings"
  - `TripsEmpty`: "No routes found to that destination"

## Testing Instructions

 * Open site when logged out to confirm new logo on auth page
 * Confirm new logo in main header
 * Cycle through all languages on all pages to confirm every component's layout adapts reasonably to string lengths
 * Confirm listing toggle checkboxes work correctly
 * Confirm popup looks right
   *  Including when a listing lacks photos. You can simulate by deleting the `` from a popup or adding `display:none` to its CSS.
 * Confirm "No routes found…" message in sidebar when attempting to route to a distant destination (eg, a listing in _Haverhill – 01832_ when using Transit).

Resolves #388


taui/src/components/route-card.js
```diff
@@ -102,7 +102,7 @@ class RouteCard extends React.PureComponent<Props> {
                   &nbsp;
                   {t('NeighborhoodDetails.FromOrigin')}
                 </span>
-                &nbsp;
+                {' '}
```
ddohler, line 105: Neat. I wasn't sure how to expect this to be turned into HTML but it just ends up as a "whitespace" node in the inspector.

taui/src/components/neighborhood-details.js
```diff
@@ -99,13 +99,12 @@ class NeighborhoodDetails extends PureComponent<Props> {
...
-      <div className='neighborhood-details__stats'>
-        <div className='neighborhood-details__rent'>
+      <div className='neighborhood-details__rent'>
+        <div>
           <div className='neighborhood-details__rent-label'>{t('NeighborhoodDetails.MaxRent')}</div>
-          <div className='neighborhood-details__rent-value'>${maxSubsidy}</div>
           <div className='neighborhood-details__rent-rooms'>{rooms}{t('NeighborhoodDetails.BedroomAbbr')}</div>
         </div>
-        <NeighborhoodListInfo neighborhood={neighborhood} />
+        <div className='neighborhood-details__rent-value'>${maxSubsidy.toLocaleString("en-US")}</div>
```
ddohler: This and the instance below should adjust depending on the user's chosen locale because different locales use different number formatting.

I think it should be sufficient to add `i18n` to line 97: `const {t, i18n} = useTranslation()`

and then to adjust this to be: `maxSubsidy.toLocaleString(i18n.language)`

If it turns out to be substantially more complicated than that then I'd ignore it because anyone using the app is most likely familiar with US number formatting conventions (because they live here, and number formatting is not hard to adjust to, in my experience).

taui/src/components/map.js
```diff
@@ -202,25 +202,16 @@ class Map extends PureComponent<Props, State> {
...
       }
-      <div className='map__popup-contents' style={{paddingTop: 20}}>
-        {rent && <h1>{t('Map.PopupDetails.RentAmount', { rent: rent })}</h1>}
-        {beds && <h2>{beds.split(' ')[0] + ' ' + t('NeighborhoodDetails.BedroomAbbr')}</h2>}
-        <div className='map__popup__line' />
-        {address && <h1>{address}</h1>}
-        <p>{t('Map.PopupDetails.ClickForDetails')}</p>
+      <div className='map__popup-details'>
+        {address && <div className='map__popup-address'>{address}</div>}
+        {rent && <div className='map__popup-rent'>{t('Map.PopupDetails.RentAmount', { rent: rent.toLocaleString("en-US") })}</div>}
```
ddohler: Same comment about switching number formatting with the language here.

taui/src/locales/zh/translations.js
```diff
@@ -72,20 +74,19 @@ export default {
...
-    MainSearchToolsLinksHeading: 'Search for {{rooms}}br with max rent $ {{maxSubsidy}}',
+    MainSearchToolsLinksHeading: 'Search for {{rooms}}br with max rent ${{maxSubsidy}}',
     MoreSearchToolsLinksHeading: '更多搜索工具',
+    RealtorApartmentsToggle: 'Realtor.com',
     RentEstimatorLink: 'Rent Estimator',
     Section8Link: 'Section 8',
-    ShowBHAApartments: '顯示 BHA',
-    ShowRealtorApartments: '顯示 Realtor.com',
+    // ShowBHAApartments: '顯示 BHA',
+    // ShowRealtorApartments: '顯示 Realtor.com',
```
ddohler:
```suggestion

```

ddohler (approved):
Looks good! I made a few comments for minor things, but they're all optional. The only other thing I noticed is that when mousing away from a listing popup, the little inverted triangle connecting the popup to the marker fades away at a different rate than the rest of the popup, resulting in a Cheshire-Cat-like effect. I tried to attach a video showing the issue but my screen capture program broke for mysterious reasons and I want to get this in before we start RRP--let me know if you can't reproduce the issue and I'll attach a recording once I can fix my tool.

lederer:
@ddohler Thanks for reviewing and the suggestions. I'm going to work on those now.

lederer:
> when mousing away from a listing popup, the little inverted triangle connecting the popup to the marker fades away at a different rate than the rest of the popup, resulting in a Cheshire-Cat-like effect

Nice catch. I see it as well now. I tried to track down the cause but I'm stumped. I'll open an issue.

PR #401, opened 2022-01-26
Maintain aspect ratio in neighborhood thumbnails

lederer:
## Overview

Some neighborhood thumbnails (eg, Fenway 02115) render with a different aspect ratio from the expected 4:3. This PR enforces the 4:3 aspect ratio.

### Before

image

### After

image

## Testing Instructions

 * Page through the neighborhood cards until you see "Boston - Fenway: 02115"
 * Confirm that the thumbnail has the same aspect ratio as other cards and the card layout is okay.

### DIFF
--- a/taui/src/sass/06_components/_neighborhood-summary.scss
+++ b/taui/src/sass/06_components/_neighborhood-summary.scss
@@ -65,6 +65,8 @@
     img {
       width: 100%;
       max-width: 120px;
+      aspect-ratio: 4 / 3;
+      object-fit: cover;
     }
   }
 


taui/src/sass/06_components/_neighborhood-summary.scss
```diff
@@ -65,6 +65,8 @@
     img {
       width: 100%;
       max-width: 120px;
+      aspect-ratio: 4 / 3;
+      object-fit: cover;
```
ddohler, line 69: Neat, hadn't seen this before.

ddohler (approved):
Works great!

PR #405, opened 2022-01-26
Fix error messages not appearing on login page

### LINKED ISSUE #395: Sign in error messages not rendering
When trying out the new sign in error messages (#367) during a polish pass (#388), I found the error messages weren't rendering at all.

### Step to repro
- Go to [staging](https://staging.echolocator.org/) or spin up `develop` locally
- Try to login with incorrect credentials

### Expected
- Banner with one of the following messages (depending on conditions):
  - "Could not find that username as entered. Usernames are case-sensitive."
  - "Username or password was incorrect. Both usernames and passwords are case-sensitive."

### Actual
- The banner appears but there is no message in it.

### Notes
- Some error messages do appear. Eg, when I tried a username with a space in it, I got "Custom auth lambda trigger is not configured for the user pool." in the banner. So this issue isn't fundamental to the banner or to error messaging.

ktohalloran:
## Overview

This PR fixes the issue with error messages not appearing correctly following a user entering incorrect login credentials.

### Demo
Screen Shot 2022-01-26 at 3 32 50 PM
Error message displayed when username does not exist.

Screen Shot 2022-01-26 at 3 33 31 PM
Error message displayed when password is incorrect.

### Notes

The `customMessages` dictionary was written to work with `react18next`; this PR also adds that translation capability into the `customAuthErrorMessageMap` function so that the correct message displays once we have corresponding errors in the Spanish and Chinese translation files.

## Testing Instructions

 * Either run `./scripts/server` or `cd taui` → `yarn start`
 * Try to log in with a fake username. You should see a banner at the top as in the image above.
 * Try to log in with a real username but a fake password. You should see a banner at the top as in the image above.

Resolves #395

### DIFF
--- a/taui/src/components/with-authenticator.js
+++ b/taui/src/components/with-authenticator.js
@@ -7,6 +7,7 @@ import { Authenticator } from 'aws-amplify-react/dist/Auth'
 import LogRocket from 'logrocket'
 import AmplifyMessageMap from 'aws-amplify-react/dist/AmplifyMessageMap.js'
 import message from '@conveyal/woonerf/message'
+import { Translation } from 'react-i18next'
 
 import {clearLocalStorage, storeConfig} from '../config'
 import {AMPLIFY_API_NAME,
@@ -27,7 +28,7 @@ function customAuthErrorMessageMap (error) {
   }
 
   if (customMessages.hasOwnProperty(error)) {
-    return message(customMessages[error])
+    return <Translation>{(t) => t(message(error, customMessages[error]))}</Translation>
   }
   // Fall back to the default messages if we can't find a custom one.
   return AmplifyMessageMap(error)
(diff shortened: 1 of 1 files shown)


taui/src/components/with-authenticator.js
```diff
@@ -27,7 +28,7 @@ function customAuthErrorMessageMap (error) {
   }
 
   if (customMessages.hasOwnProperty(error)) {
-    return message(customMessages[error])
+    return <Translation>{(t) => t(message(error, customMessages[error]))}</Translation>
```
ddohler: Nice use of `<Translation>` -- that's a great way to inject translation into just this one spot. However, I think we need to remove `message` here rather than preserve it -- I was surprised to see it, and it looks like this was my oversight when switching to `react-i18next` from `woonerf`. Oops (and sorry for leading you astray). I also found a few other instances where there are some vestigial `message` usages lying around and made #406 to get rid of them.

ddohler (changes requested):
I think this needs to be structured slightly differently, but it should be a quick change!

ddohler (approved):
Works great! 🎉

PR #408, opened 2022-01-27
Add tooltips to profile page

### LINKED ISSUE #404: Add tooltips to profile page matching demo.echosearch.org
On demo.echosearch.org there are tooltips on the profile page that explain the meanings of the Commute Time, School Quality, and Public Safety preferences. We should replicate these on the profile page.

rachelekm:
## Overview

Creates tooltips on the profile page for the following importance preferences: Commute Time, School Quality, and Public Safety. Aside from small design changes detailed in notes, this was intended to replicate the tooltips on the demo site. Translations for the tooltips were found in the translations file, so those have been added as well.

### Demo
Tooltips should display to the right on icon hover and be properly translated.
Screen Shot 2022-01-27 at 4 12 29 PM
Screen Shot 2022-01-27 at 4 12 20 PM

### Notes

- I made these icons centered between the label and the options select input. However, because each label text is of slightly different length they are not vertically aligned. I left it because it matches the demo site more closely, but can set to just `margin-left:auto` so that they are not reliant on the length of the label and will all be vertically aligned. EDITED: icons now sit on the right of select.

- I moved the tooltips to displaying on the right of the icon instead of the top so that a user can more easily hover up to next icon, an action previously blocked by the appearing tooltip.

-  I made the classNames for these tooltips the same as the tooltips elsewhere in the app to keep styling consistent.

## Testing Instructions

 * Run ./scripts/server
 * Login and navigate to '/profile'
 * Confirm tooltips appear and match the demo site
 * Select different translations and make sure each tooltip appears and has proper translations

Resolves #404


lederer (changes requested):
Awesome. Thanks for doing this. A few style adjustments:
- Move it to the right side of the `<select>` instead of in-between
- Change the color to `$gray-500`
- Change the font-size to `1.8rem`

rachelekm:
Thanks, @lederer! Just implemented those changes and updated the demo/notes, let me know if there's anything else.

taui/src/components/edit-profile.js
```diff
@@ -594,6 +596,31 @@ class EditProfile extends PureComponent<Props> {
     )
   }
 
+  importanceTooltip (props) {
```
ddohler, line 599: Good call making a separate component here.

ddohler (approved):
Looks great! Glad this was straightforward!

lederer (approved):

lederer:
Thanks for the tweaks! 💯

PR #409, opened 2022-01-28
Filter listings by neighborhood max rent

### LINKED ISSUE #402: Use neighborhood max rent when showing apartment listings
When showing listings on the map, the app should filter listings by the max rent defined for the neighborhood / number of rooms the user has specified. This is the same number that's shown in the sidebar: 
Screen Shot 2022-01-26 at 11 11 36 AM

This applies to both the Realtor.com listings and the BHA listings. There is already a parameter for the rent being passed to the API but it's set to $10000 in all cases.

rachelekm:
## Overview

This filters both BHA and Realtor listings according to the max budget, defined by the number of bedrooms and neighborhood chosen, displayed within the neighborhood details sidebar.

### Notes

I originally created a "get the budget" function in the `actions/listings.js` file that would call the utility function `getNeighborhoodById()` so that the correct neighborhood's properties could be referenced to get the max rent and populate the API query. However, since that logic is a reproduction of the `detailNeighborhood` selector and the resulting budget value is also used to display the max rent in the sidebar, it felt worthwhile to instead create a separate `estMaxRent` selector that could be applied broadly. I then moved the logic to build the query outside of the fetch listings action so that `getState()` doesn't need to be called and there is less repeatable code.

## Testing Instructions

 * Run `./scripts/server`
 * Log in and select a neighborhood
 * Confirm correct max rent displays in sidebar and used in links (Zillow, Craigslist, Hotpads, GoSection8) as well as links section header — neighborhood max rent can be found in `neighborhoods.json` or by printing `neighborhood.properties`
 * Show Realtor and BHA listings and confirm no listing exceeds the max rent displayed in sidebar 

Resolves #402


taui/src/selectors/est-max-rent.js
```diff
@@ -0,0 +1,16 @@
...
+import detailNeighborhood from './detail-neighborhood'
+
+export default createSelector(
+  state => get(state, 'data.userProfile'),
+  detailNeighborhood,
+  (userProfile, neighborhood) => {
+    if (!userProfile || !neighborhood) {
+      return 0
+    }
+    return neighborhood && userProfile && neighborhood.properties['max_rent_' + userProfile.rooms + 'br']
```
ddohler: The guards on this line shouldn't be necessary because of the guard clause above that's returning 0, but they're also not going to protect against every case because `userProfile.rooms` and `neighborhood.properties.max_rent_1_br` could also be undefined. I'm not sure whether we have [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) available on this project, but that might be one way to protect this better. Otherwise I think adding guards for those other two properties should do the trick.

ddohler (approved):
I had one minor comment but overall this looks good and is a nice refactor of the logic involved, nice work!

taui/src/selectors/est-max-rent.js
rachelekm: Ah, yup, that makes sense. Looks like optional chaining isn't available (it's interpreting my `?` as a ternary operator) so I expanded the if statement above and hopefully that should cover it more effectively.

PR #410, opened 2022-02-01
Add translation support to dock dropdown menus

### LINKED ISSUE #391: Translate text in dock dropdown menus
The wiring for translating and majority of translation strings have been added (PRs [377](https://github.com/azavea/echo-locator/pull/377) and [390](https://github.com/azavea/echo-locator/pull/390)); however, the "from destination" and Peak/Off Peak/Peak No Express/Off Peak No Express dropdown menus in the neighborhood detail view dock have not been translated. For the "from destination" menu, there seems to be an issue with how state understands state values set in one language and then displayed in another, as the language of the destination tag always matches the language in which the destination was added, rather than the language controlled by the toggle. For the Peak/Off Peak menu, there seems to be a disconnect between how the option labels need to look to work with `i18next` and the option labels that state expects. Overall, it may be worth exploring how `i18next` and state work (or don't) together in the [React-i18next documentation](https://react.i18next.com/).

Following identification of and a solution for the disconnect, relevant translation strings also need to be added to each language's relevant translation file for the Peak/ Off Peak menu (all destinations already have translations). As a note, as of the creation of this issue, we only had translations for "Peak" (什麼時候高峰期) and "Off Peak" (非高峰) in Chinese. Additional translations are needed for all 4 options in Spanish and for the "No Express" options in Chinese.

ktohalloran:
## Overview

This PR adds translation support to the 'from' and 'network' dropdown menus in the dock. While translation support was added in [PR 377](https://github.com/azavea/echo-locator/pull/377) and translations strings were added in for most of the  the website in [PR 390](https://github.com/azavea/echo-locator/pull/390), translating the menus was more complex due to the connection between those menus and state. 

### Demo

Screen Shot 2022-02-01 at 8 13 06 AM

Screen Shot 2022-02-01 at 8 13 16 AM

The 'from' dropdown menu translated into both Chinese and Spanish.

Screen Shot 2022-02-01 at 8 13 28 AM

The 'network' menu translated into Chinese. There are currently only 2 translated options for this menu; placeholder English strings have been used for all 4 options in Spanish and the "Peak No Express" and "Off Peak No Express" options in Chinese.

### Notes

**Consolidation**
Initially, the addition of the `'TripPurpose.'` prefix used for `react-18next` to reference the relevant translation was being done in 3 different spots: `neighborhood-detail.js`, `route-card.js`, and `edit-profile.js`, with state storing only the suffix and dropping the prefix. This has been centralized to only occur in `edit-profile.js`, where new profile destinations are saved and then passed around state with the prefix already attached.

**Method of Adding Translation Support**
The content in these menus was not being translated properly before because the destination name was being saved to state as a translated string; as a result, `react-18next` could not find it to translate it on language change. This PR fixes the problem by reconstructing the `destination` object type to store label (destination address), value (destination purpose), and position. This is then used the to construct and temporarily replace the label with the `<name> : <label>` string used as the menu options, which is then taken back apart upon being sent to state, rather than storing the translated label string. A similar tactic was taken with the network menu, where label is replaced temporarily by the translated string for populating the menu but restored to its former name on being saved to state via a new 3rd `networkName` property on the `network` object. I have added comments into the code to explain this process, as well.

This approach worked for everything except for profiles that had been made prior to this addition, as their destination purpose was saved without the `TripPurpose` prefix. This was fixed by adding in logic that checks for older profiles and adds the necessary prefix onto it for proper translation.

## Testing Instructions

 * Either run `./scripts/server` or `cd taui` → `yarn start`
 * Log in and add a couple of new destinations to your profile.
 * Since we are missing most translations for the 'network' menu, it may be helpful to add some kind of temporary and distinguishing marker to the relevant placeholder translation strings in the Spanish and Chinese translation files, e.g. I added ', es' and ', ch' to the end of the strings for the sake of testing.
 * Click on the translation buttons&mdash;you should see the value in both menus change. Select the menu and verify that all of the values in the menu have been translated, as well.
 * One of the issues I ran into was that while the default value and menu options were translating as they should, an option, once selected, would not translate on language change. Verify that this also works by selecting an option from the menu and then changing the language.
 * To check that the logic around old profiles works, checkout `develop` and add a new destination. Then return to this branch and check that your new destination translates as it should.

Resolves #391

### DIFF
--- a/taui/src/components/edit-profile.js
+++ b/taui/src/components/edit-profile.js
@@ -482,7 +482,7 @@ class EditProfile extends PureComponent<Props> {
     const options = PROFILE_DESTINATION_TYPES.map((key) => {
       // expects each type in constants to have a label in messages
       const messageKey = 'TripPurpose.' + key
-      return <option key={messageKey} value={messageKey}>{t(messageKey)}</option>
+      return <option key={key} value={key}>{t(messageKey)}</option>
     })
 
     return (
--- a/taui/src/components/edit-profile.js
+++ b/taui/src/components/edit-profile.js
@@ -482,7 +482,7 @@ class EditProfile extends PureComponent<Props> {
     const options = PROFILE_DESTINATION_TYPES.map((key) => {
       // expects each type in constants to have a label in messages
       const messageKey = 'TripPurpose.' + key
-      return <option key={key}>{t(messageKey)}</option>
+      return <option key={messageKey} value={messageKey}>{t(messageKey)}</option>
     })
 
     return (
--- a/taui/src/components/form.js
+++ b/taui/src/components/form.js
@@ -99,10 +99,13 @@ class Form extends React.PureComponent<Props> {
       return
     }
     const position = destination.location.position
+    const translatablePurpose = destination.purpose.startsWith('TripPurpose')
+      ? destination.purpose
+      : 'TripPurpose.' + destination.purpose
     return (position.lat !== 0 && position.lon !== 0) ? {
       label: destination.location.label,
       position: position,
-      value: destination.purpose
+      value: translatablePurpose
     } : null
   }
 
@@ -136,10 +139,13 @@ class Form extends React.PureComponent<Props> {
     const {destination, network} = this.state
     const destinations: Array<AccountAddress> = userProfile ? userProfile.destinations : []
     const locations = destinations.map(d => {
+      const translatablePurpose = d.purpose.startsWith('TripPurpose')
+        ? d.purpose
+        : 'TripPurpose.' + d.purpose
       return {
         label: d.location.label,
         position: d.location.position,
-        value: d.purpose
+        value: translatablePurpose
       }
     })
     const locationsWithLabels = locations.map(loc => {
--- a/taui/src/components/form.js
+++ b/taui/src/components/form.js
@@ -100,9 +100,9 @@ class Form extends React.PureComponent<Props> {
     }
     const position = destination.location.position
     return (position.lat !== 0 && position.lon !== 0) ? {
-      label: destination.purpose + ': ' + destination.location.label,
+      label: destination.location.label,
       position: position,
-      value: position
+      value: destination.purpose
     } : null
   }
 
@@ -116,9 +116,9 @@ class Form extends React.PureComponent<Props> {
 
   selectDestination = (option?: ReactSelectOption) => {
     const destinationObj = option ? {
-      label: option.label,
+      label: option.label.split(':')[1],
       position: lonlat(option.position),
-      value: option.position
+      value: option.value
     } : null
     this.setState({destination: destinationObj})
     this.props.updateOrigin(destinationObj)
@@ -137,12 +137,15 @@ class Form extends React.PureComponent<Props> {
     const destinations: Array<AccountAddress> = userProfile ? userProfile.destinations : []
     const locations = destinations.map(d => {
       return {
-        label: d.purpose + ': ' + d.location.label,
+        label: d.location.label,
         position: d.location.position,
-        value: d.location.position
+        value: d.purpose
       }
     })
-    const destinationFilterOptions = createDestinationsFilter(locations)
+    const locationsWithLabels = locations.map(loc => {
+      return {...loc, label: t(loc.value) + ': ' + loc.label}
+    })
+    const destinationFilterOptions = createDestinationsFilter(locationsWithLabels)
     const useNetworks = this.getProfileNetworks(this.props.networks, userProfile)
     const networks = useNetworks.map(n => ({label: n.name, value: n.url}))
     const networkFilterOptions = createNetworksFilter(networks)
@@ -158,13 +161,13 @@ class Form extends React.PureComponent<Props> {
             className='map-sidebar__select'
             clearable={false}
             filterOptions={destinationFilterOptions}
-            options={locations}
+            options={locationsWithLabels}
             optionHeight={SELECT_OPTION_HEIGHT}
             onChange={this.selectDestination}
             placeholder={t('Geocoding.StartPlaceholder')}
             style={SELECT_STYLE}
             wrapperStyle={SELECT_WRAPPER_STYLE}
-            value={destination}
+            value={{...destination, label: t(destination.value) + ': ' + destination.label}}
           />
         </div>
         {!userProfile.hasVehicle && <div className='map-sidebar__field'>
--- a/taui/src/components/form.js
+++ b/taui/src/components/form.js
@@ -48,6 +48,9 @@ class Form extends React.PureComponent<Props> {
 
     const useNetworks = this.getProfileNetworks(networks, userProfile)
 
+    // TODO: Add networkKey for react-18next's use to the networks as they
+    // come into the app in store.yml and config.json, rather than adding this
+    // 3rd networkName value here (https://github.com/azavea/echo-locator/issues/411)
     this.state = {
       destination,
       network: useNetworks ? {
@@ -99,15 +102,10 @@ class Form extends React.PureComponent<Props> {
       return
     }
     const position = destination.location.position
-    // check for older profiles that contain destination purposes not saved with
-    // the 'TripPurpose.' prefix required by react-i18next
-    const translatablePurpose = destination.purpose.startsWith('TripPurpose')
-      ? destination.purpose
-      : 'TripPurpose.' + destination.purpose
     return (position.lat !== 0 && position.lon !== 0) ? {
       label: destination.location.label,
       position: position,
-      value: translatablePurpose
+      value: destination.purpose
     } : null
   }
 
@@ -116,14 +114,12 @@ class Form extends React.PureComponent<Props> {
     const first = useNetworks[0]
     const network = {label: first.name, value: first.url, networkName: first.name}
     this.setState({network})
-    this.props.setActiveNetwork(network.name)
+    this.props.setActiveNetwork(network.label)
   }
 
   selectDestination = (option?: ReactSelectOption) => {
     const destinationObj = option ? {
-      // restore the label to contain only the address and remove translated
-      // destination name for re-generation
-      label: option.label.split(':')[1],
+      label: option.label,
       position: lonlat(option.position),
       value: option.value
     } : null
@@ -143,20 +139,15 @@ class Form extends React.PureComponent<Props> {
     const {destination, network} = this.state
     const destinations: Array<AccountAddress> = userProfile ? userProfile.destinations : []
     const locations = destinations.map(d => {
-      // check for older profiles that contain destination purposes not saved with
-      // the 'TripPurpose.' prefix required by react-i18next
-      const translatablePurpose = d.purpose.startsWith('TripPurpose')
-        ? d.purpose
-        : 'TripPurpose.' + d.purpose
       return {
         label: d.location.label,
         position: d.location.position,
-        value: translatablePurpose
+        value: d.purpose
       }
     })
     const locationsWithLabels = locations.map(loc => {
       // generate temporary, translated destination labels menu options
-      return {...loc, label: t(loc.value) + ': ' + loc.label}
+      return {...loc, label: t(`TripPurpose.${loc.value}`) + ': ' + loc.label}
     })
     const destinationFilterOptions = createDestinationsFilter(locationsWithLabels)
     const useNetworks = this.getProfileNetworks(this.props.networks, userProfile)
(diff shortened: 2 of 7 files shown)


taui/src/components/edit-profile.js
```diff
@@ -482,7 +482,7 @@ class EditProfile extends PureComponent<Props> {
     const options = PROFILE_DESTINATION_TYPES.map((key) => {
       // expects each type in constants to have a label in messages
       const messageKey = 'TripPurpose.' + key
-      return <option key={key}>{t(messageKey)}</option>
+      return <option key={messageKey} value={messageKey}>{t(messageKey)}</option>
```
ddohler: This works but it has the side-effect (as you noticed) that existing profiles become incompatible with the new code and need to be migrated. You should be able to achieve the same change, with fewer side-effects, by doing this instead:
```suggestion
      return <option key={key} value={key}>{t(messageKey)}</option>
```

taui/src/components/form.js
```diff
@@ -99,26 +99,33 @@ class Form extends React.PureComponent<Props> {
       return
     }
     const position = destination.location.position
+    // check for older profiles that contain destination purposes not saved with
+    // the 'TripPurpose.' prefix required by react-i18next
+    const translatablePurpose = destination.purpose.startsWith('TripPurpose')
+      ? destination.purpose
+      : 'TripPurpose.' + destination.purpose
```
ddohler: I think this can be avoided by making the suggested changes to `edit-profile.js` and using `destination.purpose` directly for `value` below.
```suggestion

```

taui/src/components/form.js
```diff
@@ -99,26 +99,33 @@ class Form extends React.PureComponent<Props> {
...
     return (position.lat !== 0 && position.lon !== 0) ? {
-      label: destination.purpose + ': ' + destination.location.label,
+      label: destination.location.label,
       position: position,
-      value: position
+      value: translatablePurpose
```
ddohler:
```suggestion
      value: destination.purpose
```

taui/src/components/form.js
```diff
@@ -136,15 +143,25 @@ class Form extends React.PureComponent<Props> {
     const {destination, network} = this.state
     const destinations: Array<AccountAddress> = userProfile ? userProfile.destinations : []
     const locations = destinations.map(d => {
+      // check for older profiles that contain destination purposes not saved with
+      // the 'TripPurpose.' prefix required by react-i18next
+      const translatablePurpose = d.purpose.startsWith('TripPurpose')
+        ? d.purpose
+        : 'TripPurpose.' + d.purpose
```
ddohler: Same comment here about removing this (which also cuts down on the duplication of having this in two places in the same file):
```suggestion
```

taui/src/components/form.js
```diff
@@ -136,15 +143,25 @@ class Form extends React.PureComponent<Props> {
...
       return {
-        label: d.purpose + ': ' + d.location.label,
+        label: d.location.label,
         position: d.location.position,
-        value: d.location.position
+        value: translatablePurpose
```
ddohler:
```suggestion
        value: destination.purpose
```

taui/src/components/form.js
```diff
@@ -136,15 +143,25 @@ class Form extends React.PureComponent<Props> {
...
       }
     })
-    const destinationFilterOptions = createDestinationsFilter(locations)
+    const locationsWithLabels = locations.map(loc => {
+      // generate temporary, translated destination labels menu options
+      return {...loc, label: t(loc.value) + ': ' + loc.label}
```
ddohler: If you make the other suggested changes then you'll also have to change this in order to generate the translation key correctly because it'll no longer be completely stored in the destination.
```suggestion
      return {...loc, label: t(`TripPurpose.${loc.value}`) + ': ' + loc.label}
```

taui/src/components/form.js
```diff
@@ -158,13 +175,13 @@ class Form extends React.PureComponent<Props> {
...
             filterOptions={destinationFilterOptions}
-            options={locations}
+            options={locationsWithLabels}
             optionHeight={SELECT_OPTION_HEIGHT}
             onChange={this.selectDestination}
             placeholder={t('Geocoding.StartPlaceholder')}
             style={SELECT_STYLE}
             wrapperStyle={SELECT_WRAPPER_STYLE}
-            value={destination}
+            value={{...destination, label: t(destination.value) + ': ' + destination.label}}
```
ddohler: This is a good opportunity for cleanup of the existing code -- the [`react-select` docs](https://github.com/JedWatson/react-select/tree/v1.x#usage) indicate that `value` should be "either a string or number" and should get passed to the `Select` element whenever the selected option changes, so I think it should be safe to use just `destination.value` here:
```suggestion
            value={destination.value}
```

taui/src/components/form.js
```diff
@@ -179,7 +196,8 @@ class Form extends React.PureComponent<Props> {
             placeholder={t('Map.SelectNetwork')}
             style={SELECT_STYLE}
             wrapperStyle={SELECT_WRAPPER_STYLE}
-            value={network}
+            // convert network name to match react-i18next's reference phrase and translate
+            value={{...network, label: t('Map.NetworkOptions.' + network.networkName.split(' ').join(''))}}
```
ddohler, line 190: Since this also uses `react-virtualized-select` I _think_ it should be possible to just use `network.url` here.

taui/src/components/neighborhood-details.js
```diff
@@ -76,7 +76,7 @@ class NeighborhoodDetails extends PureComponent<Props> {
           {t('Units.Mins')}&nbsp;
           <ModesList segments={bestJourney} />&nbsp;
           {t('NeighborhoodDetails.FromOrigin')}&nbsp;
-          {currentDestination && t('TripPurpose.' + currentDestination.purpose).toLowerCase()}
+          {currentDestination && t(currentDestination.purpose).toLowerCase()}
```
ddohler: If `TripPurpose` is no longer part of `destination.purpose` then this change should no longer be necessary.

taui/src/components/route-card.js
```diff
@@ -104,7 +104,7 @@ class RouteCard extends React.PureComponent<Props> {
                 </span>
                 {' '}
                 <span className='neighborhood-summary__location'>
-                  {currentDestination && t('TripPurpose.' + currentDestination.purpose).toLowerCase()}
+                  {currentDestination && t(currentDestination.purpose).toLowerCase()}
```
ddohler: Ditto here, if `TripPurpose` is kept out of `destination.purpose` then this can be left as it was.

taui/src/components/form.js
```diff
@@ -136,15 +143,25 @@ class Form extends React.PureComponent<Props> {
...
-    const destinationFilterOptions = createDestinationsFilter(locations)
+    const locationsWithLabels = locations.map(loc => {
+      // generate temporary, translated destination labels menu options
+      return {...loc, label: t(loc.value) + ': ' + loc.label}
+    })
+    const destinationFilterOptions = createDestinationsFilter(locationsWithLabels)
     const useNetworks = this.getProfileNetworks(this.props.networks, userProfile)
-    const networks = useNetworks.map(n => ({label: n.name, value: n.url}))
+    // generate temporary, translated network labels for menu options
+    const networks = useNetworks.map(n => ({label: t('Map.NetworkOptions.' + n.name.split(' ').join('')), value: n.url, networkName: n.name}))
```
ddohler, line 155: It would be worth checking to see if you can add a `networkKey` to the network configuration in `taui/configurations/default/store.yml` -- that defines the initial state of the Redux state, so if it works to add `networkKey` there, you should be able to define the translation key at that point, which would avoid the need to strip spaces from the network's `name` field in order to get the appropriate translation key.

In general, if you find yourself manipulating or parsing strings that were intended for human consumption in order to transform them into something programmatically meaningful, that's a pretty strong signal that something should be changed about the application's data model. The reason for this rule of thumb is that parsing human-readable strings often creates implicit requirements that can lead to errors or frustration down the line. In this case, stripping the spaces creates an implicit requirement that "all network translation keys must be exactly equal to the name of the network with internal spaces removed" which could be a source of confusion for someone trying to add a new network in the future if they don't know about that rule.

All that said, if it's not straightforward to add a new field to the network objects from the get-go for some reason, then this may be the best we can do.

taui/src/components/form.js
```diff
@@ -99,26 +99,33 @@ class Form extends React.PureComponent<Props> {
...
-    this.props.setActiveNetwork(network.label)
+    this.props.setActiveNetwork(network.name)
   }
 
   selectDestination = (option?: ReactSelectOption) => {
     const destinationObj = option ? {
-      label: option.label,
+      // restore the label to contain only the address and remove translated
+      // destination name for re-generation
+      label: option.label.split(':')[1],
```
ddohler: I'm not sure whether this is necessary -- I couldn't find a place where `destination.label` is getting used. Does it break anything to revert this change and use `option.label` here?

ddohler (changes requested):
Thanks for tracing through the data flow here! This works as-is, but I think the scope of the necessary changes could be reduced quite a bit with a tweak to how the options are handled in `edit-profile.js`. I'm reasonably sure that the changes I've suggested work but I may have missed something, so please let me know if anything seems off!

taui/src/components/form.js
ktohalloran: While setting the Select menu `value` to `destination.value` did work for the 'from' menu, setting the Select value to `network.url` here only partially did in that it does allow you to update your network but it does not display a default value or the selected value, using the 'Set time of day' placeholder message instead. Despite trying for quite a while to figure out the difference between the two, I'm still not entirely sure why; my best guess is that it comes down to the difference between the location menu's `updateOrigin` vs the network menu's `setActiveNetwork` functions where `updateOrigin` is being given the entire `destination` object and `setActiveNetwork` is only being given a string. My guess is that those functions eventually pass these values back to the menus to display both the default value on page load and following a selection and if state only has a piece of the object, as in this menu, it displays the placeholder option instead.

That being said, what's the best way to move forward with this? I can keep exploring to figure out exactly what's happening or attempt to get `setActiveNetwork` to take in more than a string; what do you think are next steps with this?

taui/src/components/form.js
ktohalloran: This has been surprisingly difficult to accomplish. I added a `networkKey` value to both `store.yml` and `taui/config.json`; however, somewhere in between state and the `Form` component, it loses the property. I can see that in `taui/src/actions/network.js`, the property exists on `state.data.networks`, but it does not exist on `json.networks` or on `c.networks` as a part of the `dispatch` action on line 90. My thought was perhaps it had something to do with what was being stored in local storage, which I tested by logging out and therefore triggering `clearLocalStorage`; that didn't work, so I thought perhaps it was something unique to configurations attached to the `TAUI_CONFIG_KEY` and other profile keys that were being passed into `retrieveConfig` in `network.js`. I tested this theory by creating a new profile with the new property on the networks, but that also did not work. Do you have any suggestions on what to try next?

taui/src/components/form.js
ddohler: Darn, that's too bad. I don't think we need to spend a ton of effort on this particular point; it's at odds with the docs, but aside from that I don't see any particular problem with passing in the whole object here. Thanks for looking into it!

taui/src/components/form.js
ddohler: Thanks for digging in! I think this definitely qualifies as "not straightforward" at this point -- it looks like the network data from those files is getting swapped out with data from another source (probably the `.dat` files that it downloads, though I haven't confirmed that), so we'll have to figure out where this is happening and then how to thread through the data from the configuration. Let's make a technical debt issue for it, add a TODO to the code and revisit it toward the end of the project if we have time.

ktohalloran: Ok, great--I can make that technical debt issue now. Thank you!

ktohalloran:
Thanks for looking at this! I've made the changes requested above and submitted a new issue for the network configuration so it's ready for re-review.

taui/src/components/form.js
```diff
@@ -48,10 +48,13 @@ class Form extends React.PureComponent<Props> {
 
     const useNetworks = this.getProfileNetworks(networks, userProfile)
 
+    // TODO: Add networkKey for react-18next's use to the networks as they
+    // come into the app in store.yml and config.json, rather than adding this
+    // 3rd networkName value here
```
ddohler: When adding a TODO / issue pair like this we should also include a link to the tracking issue so that it's clear where to go if someone bumps into this in the code.

ddohler (approved):
Looks good! I had one tiny cleanup comment but this is good to go after addressing that. Thanks for working through some of the tricky bits of this!

PR #413, opened 2022-02-18
Add ADR for Django architecture and deployment

### LINKED ISSUE #396: Plan Django deployment and user migration
In order to migrate users to the new Django authentication system, we'll have to migrate users from the old system. This sets up a situation where we potentially have a chain of dependent tasks that we'll need to complete to successfully deploy. I spoke with @jwalgran today and Civic Apps seems to have had a good experience using Auth0 as an authentication provider that then creates users in Django, so that might be worth looking into because it might allow us to avoid implementing email confirmation and an authentication interfaces from scratch.

This task is to figure out:
- What steps are required in order to successfully migrate all users to the Django authentication system
- What order the steps should be done in
- Whether any downtime will be required or whether any precautions need to be taken during certain steps in order to avoid data loss.
- Whether we have multiple options for approaching this (we almost certainly do), and choose a preferred option.

Here's a list, in no particular order, of the steps that I know will need to be taken at some point in order to carry out the deployment; this is most likely _not_ a complete list:
- Deploy Django with API that provides required authentication / profile update capabilities (possibly by leveraging Auth0)
- Deploy frontend updated to use Django (Auth0?) authentication API
- Copy all counselor accounts to new Django database (or maybe to Auth0?)
- Copy all non-counselor accounts (and associated profile information) to new Django database
- Notify copied user accounts that they need to reset their passwords
- Test deployment steps on staging (the staging authentication is only partially functional, but we should test what we can).

The output of this task should be a deployment plan. This will most likely take the form of at least one writeup in a document, but may also include creating issues with dependencies for the various tasks.

## Notes
See #394 for information on how the existing authentication and profile system works.

ddohler:
## Overview

This is an ADR evaluating whether to migrate to a Django-only authentication solution or a Django+Auth0 solution, and then providing a high-level work plan for accomplishing said migration.

### Demo

N/A

## Testing Instructions

 * Read through the ADR and work plan with the following questions in mind:
  * Are the two options clearly enough explained?
  * Are any major differences missing?
  * Are any major pros or cons missing?
  * Are important steps missing from the work plan or are any steps unclear?
  * Do you agree with the proposed decision? If not, why not?

Resolves #396

### DIFF
--- a/doc/arch/adr-001-identity-and-authorization.md
+++ b/doc/arch/adr-001-identity-and-authorization.md
@@ -1,0 +1,136 @@
+# Identity and Authorization
+
+## Context
+
+_Describe the problem and provide any relevant background information._
+
+ECHO's identity and authorization system suffers from two significant shortcomings:
+1. Due to funding restrictions on the initial project, it was developed using an unfamiliar technical stack (AWS Amplify, Cognito, and Lambda) that turned out to be somewhat inflexible and difficult to administer.
+2. The initial identity / authorization requirements of the application were unusual and somewhat complicated, because while Cognito uses emails as user identifiers, the application also required the ability to identify users via voucher numbers, and also allows counselors to view and edit all profiles. This made the identity and authorization handling code quite complex and error-prone. These requirements have since been dropped and the client desires to switch to a standard email-based identity system with voucher numbers tracked only as metadata.
+
+_What do we need to decide?_
+
+We've already decided to migrate the application backend to Django, for a variety of reasons, and had planned to migrate the identity and authorization parts of the app to Django as well. However, recent discussion with Civic Apps exposed that they've been very happy with their use of Auth0 with Django for some of their recent projects, so using Auth0 + Django seemed worth considering. Additionally, we need to plan how to migrate user data from the existing system to the new one.
+
+_Which options exist?_
+
+No matter which option we pick, it will impact the work we do in three distinct phases:
+
+1. How we implement an identity / authorization system into the application using the new technology
+2. How we migrate the existing users and their profile data to the new system
+3. How users interact with the finished system and how we maintain the system long-term
+
+Therefore, we need to consider how each of our options would impact these three phases of work.
+
+The two primary options we are considering are:
+
+1. Using Django alone as an authentication provider
+2. Using Auth0 as an identity provider and integrating it with Django's authorization system
+
+_Evaluate each option_
+
+### Django alone
+
+#### Implementation
+
+We would need to implement a database model or models to store users and their profile data. We would need to create API endpoints for signing up, email verification, logging in, logging out, changing passwords, resetting passwords, and updating the user's profile. We would need to add integration with an email service such as SES in order to support email verification and password reset flows and create appropriate email templates.
+
+We would also need to replace the existing frontend components and logic (currently written in customized Amplify React components) with custom React components or Django templates. These components/templates would need to support internationalization similarly to the rest of the application.
+
+#### Migration
+
+We would need to create a new user account in Django for each Cognito user. We would need to copy the profile data for each user from S3 and associate it with the appropriate Django user account. This would most likely be accomplished by some type of script that could run as a management command. Once the application is deployed, we would need to email all users instructing them to reset their password. Issues requiring manual intervention could be addressed by using the Django admin.
+
+#### User interaction and maintenance
+
+Users would most likely experience some visual changes to the login screen and different email messages. Administrators would have to learn a new administration system, but the Django admin offers many more features than the Cognito console, so this would be an improvement.
+
+It's not expected that there would be additional maintenance beyond the usual required to operate a Django instance (i.e. periodic Django and Python package updates along with Postgresql updates).
+
+### Django + Auth0
+
+#### Implementation
+
+We would need to set up an Auth0 account for the project. We would need to implement a database model or models to store users and their profile data (in this setup we are envisioning using Auth0 as an identity provider only, so Auth0 handles authentication, but does not store any profile information--instead, it sends a unique user ID to the Django app whenever a user logs in, and Django is responsible for associating the user's profile information with that user ID. Auth0 _can_ also store profile information, but it's just stored as a JSON blob--within a Django ecosystem it'll be easiest to simply use a Django model).
+
+We would need to create "callback" (login) and logout endpoints and [integrate Django with Auth0](https://auth0.com/docs/quickstart/webapp/django/01-login), but user creation, email verification, and password resets would be handled by Auth0. Initially, we would not need to integrate with an email sending service, but [using your own email service is necessary if you want to customize the email templates](https://auth0.com/docs/customize/email/email-templates).
+
+It's unclear how much Auth0 configuration needs to be done via click-ops; there is a [Terraform Auth0 provider](https://registry.terraform.io/providers/alexkappa/auth0/latest/docs) but I'm not sure how robust it is (though it does seem to be actively maintained).
+
+As a side note, `woonerf` (the utility library that Conveyal used when building the application) [includes functionality for integrating with Auth0](https://github.com/conveyal/woonerf#auth0). However, `woonerf` has not been actively maintained for 3-4 years, and Conveyal does not seem to be using it in new applications, so we should consider it deprecated and build Auth0 support on our own.
+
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Is there a usual way we do email for auth in Django? Are we going to want the emails to come from the [echodomain](https://echolocator.org/)? To get around spam filters that could be non-trivial.

Are we sure all potential (or current) users have an email address to use? Could we be asked to add phone # logins in the future?

Is passwordless an option? (or maybe just for non-counselors?)
1 Example: Click login with email, get sent a link, click it, you are logged in on that browser for say 30 days.
 * for some users their shared password might be the most sensitive info we store
 * Removes most reasons from considering to add social login in future
 * Will simplify migration - as no need to deal with (most) passwords or send an email blast to a lot of people
 * this actually could work on django or auth0 from my quick googling.

ddohler:
@BryanQuigley Thanks for reviewing! Answers as best I can give them below:

> Is there a usual way we do email for auth in Django? Are we going to want the emails to come from the [echodomain](https://echolocator.org/)? To get around spam filters that could be non-trivial.

We've used https://github.com/azavea/django-amazon-ses in the past and I _believe_ that it's still our go-to option, but we haven't spun up a Django project on our team in a while so I would probably double-check when we get to that point. I do think we'll want to send from the application domain, and I believe that using SES makes that relatively straightforward. My understanding is that AWS will automate some or most of the anti-spam configuration for us if we also manage the domain in Route53, which we already do.

> Are we sure all potential (or current) users have an email address to use? Could we be asked to add phone # logins in the future?

That is a great question. I just double-checked, and I'm certain that all current users have emails addresses because we've got the Cognito pool configured to disallow phone numbers:
Screen Shot 2022-02-21 at 11 49 38 AM

I'm not so certain that we won't be asked to add phone numbers in the future. If we were, we could certainly do it with Twilio or whatever SMS-sending service AWS provides, but this definitely is another example of the category of "things that we don't know whether we'll be asked to do, but that would be much easier to do with Auth0 than Django if they ever do come up".

> Is passwordless an option? (or maybe just for non-counselors?)

This is a really good idea (and yes, probably for non-admin users)! I've noticed a lot of less-technical people that I know effectively do this anyway because they subscribe to the "Why bother to remember my passwords when I can reset them with my email?" school of password management, so it may be that this is a friendlier pattern for apps intended for the general population anyway. And doing passwordless should simplify several parts of the user migration and implementation.

I think we'd want to run it by the client first, and I'd also like to get the opinion of someone from the Design team, but I'm not sure who's on the project now that Scott has left. @stephwall who should we talk to for design / UX questions on this project now? And can we run this by BHA assuming the designers are in favor?

ktohalloran:
I remember that part of the reason Civic Apps used Auth0 was for security reasons to pull identifying information apart from login credentials. It doesn't sound like that level of security is necessary here, but does a Django-only approach have any potential downsides from a security standpoint?

ddohler:
> does a Django-only approach have any potential downsides from a security standpoint?

@ktohalloran That's a very good question that's unfortunately pretty tough to answer concretely. In general terms, by outsourcing some functionality to Auth0, we inherit whatever strengths and weaknesses Auth0 has with respect to the functionality we're asking them to handle. For example, Auth0 presumably has more resources to devote to security than we do, but they are also a high-profile, centralized target and this app won't be.

One area that I can talk about concretely is that if we handle usernames and passwords ourselves, we'll have copies of all users' hashed passwords, which can be a risk because the hashed password can theoretically be used to retrieve the user's plaintext password, if a weakness is found in the hashing scheme. One common way that users' accounts are compromised is if they use the same password across multiple web sites and one of those sites is compromised and uses weak password hashing. Then an attacker can get the password out of the compromised site and try it on other sites. By having yet another copy of our users' (presumably reused) hashed passwords, we're somewhat increasing the risk to them (though in general, Django is pretty good about choosing secure defaults for password hashing). But this isn't exactly a difference because it's a risk with Auth0 as well. That's one reason why I like the passwordless idea -- it means that we wouldn't be storing passwords anywhere so we wouldn't have that information even if the site did get compromised.

doc/arch/adr-001-identity-and-authorization.md
```diff
@@ -0,0 +1,134 @@
...
+## Decision
+
+_What have we decided to do?_
+
+We have a (relatively) healthy budget and timeframe for completing this work, and part of the justification for undertaking the switch to Django was customizability and ease of administration. Given that, it probably makes the most sense to pursue a Django-only solution so that everything stays in one place and is as customizable as possible (even if that customization costs more in the long run).
+
+## Consequences
+
+- Customizations to the login pages will be relatively straightforward.
+- Customizations to the authentication process (e.g. adding social logins or MFA) will be relatively more complex.
```
maurizi: I'm sure there are other libraries available as well, but I've used `django-allauth` in the past for google / social media authentication and been pleasantly surprised at what the configuration entailed.

doc/arch/adr-001-identity-and-authorization.md
```diff
@@ -0,0 +1,134 @@
...
+5. Create an admin-configurable flag to prevent user login or signup temporarily. The primary risk during this deployment process will be data loss if a user changes their data between when we start the migrating and when we finish it, so we should prevent users from logging in while the migration is in process.
+6. Develop and test a method to tell all users to reset their password. This could be accomplished via a script and whatever email service we choose to use, or through BHA infrastructure if they have something like MailChimp.
+
+Once the above are all complete, the deployment plan should look something like this:
+1. Coordinate with BHA on a deployment window
+2. Queue up "reset your password" email blast
+3. At start of window, deploy to production with login/signup flag set to false.
+4. Run user migration scripts
+5. Switch login/signup flag to true
+6. Send "reset your password" email blast
```
maurizi, line 133: If we can customize the email per user with a script, we can take advantage of Django's signing support to generate links that will take people directly to a "change your password" page - or maybe even just log them in and take them to the homepage if we're going passwordless.

- Django docs (they're a bit low-level): https://docs.djangoproject.com/en/4.0/topics/signing/
- A blog post about how to do this for unsubscribe links (which is what I've used it for in the past) http://www.grokcode.com/819/one-click-unsubscribes-for-django-apps/

maurizi (approved):
We've done a lot of Django authentication setups in the past (I don't think we've ever done Auth0 on the team?), so I think this is definitely the less risky option to choose.

I agree with @BryanQuigley that a password-less login option would be nice, and it could make the transition much simpler - and looking around was able to find at least one library that supports it and looks well maintained.

rachelekm (commented):
Are we going to continue to use the `@conveyal/woonerf/fetch` library to call our Django API endpoints and handle auth state on the frontend? I do remember that `fetch` will automatically dispatch `INCREMENT_FETCH` and `DECREMENT_FETCH` actions on every call (annoying but certainly manageable), but do we anticipate any other potential side effects? Interestingly `woonerf` has a [Auth0 login component](https://github.com/conveyal/woonerf#auth0), but I'm not sure that's even a bonus given how the library has worked for us so far and even then I agree that Django authentication is less risky and has more benefit long term.

doc/arch/adr-001-identity-and-authorization.md
ddohler: I've used `django-social-auth` in the distant past and also had a good time, so yeah, we've probably got some decent libraries to choose from. I'll update the doc to make that clear.

doc/arch/adr-001-identity-and-authorization.md
ddohler: Cool! I've never used Django signing before, that looks really neat.

ddohler:
@rachelekm That's a really interesting point; I hadn't even thought to check whether `woonerf` might have something applicable, but I agree with you that doubling down on using more `woonerf` features is probably not the way we want to go, given that it no longer seems to be maintained. I'll update the doc to mention this.

I hadn't thought specifically about whether to use `woonerf` for the Django API interaction, but I think we should probably avoid adding new uses of it, given the struggles you had with `INCREMENT_FETCH` and the aforementioned lack of maintenance. I'll update to mention that as well.

ddohler:
Thanks all for the comments! I've updated the document to reflect some of the discussion here and I'm going to start opening issues for implementation. I'll leave this open until Wednesday is case there is further discussion before RRP.

Issue #376, opened 2021-11-09
Switch to a different bundler


ddohler:
Currently the project uses Mastarm, but that project is a custom Conveyal tool that is no longer maintained, and we've had to [fork it](https://github.com/azavea/mastarm) to fix bugs. Switching to a more mainstream tool such Webpack would improve maintainability of the project.

Part of this should include adjusting how environment variables are handled for the frontend and pulling them in from S3 via `scripts/bootstrap`. See [this discussion for details](https://github.com/azavea/echo-locator/pull/419#discussion_r827235296).

We should also move `taui` into the `src` directory.

aaronxsu:
Hey @ddohler , I made some notes after some high-level investigation, and would like to ask you to review if the proposed changes make sense before proceeding.

cc @rachelekm and @ktohalloran , please also take a look and feel free to join the discussion!

If there is any infra related changes, I will make notes of them along the way and let Bryan know after he comes back from vacation.

### Overview

I did some light investigation today trying to understand better the scope of the issue- what the current status is, what options we have to decouple and replace, and what related scripts for local dev and CI/CD will need updates. I laid out my understanding of these aspects in the following sections, and would like to invite folks to join this discussion, since this is an 8-point card and it potentially has some downstream (positive) impact for development and deployment.

### What is the current status?

In order to remove  [`mastarm`](https://github.com/azavea/mastarm) and replace it with more mainstream alternatives, we need to understand what the current status is. The frontend of this app uses `mastarm` for a handful of tasks which seem to have extended what I think a conventional bundler would do (to my understanding). `mastarm` is used for the following tasks, as specified in the task runner scripts in [`package.json`](https://github.com/azavea/echo-locator/blob/develop/taui/package.json#L6-L18).

- transpile and build the app as static assets
- transpile and run the app locally with a proxy config
- lint JS code
- lint CSS code
- handle deployment of the app to AWS

In addition to the above, `mastarm` also consumes some yaml files under `/configurations` for different types of tasks:

- `env.yml` to store environment variables for the app runtime
- `store.yml` to store initial values for some fields in the `redux` store
- `settings.yml`- this is something I am not entirely sure, but it looks like it is mainly for deployment purpose

### What to decouple and replace?

`mastarm` handles of lot of tasks through CLI commands and arguments, which are used as task runner scripts in the app frontend. Additionally, it also has some dependencies that are critical as `devDependencies` for this app frontend. We need to decouple them and replace with the following mainstream options:

[...]

PR #419, opened 2022-03-10
Bootstrap Django API

### LINKED ISSUE #414: Add Django to development environment
- Add a Django application, backed by Postgresql, to the project. We haven't set up a Django / React project from scratch in a while, so it would be a good ideal to consult with someone on CA and then copy from their best-practice project as much as possible.
- `scripts/setup`, `scripts/update`, and `scripts/server` should all work as expected (i.e. they should get the project working and run migrations as appropriate).
- Also add a `scripts/manage` to issue Django management commands.
- Choose a port for Django runserver and add it to the team ports document.
- You don't need to take any steps to implement CI or infrastructure.
- The frontend should be able to hit a basic API endpoint implemented in Django and confirm success. This doesn't have to do anything useful except confirm that the React frontend has successfully communicated with the Django API. I expect that we'll want the React app and the Django app to function from the same domain (rather than communicating from localhost:9966 for the React app to localhost:8000 for Django, for example) but consult with CA on this.

aaronxsu:
## Overview

This PR:
- adds `django` and `database` services for local dev as API and DB
- adds/updates `./scripts/bootstrap`, `./scripts/setup`, `./scripts/update`, `./scripts/manage`, `./scripts/server`, `./scripts/dbshell`
- adds a simple `/api/` API route and returns a message that shows the API works as expected
- adds a small test on the frontend to allow testing the communication with local dev API

### Notes

1. I made some notes as single comments in the files changed below. Feel free to discuss!
2. I did not add in Vagrant in this solution, and would like to propose to start from just using Docker. If we find it easier to add Vagrant as a layer in between in the future, I'd be happy to take that issue as well.

## Testing Instructions

 * Make sure your local docker app is running
 * At the root of the directory, run `./scripts/setup`, which should bootstrap the app, install deps, build containers, and run migrations
 * Run `./scripts/server`, which should spin up a local DB, frontend, and Django API server
 * After the Django API is up and running, make sure the following is successful
```
curl --location --request GET 'http://localhost:8081/api'
```
* Go to http://localhost:9966, open your network tabs in the dev tools, and filter to the `/api/` call. Make sure the request is also successful
* At the root of the directory, run `./scripts/manage migrate` and make sure it runs migrations, although there is no migration to run now, which is fine
~At the root of the directory, run `./scripts/dbshell` and make sure it takes you to a local `psql` shell and you could do normal SQL stuffs there~

Resolves #414


backend/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM quay.io/azavea/django:3.2-python3.9-slim
```
aaronxsu: Let's use `django 3.2` and `python 3.9` for this project API!

backend/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN mkdir -p /usr/local/backend
+WORKDIR /usr/local/backend
+
+COPY requirements.txt /usr/local/backend/
+RUN set -ex \
+    && buildDeps=" \
+    build-essential \
+    " \
+    && deps=" \
+    postgresql-client-13 \
```
aaronxsu: client for `postgres 13`

backend/echo/settings.py
```diff
@@ -0,0 +1,158 @@
...
+
+WSGI_APPLICATION = 'echo.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/3.2/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.contrib.gis.db.backends.postgis',
```
aaronxsu: Let's use the `postgres-postgis` engine

backend/api/views.py
```diff
@@ -0,0 +1,9 @@
+from django.shortcuts import render
+from rest_framework.response import Response
+from rest_framework import status
+from rest_framework.decorators import api_view
+
+# This will return a list of books
+@api_view(["GET"])
```
aaronxsu: This is just for testing purpose for now to make sure that the API is working as expected. We should remove this in the future.

scripts/bootstrap
```diff
@@ -0,0 +1,33 @@
...
+        "Usage: $(basename "$0")
+Pull down secrets from S3.
+"
+}
+
+function pull_env() {
+    pushd "${DIR}"
+
+    echo "Pulling .env from s3://${1}"
+    # aws s3 cp "s3://${1}/.env" ".env"
```
aaronxsu, line 24: I am assuming that we would pull some env vars from S3 here. Please let me know if this should be something else.

scripts/dbshell
```diff
@@ -0,0 +1,20 @@
...
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        docker-compose up -d database
+        docker-compose \
+            exec database bash -c "psql -U echolocator -d echolocator"
```
aaronxsu: The hope here is to let us get into a local db shell and see what is up in there if we'd like to check something quickly.

README.md
```diff
@@ -67,6 +67,21 @@ versions are not installed).
...
+| Script      | Use                                                        |
+| ----------- | ---------------------------------------------------------- |
+| `bootstrap` | Pull down secrets from S3                                  |
+| `dbshell`   | Open a `psql` shell connected to the local dev database    |
+| `infra`     | Execute Terraform subcommands with remote state management |
+| `server`    | Start the frontend and backend services                    |
+| `setup`     | Setup the project development environment                  |
+| `test`      | Run linters and tests                                      |
+| `update`    | Update project, assemble, run migrations                   |
+
```
aaronxsu, line 83: Please let me know if these suffice. I am happy to add more scripts or more descriptions of other scripts that were written before.

backend/echo/settings.py
```diff
@@ -0,0 +1,162 @@
...
+DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
+
+WATCHMAN_ERROR_CODE = 503
+WATCHMAN_CHECKS = (
+    'watchman.checks.databases',
+    'api.checks.gazetteercache'
+)
+
+# CORS settings
+CORS_ORIGIN_WHITELIST = [
```
aaronxsu: Though it looks like OAR did something fancy here: https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/django/oar/settings.py#L380-L399 -- If this whitelist won't suffice, I am happy to emulate what OAR does in the above linked code block.

backend/requirements.txt
```diff
@@ -0,0 +1,5 @@
+django-cors-headers==3.11.0
```
aaronxsu: I think these might be the bare minimum.

docker-compose.yml
```diff
@@ -10,3 +10,52 @@ services:
...
+    command:
+      - "-b :8081"
+      - "--reload"
+      - "--timeout=90"
+      - "--access-logfile=-"
+      - "--error-logfile=-"
+      - "--log-level=debug"
+      - "echo.wsgi"
+    ports:
+      - "8081:8081"
```
aaronxsu: I saw the requirement to "add to team ports document"- could you let me know where you find that doc?

docker-compose.yml
```diff
@@ -10,3 +10,52 @@ services:
...
+  django:
+    image: echolocator
+    environment:
+      - POSTGRES_HOST=database
+      - POSTGRES_PORT=5432
+      - POSTGRES_USER=echolocator
+      - POSTGRES_PASSWORD=echolocator
+      - POSTGRES_DB=echolocator
+      - DJANGO_ENV=Development
+      - DJANGO_SECRET_KEY=secret
```
aaronxsu, line 23: I think we would need an `.env` file at a point for these env vars.

README.md
```diff
@@ -67,6 +67,21 @@ versions are not installed).
...
 
 
+### STRTA
+
+This project uses [`scripts-to-rule-them-all`](https://github.com/azavea/architecture/blob/master/doc/arch/adr-0000-scripts-to-rule-them-all.md) to bootstrap, test, and maintain projects consistently across all teams. Below is a quick explanation for the specific usage of each script.
+
+| Script      | Use                                                        |
+| ----------- | ---------------------------------------------------------- |
+| `bootstrap` | Pull down secrets from S3                                  |
+| `dbshell`   | Open a `psql` shell connected to the local dev database    |
```
ddohler:
```suggestion
| `manage`   | Running Django manage.py commands    |
```

As noted below, we don't really need this as a separate command if we already have `./scripts/manage` available, so I think we can just document `manage` and call it good.

scripts/server
```diff
@@ -20,6 +19,6 @@ then
     then
         usage
     else
-        docker-compose up --build
```
ddohler, line 23: Nice catch! I tried to figure out why this was put here in the first place, but couldn't. Taking it out doesn't seem to hurt anything that I can find, and it certainly speeds up development because the frontend container doesn't have to get rebuilt every time you start development. If we do find that removing this causes problems, hopefully we can find a different solution than just reverting this change, because development is a lot more pleasant now that `./scripts/setup` doesn't take several minutes to get going.

scripts/update
```diff
@@ -0,0 +1,33 @@
...
+
+        # Bring up PostgreSQL and Django in a way that respects
+        # configured service health checks.
+        docker-compose up -d database django
+
+        # Apply any outstanding Django migrations
+        docker-compose \
+            run --rm --entrypoint python \
+            django \
+            manage.py migrate --no-input
```
ddohler: I think we need to add `docker-compose stop` at the end here so that this doesn't leave the containers running after the script finishes.

scripts/update
```diff
@@ -0,0 +1,33 @@
...
+    else
+        # Ensure container images are current
+        docker-compose build
+
+        # Bring up PostgreSQL and Django in a way that respects
+        # configured service health checks.
+        docker-compose up -d database django
+
+        # Apply any outstanding Django migrations
+        docker-compose \
```
ddohler: Since we have `./scripts/manage` already available, another way to handle this would be to only bring up the `database` service on line 25 and then use `./scripts/manage migrate` here.

scripts/update
```diff
@@ -0,0 +1,33 @@
...
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        # Ensure container images are current
+        docker-compose build
+
+        # Bring up PostgreSQL and Django in a way that respects
+        # configured service health checks.
+        docker-compose up -d database django
```
ddohler, line 31:
```suggestion
        docker-compose up -d database
```
(To match the comment below)

scripts/update
```diff
@@ -0,0 +1,33 @@
...
+
+        # Apply any outstanding Django migrations
+        docker-compose \
+            run --rm --entrypoint python \
+            django \
+            manage.py migrate --no-input
+    fi
```
ddohler, line 40: I think static files will need to be accounted for here as well; they're necessary for the Django admin to function correctly and for the Django Rest Framework browsable API. I think we'll need to add `./scripts/manage collectstatic --no-input` here and [adjust `settings.py`](https://docs.djangoproject.com/en/3.2/howto/static-files/).

scripts/bootstrap
ddohler: Hmm, this is an interesting point. The backend of the application doesn't need any environment variables yet, and the frontend environment variables are managed by this section in the README:
> To start developing, create a set of Taui environment variables for development:
>
> ```
> $ cp taui/configurations/default/env.yml.tmp taui/configurations/default/env.yml
> ```
>
> Make sure to edit `env.yml` to set the appropriate secrets for development.

So I think probably the "correct" thing to do here (in quotes because I'm not sure it's actually the best thing to do) would be to have this script perform those instructions, i.e., copy the `env.yml` file and fill in the appropriate values. However, that seems like it's kind of outside of the scope of this task because the `env.yml` file is only related to frontend concerns; it won't impact Django, and the Django app currently doesn't need a separate `.env` file for local development anyway.

Additionally, if we ever do need an `.env` file for local development, we'll then have two different types of environment files for the frontend and backend (`env.yml` and `.env`, respectively). That seems unnecessarily complicated, and potentially confusing or harmful, so I think we would want to prioritize eliminating the difference, which we could do by [migrating away from `mastarm`](https://github.com/azavea/echo-locator/issues/376).

So to summarize, I think we can probably skip adding this file for now (or leave it as a placeholder) because the most useful thing it could do at the moment (copying env.yml) isn't related to Django and might not last very long anyway. We should probably update #376 with a note about this discussion since this seems like a point in favor of doing that.

How does that sound?

docker-compose.yml
ddohler: Sent a link to the ports document out of band; this is just a reminder to update this file to match whatever port we choose.

docker-compose.yml
```diff
@@ -10,3 +10,52 @@ services:
...
+  database:
+    image: postgis/postgis:13-3.1
+    expose:
+      - "5432"
+    environment:
+      - POSTGRES_USER=echolocator
+      - POSTGRES_PASSWORD=echolocator
+      - POSTGRES_DB=echolocator
+    healthcheck:
+      test: [ "CMD", "pg_isready", "-U", "echolocator" ]
```
ddohler, line 56: I'm not sure that this is having the intended effect. When I bring up the project from scratch (which can be simulated by running `docker-compose stop && docker-compose down`), I consistently receive an error connecting to the database:

```
django.db.utils.OperationalError: connection to server at "database" (172.25.0.2), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
```

Everything works fine on the second try, however, so I'm not sure whether this is a race condition or some kind of different behavior that occurs when creating the container for the first time.

scripts/dbshell
ddohler: Django makes `manage.py dbshell` available, so we can already get into the database via `./scripts/manage dbshell`. Having a separate script does save some typing, but I'm not sure the extra overhead of maintaining another script is worth it to save a few characters. I'd probably just remove this file, but I'm not opposed to keeping it if you think it would be helpful.

backend/api/views.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.response import Response
+from rest_framework import status
+from rest_framework.decorators import api_view
+
+# This will return a list of books
```
ddohler: Looks like this is a leftover from somewhere else?

backend/api/views.py
ddohler: This would be great as a comment in the code as well, so that future visitors know that this is safe to remove.

backend/echo/settings.py
ddohler: I think the differences with OAR extend beyond just the different CORS configuration. The CORS configuration on OAR is actually not relevant for the main OAR site, because OAR serves the static frontend and the API from the same host and port (e.g. while visiting `https://openapparel.org/` one might see a request to `https://openapparel.org/api/contributor-lists/?contributors=2422`). This means that CORS doesn't apply at all for the application's frontend because the API requests aren't cross-origin in the first place. I believe the fancier setup in OAR is to allow use of the OAR API from within certain other websites that have some kind of relationship with OAR, which isn't really something we expect to do.

The current setup here, by contrast, is cross-origin because the API and frontend are hosted on different ports, and if we continue that architecture through to deployment, we'll end up with the API hosted on something like `api.echosearch.org/api/` while the rest of the app is hosted at `www.echosearch.org`, and then we'll have to add `*.echosearch.org` to this whitelist.

I see a few different ways to proceed from here.

If we want to get closer to how OAR does things, OAR uses `django-spa` and `whitenoise` to serve the static files from within the Django app, and they use Webpack's proxy feature to pass through requests to `/api/` and `/admin/` through to the Django server. Mastarm supposedly [has a proxy feature](https://github.com/azavea/mastarm#build), but it's not clear how well it works, so it may be that to use proxying in the build tool we'd need to [migrate away from Mastarm](https://github.com/azavea/echo-locator/issues/376).

Alternatively, we could stick with the status quo and commit to making the app cross-origin for the foreseeable future. This would avoid the work of upgrading Mastarm, and would allow us to leave the existing deployment scheme for static files largely untouched. If we switched to hosting the files in Django, we'd need to swap out the existing S3-based hosting with Django static hosting for everything. However, this would potentially make the app more confusing for administrative users because the Django admin would have to be accessed at something like `api.echosearch.org/admin/`; if we put everything on the same domain it could be `www.echosearch.org/admin/` which seems friendlier.

Overall, Mastarm has been a continuing source of confusion and frustration (including one other thread in this PR), so I'm leaning toward the same-origin solution so that we have another excuse to get rid of it, but I could also see sticking with a cross-origin solution to avoid the need to rework the static deployment (even in an imaginary world where we were already using a more standard bundler).

So I think my suggested path forward would be to check out Mastarm's proxy feature and try to add `django-spa` if it seems to work. If that all works out, then we can eliminate `CORS` support entirely. If we run into roadblocks then I think we can leave this as-is and prioritize removing Mastarm. How does that sound to you?

ddohler (changes requested):
Thanks for putting this together! Overall, this looks really good. I found some smaller things that I think need to be addressed, and some larger architecture discussion items that might potentially lead to more work, or to other issues being created. Let me know what you think!

scripts/dbshell
aaronxsu: Agreed- makes sense to me. I will remove this script.

scripts/server
aaronxsu: Thanks. This sounds good to me!

README.md
aaronxsu: Sounds good. I will remove it 👍🏼

backend/api/views.py
aaronxsu: Oops. Yes, I will remove it.

backend/api/views.py
aaronxsu: Makes sense. Lemme add a todo note in here so that we will remove it later.

scripts/bootstrap
aaronxsu: Ah, this is a very good point. I agree- it does not make much sense to me either to copy the `env.yml`from the template file and populate the values, since:1. we would potentially move away from `mastarm`; 2. the values would need some human op to grab from last pass and fill in manually. And yea, we indeed do not need any env vars for the Django dev yet. I am with you on leaving this `bootstrap` script as is and treat the current code comment as a placeholder!

scripts/update
aaronxsu: Smart choice and less verbose. I like it! I will update this.

scripts/update
aaronxsu: Ah, good point. I like this, since this will be cleaner. Will make this change!

scripts/update
aaronxsu: Good catch. I will add this one as well.

docker-compose.yml
aaronxsu: Hmmm, that looks odd indeed. I tried running `docker-compose stop && docker-compose down` and then `./scripts/setup` and then `./scripts/server`. However, I did not find the above error message in the logs. Did I miss anything step here? 🤔

docker-compose.yml
aaronxsu: Sounds good. I ended up using port `8085` for the Django server. Here is the PR for updating the docs: https://github.com/azavea/geospatial-apps/pull/261

backend/echo/settings.py
aaronxsu: Yes, this plan sounds very good to me. I think the reasoning here makes a lot of sense. 

I do think we should start by testing if adding `django-spa` and `whitenoise` to serve static files and also update the `proxy` flag for `mastarm build` works. If this won't work as expected, I am with you on the fact that switching away to another bundler is a wise choice here since `mastarm` is not maintained and we now have a chance to make the switch since we need to make the local frontend dev work with a Django API. I do not think that the currently implemented `CORS` solution is a way to go considering the downstream build and infra overhead.

I did take a stab at adding `django-spa` and `whitenoise`, plus modifying the `--proxy` flag in [this commit](https://github.com/azavea/echo-locator/pull/419/commits/432b9113b09b55deb83ccd8d60d0759d4a228ea4), but I think I was missing something there and it did not work for me locally. I think it might be that it expects to find a `index.html` in `backend/static`, but `mastarm build` does not seem to produce such a thing while only produces something like `taui/assets/index.js`. Plus, I think we will need a way to mount the directory of bundled frontend result to where the django docker container can find in the `backend/static`, but I could be wrong.

Thus, given the above mentioned confusion, I am happy to pair with you when you get a chance to have this figured out together!

aaronxsu: Update: discussed about the above on Slack, and it [this commit](https://github.com/azavea/echo-locator/pull/419/commits/c09e13776a1ea54109463fc12f4d7d19d14e175d) fixed the issue for local development 🎉

README.md
```diff
@@ -67,6 +67,20 @@ versions are not installed).
...
 
 
+### STRTA
+
+This project uses [`scripts-to-rule-them-all`](https://github.com/azavea/architecture/blob/master/doc/arch/adr-0000-scripts-to-rule-them-all.md) to bootstrap, test, and maintain projects consistently across all teams. Below is a quick explanation for the specific usage of each script.
+
+| Script      | Use                                                        |
+| ----------- | ---------------------------------------------------------- |
+| `bootstrap` | Pull down secrets from S3                                  |
+| `infra`     | Execute Terraform subcommands with remote state management |
```
ddohler, line 77: We should also include docs for `manage` here.

backend/api/views.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.response import Response
+from rest_framework import status
+from rest_framework.decorators import api_view
+
+# TODO: remove this later since this is to test API
```
ddohler: Our normal team practice is that all TODOs should have a link to a GitHub issue so that they aren't forgotten. We didn't have an issue for this, so I went and [made one](https://github.com/azavea/echo-locator/issues/422), but the more I'm thinking about it, the more I think the best thing to do is just to remove this before merging, because it's more of a debugging check to prove that everything works than something that needs to live beyond this PR.

taui/src/index.js
```diff
@@ -163,6 +163,18 @@ class InitializationWrapper extends React.Component {
     }
   }
 
+  // TODO: remove this later
```
ddohler: Same comment here as above, this is probably best removed before merging, that way we can be certain not to lose track of it later.

scripts/bootstrap
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
```
ddohler, line 2: This could use a TODO comment with a link to #376 since I think that is where this would be most likely to get fleshed out.

backend/Dockerfile
ddohler: One thing I forgot to mention in the previous review (sorry!) that I think would be good to do is make the directory structure match our standard structure a bit more closely. Normally we nest application code under `src/`, but that didn't matter so much when there was only a frontend to this app. However, now that we've got both a backend and a frontend, I think it would be best to try to work our way back to our standard structure. Moving the frontend might break some things, so if we can just start by putting the Django API inside of `src/` now, then we can move `taui/` into the `src/` folder as part of #376 . How does that sound?

docker-compose.yml
ddohler: Okay, I think I see what's going on: the `healthcheck` option [is valid for v2.1 Compose files, but was _removed_ for v3 Compose files](https://github.com/peter-evans/docker-compose-healthcheck/issues/3) (which we're using). So I don't think that the Django container is waiting for the database container at all. I'm not sure why it's working for you and not for me, but I suspect it's just a difference between our two systems such that the database container comes up a bit slower on my system and exposes the race condition.

It looks like the easiest solution may be to simply switch the compose file syntax version to `2.1` instead of `3`. That seemed to solve the issue for me.

scripts/update
```diff
@@ -0,0 +1,35 @@
...
+        # configured service health checks.
+        docker-compose up -d database
+
+        # Apply any outstanding Django migrations
+        ./scripts/manage migrate
+
+        # Collect Django static files
+        ./scripts/manage collectstatic --no-input
+        
+        docker-compose stop
```
ddohler, line 39: The database connection errors that I was getting exposed that this doesn't stop the database container when there's an error. I was able to get the containers to clean up correctly by adding this to the script:

```
function cleanup() {
  docker-compose stop
}

trap cleanup ERR
```

taui/src/index.js
```diff
@@ -163,6 +163,18 @@ class InitializationWrapper extends React.Component {
...
+  // This is to test and prove that the local Django API works
+  componentDidMount () {
+    fetch('http://localhost:9966/api/').then(function (response) {
```
ddohler: No need to change this if you're going to remove this whole thing, but this can be just `/api/` since this app isn't operating cross-origin any longer:
```suggestion
    fetch('/api/').then(function (response) {
```

taui/package.json
```diff
@@ -14,7 +14,7 @@
     "prestart": "yarn",
     "pretest": "yarn",
     "semantic-release": "semantic-release",
-    "start": "mastarm build --serve --watch",
+    "start": "mastarm build --serve --watch --proxy http://django:8085",
```
ddohler, line 17: One thing I noticed is that if you try to access `/api/` through the proxy (i.e. `localhost:9966/api/`) the static files don't show up and return 404s. However, I don't think we need to fix that because we're planning to shift away from Mastarm anyway, and everything works fine if you access the Django container directly via `localhost:8085/api/`.

ddohler (approved):
This is working well! I found a few more things that I think should be changed but it looks like they've all got pretty easy resolutions, so this is good to go after cleaning those up!

backend/api/views.py
aaronxsu: Thanks for creating the issue. It makes sense to me to remove this API endpoint at this time.

taui/src/index.js
aaronxsu: Sounds good!

backend/Dockerfile
aaronxsu: That sounds like gold to me! I am happy to follow our usual practice here.

PR #420, opened 2022-03-11
Remove bedrooms filter for BHA listings

### LINKED ISSUE #418: Don't filter BHA listings by number of bedrooms
There's an issue with the BHA listing endpoint where passing the number of bedrooms as a query parameter doesn't work correctly. We don't have a way to resolve the root cause right now, so as a workaround we should omit the bedroom parameter (or send it as null, whichever one is easiest and works). We should continue to filter by max rent, assuming that works correctly.

We should continue to filter Realtor.com listings by the number of bedrooms and max rent.

rachelekm:
## Overview

Currently, we fetch BHA listings by querying an API endpoint with filter parameters. However, the number of bedrooms parameter is not working and causing valid listings to be filtered out of the response. As a temporary fix, this removes the bedrooms parameter and now the listings received are only filtered by zip code and max rent.

### Notes

Setting the `rooms` query parameter to `null` is not a valid option and will return an API response of `null`. Setting the query parameter to 0 successfully returns all listings filtered by the remaining parameters, however I chose to move forward with completely removing the `rooms` parameter since we've had difficulty with this API's query requirements before and this seemed like the least risky option. If the `rooms` parameter becomes mandatory, however, using 0 is an option.

## Testing Instructions

 * Run `./scripts/server`
 * Log in and select a neighborhood (usually the 'Boston - Dorchester – 02121' neighborhood has multiple BHA listings)
 * Confirm BHA listings appear and are filtered by max rent and zip code

Resolves #418

### DIFF
--- a/taui/src/actions/listings.js
+++ b/taui/src/actions/listings.js
@@ -51,9 +51,10 @@ export const setBHAListings = (payload: Listing | ListingQuery) => (dispatch: Di
     dispatch({type: BHA_ACTION_TYPE, payload: payload})
   } else {
     const zipcode = payload.query.zipcode
-    const rooms = payload.query.rooms
     const budget = payload.query.budget
-    const urlWithQuery = `${BHA_BASE_URL}zipcode=${zipcode}&rooms=${rooms}&budget=${budget}`
+    // The BHA API is not currently filtering by rooms correctly
+    // As a temporary fix, we took out the bedrooms query parameter so no listings are accidentally hidden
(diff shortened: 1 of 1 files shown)


ddohler (approved):
This should be re-targeted at `release/2022-03-02` but otherwise everything looks great! I bumped the number of bedrooms in my profile up to 3 and confirmed that I got apartments with fewer bedrooms in Dorchester and that `rooms` was missing from the query string.

rachelekm:
Whoops, thanks for the catch! Updated the base branch to `release/2022-03-02`.

Issue #415, opened 2022-02-28
Add Django to CI and add staging infrastructure


ddohler:
- `scripts/cibuild` and `scripts/cipublish` should work and should publish Django infrastructure to staging.
- `scripts/test` should run and formatting / linting issues should fail the build (including Django-related issues; I believe this is already true for the exsting React code).
- CI building should be triggered on PRs and staging deployments should be triggered on merges to `develop`, as things currently are.
- We may want to switch CI providers; we typically use GitHub Actions for more recent apps, and so it may be easier to just switch to a more recent example of a Django/React app deployed with GH Actions rather than attempting to build the same functionality in Travis.
- We may need to change how the frontend and static assets (like neighborhood photos) are being deployed.

PR #425, opened 2022-03-29
terraform upgrade to 1.0.0 /aws 3.75.1

### LINKED ISSUE #423: Update terraform to modern version
We are currently on terraform 0.11.11 which doesn't seem to work (0.11.15 does), but we should upgrade to terraform 1.0 before doing all this infra work. 

This does block using our ecr repo (although I might just use upstream as it seems to do it all nowadays.

BryanQuigley:
## Overview

Pushed it as high as I could without getting rid of the s3 site for now (blocks aws 4+).

Broke into 3 commits to match the bigger changes in the process (and make it easier to do in production too - or clear out production entirely).

Resolves #423


BryanQuigley:
I'm going to build at least a few more elements of #415 on top of this.

PR #434, opened 2022-03-31
Add in missing copy and tooltip on edit profile page

### LINKED ISSUE #433: Add missing copy & tooltip to profile page
We're missing the "Why are you asking this?" text and associated tooltip that's on the demo site profile page.

Note: The eventual PR for this should be targeted at the existing release branch rather than `develop`.

ktohalloran:
## Overview

The edit profile page was missing context on why users should put in addresses they visit frequently. This PR adds that copy, tooltip, and relevant translations.

### Demo

Screen Shot 2022-03-31 at 10 29 58 AM

## Testing Instructions

 * Start the server and navigate to **http://localhost:9966/profile**
 * Verify that the copy appears under the "What places do you visit often (besides your home)?" title and that the tooltip appears when you hover over it
 * Change the language and verify that the copy and tooltip both translate as they should

Resolves #433

### DIFF
--- a/taui/src/components/edit-profile.js
+++ b/taui/src/components/edit-profile.js
@@ -561,6 +561,23 @@ class EditProfile extends PureComponent<Props> {
     return (
       <div className='account-profile__destinations'>
         <h3 className='account-profile__label'>{t('Profile.Destinations')}</h3>
+        <span>
+          <div className='account-profile__button account-profile__button--tertiary account-profile__button--iconLeft'
+            data-tip={t('Tooltips.ProfileAddresses')}
+            data-for={`tooltip-profile-addresses`}>
+            {t('Profile.WhyAddresses')}
+          </div>
+          <ReactTooltip
+            clickable
+            html
+            effect='solid'
+            place='top'
+            isCapture
+            delayHide={TOOLTIP_HIDE_DELAY_MS}
+            className='map-sidebar__tooltip'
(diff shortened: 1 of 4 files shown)


taui/src/components/edit-profile.js
```diff
@@ -561,6 +561,23 @@ class EditProfile extends PureComponent<Props> {
...
+          <div className='account-profile__button account-profile__button--tertiary account-profile__button--iconLeft'
+            data-tip={t('Tooltips.ProfileAddresses')}
+            data-for={`tooltip-profile-addresses`}>
+            {t('Profile.WhyAddresses')}
+          </div>
+          <ReactTooltip
+            clickable
+            html
+            effect='solid'
+            place='top'
```
rachelekm, line 574:
```suggestion
            place='top'
            offset='{"top": -10}'
```
There's a larger gap between text and tooltip box here than the other icons, so this brings the tooltip down a little to look more standard. Not sure if there's a more elegant solution though?

rachelekm (approved):
Looks great! One comment to bring the tooltip a little closer to the text, like below, but definitely not necessary:
Screen Shot 2022-03-31 at 3 07 39 PM

ktohalloran:
Thanks for your review! I agree the tooltip looked awkwardly far away so thanks for fixing that. Merging now!

PR #435, opened 2022-03-31
Add Django admin

### LINKED ISSUE #426: Set up Django admin site
Just use [the out-of-the-box functionality](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/) for now: users with the `staff` or `superuser` attribute set can access the Django admin site by logging in at `/admin/`. Add instructions to the README for creating a user with admin access.

rachelekm:
## Overview

Creates an `/admin` site accessible to users with `staff` or `superuser` attributes. Also updates the `README.md` to walk users through setting up a superuser account and logging in.

### Notes
I’ve noticed that the `/admin` site navigation is really slow. The request for `admin/auth/users` for example took ~1 min and navigation back to `/admin` took ~1 min as well. It doesn’t appear consistent and I set up the `django-debug-toolbar` locally to explore further but didn’t find anything particularly insightful (how/which template files and sql queries made seemed normal), so I’m leaving it for now unless it becomes a problem down the road.
Browser timing info from toolbar:
Screen Shot 2022-03-31 at 10 58 12 AM

## Testing Instructions

 * Follow the instructions in the `README.md` to create a user account
 * Run ` ./scripts/server`
 * Navigate to http://localhost:8085/admin and login with new credentials
 * Confirm login and `Groups` and `Users` models exists 

Resolves #426

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,27 @@ This project uses [`scripts-to-rule-them-all`](https://github.com/azavea/archite
 | `test`      | Run linters and tests                                      |
 | `update`    | Update project, assemble, run migrations                   |
 
+#### Creating a staff user account
+
+From the project directory, run:
+```
+./scripts/manage createsuperuser
+```
+Fill out the prompts for a username, email, and password.
 
 ### Logging In
 
-Once it is running, log in using staging credentials. From there, you can make a Client ID by
+Run: 
+```
+./scripts/server
+```
+
+#### Site
+Navigate to http://localhost:9966 and log in using staging credentials. From there, you can make a Client ID by
 entering and searching for a random 6-8 digit number, then making a new profile.
+
+#### Admin
+Navigate to http://localhost:8085/admin and login with your credentials created through the `createsuperuser` command above.
 
 ## Data
 
--- a/src/backend/echo/urls.py
+++ b/src/backend/echo/urls.py
@@ -19,4 +19,6 @@ from django.conf.urls.static import static
 
 from echo import settings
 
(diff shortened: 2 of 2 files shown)


ktohalloran (commented):
Looks great! I tried to look into why it was slow but didn't come to any helpful conclusions, as it was similarly inconsistent for me. The only thing I can add is that some of the times it got stuck, I'm saw some `[CRITICAL] WORKER TIMEOUT` messages in my terminal; other times, I saw a lot of database commands doing things like setting my time zone. When the page finally loaded, I would see a lot of "GET" requests for static files in my terminal. I tried to do some reading about how static files are supposed to work, because this seemed strange to me, but again came to no conclusions. Speed aside, though, the admin site does work as I expected so good job! 👍

ddohler (approved):
Looks great! I didn't experience any slowness with the admin at all; everything was very snappy. Based on that, I suspect that this may be a Docker-on-Mac-specific issue. If you haven't already, it would be worth checking to see that you're running the latest version of Docker -- I know that they've been working hard to improve Mac performance. If that doesn't turn up anything, can you open an issue so that we can dig in further if necessary?

rachelekm:
Thanks, all! @ktohalloran, thanks for investigating. I noticed `[CRITICAL] WORKER TIMEOUT` this time around as well so I think that'll be a helpful clue. @ddohler, glad to hear it's quick on your end! I just ran updates through Docker Desktop and am now on Docker Engine v20.10.13 and Docker Compose v1.29.2. Unfortunately after updates and rebuilding images I'm still bumping into performance issues so I'll go ahead and open an issue.

PR #437, opened 2022-04-06
Add email console backend for development

### LINKED ISSUE #416: Set up email support in development
The Django app should be able to send emails to users.
- In development, email text should be printed to the server console

maurizi:
## Overview

Adds support for sending emails to the console logs in development, and enables Reset password for the Django admin (for testing purposes, but I thought we might want to leave it enabled, given that we intend for BHA users to regularly use the Django admin).

### Demo

Screenshot 2022-04-06 141618
Screenshot 2022-04-06 141628

## Testing Instructions

 * `scripts/server`
 * Go to the Django admin site and use the new reset password link
 * You should see an email printed to the console

Resolves #416

### DIFF
--- a/src/backend/echo/settings.py
+++ b/src/backend/echo/settings.py
@@ -43,6 +43,8 @@ if ENVIRONMENT == 'Development':
     ALLOWED_HOSTS.append('localhost')
     ALLOWED_HOSTS.append('django')
 
+    EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+
 # Application definition
 
 INSTALLED_APPS = [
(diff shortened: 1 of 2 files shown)


rachelekm (approved):
Email works great 👍 The password reset starts at `admin/login` and eventually provides a "log in" link to `accounts/login`, which we don't actually have a template for so I'm getting an error page. We could either change the template link so it redirects to `admin/login` or create a totally new template for the eventual login page, but either way that seems like a separate task so maybe better to leave off?

ddohler (approved):
I haven't tested this but the code looks great, so in the interest of time I'm going to call it good based on @rachelekm 's testing.

Issue #428, opened 2022-03-29
Set up Housing Seeker profile fields


ddohler:
Housing seekers need to be able to store the following fields on their profile:

- Username (will always be an email address, required)
- Full name (string, optional)
- Do you have a voucher? (boolean, required, default False)
  - Voucher number (string, optional, only used if they say they've got a voucher)
  - Number of bedrooms on voucher (numeric, optional, only used if they say they've got a voucher)
- Rent budget (numeric, optional, only used if they _don't_ have a voucher)
- Desired number of bedrooms (numeric, optional, only used if they _don't_ have a voucher)
- Travel mode (required, choose between 3 options: 1) car 2)bus/train 3) bus/train + express bus or commuter rail)
- Destinations. Each profile must have at least 1 Destination. Currently we limit the number of Destinations per profile to 3 on the client-side; I don't think that limit needs to be enforced strictly, so we can leave it out server-side if it's easier. A Destination has the following fields:
  - Address (string, required)
  - Purpose (choice, required, choices are: 1) Work 2) Day care 3) Family 4) Friends 5) Worship 6) Doctor 7) Other
  - Primary (boolean, default false, but exactly one of the Destinations associated with a profile must be marked Primary)
- Priorities. All priorities are choice fields with the following possible choices: 1) Not important 2) Somewhat important 3) Important 4) Very important. 
  - Commute priority (choice, default Not important) 
  - School quality priority (choice, default Not important)
  - Public safety priority (choice, default Not important)

These fields should be editable from the Admin

PR #438, opened 2022-04-06
Add Django to CI and add staging infrastructure

BryanQuigley:
Environment name stgdjango
Externally stg.echosearch.org

Using aws 4.8 and terraform 1.0
Move from azavea cert, ecr and rds
New cibuild/cipublish that works with Django images
Latest images will be used by default when ever plan/apply is run per force_new_deployment
Dropped travisCI
Add Github CI

Items to look at in future:
deployment/terraform-django/app.tf
 * healthcheck is healthy when 400 response (change to 200 eventually)
 * hardcoded to staging as our environment name is stg.
Dropped automatic cert validation terraform
	due to inability to provision staging.es worked around with name change
	staging -> stg.  staging failed with aws console too

Resolves #415

## Testing Instructions

 * Look out CI output to confirm it's doing the correct thing.

PR #439, opened 2022-04-07
Create Django User groups

### LINKED ISSUE #427: Create user roles and permissions
There are three roles that we're concerned about:
1. Housing Seeker: Have no ability to access the admin site (can only interact with the site via the frontend)
2. Counselors: Can log in to the admin site and only view / modify Client users and their profiles
3. Administrators: Can log in to the admin site and have full access.

See also #428 for a follow-on task, in case that influences the implementation here.

rachelekm:
## Overview

Creates three different `User` groups with variable levels of permissions: Superuser (not actually an explicit group, just `superuser` access), Counselor (admin access, but only able to view/modify `Users` in the `HouseSeeker` group and their respective `UserProfiles`), HouseSeekers (do not have admin access at all). The groups and their permissions are created through a Django management command that's run during `scripts/setup`. Right now `UserProfiles` are empty to be populated in #428, but it seemed easier to setup permissions by creating a basic model to get started so I left it in.

### Notes

- I found programmatically creating groups with assigned permissions through a management command the easiest way for permissions to be setup on app start. It's assumed that `HouseSeeker` should be able to add/delete/edit/view their own `UserProfile`, but since it won't be through the Admin site I left those permissions off for now. If we wanted to add any other permissions (like for the API models) the `MODELS` list in the management command would need to be updated.
- I created a receiver so that when a `User` is added to the `Counselor` group it will automatically update `is_staff` to `True`, and set to `False` when removed from `Counselors`. I'm hoping this makes roles/permissions simpler, but I'm still on the fence if this is worthwhile vs just making sure on `User` creation we assign to group and also set `is_staff` to True. It seemed simpler, so I left it in.
- When a `Counselor` logs in and views a `HouseSeeker` User, all fields will be visible. I originally started taking off the Permissions section since that didn't seem relevant to `Counselor`/something they shouldn't be able to edit, but I began to go down the rabbit hole on ways the Counselor admin should look different than the Superuser and felt like that could end up being it's own task so I put it to the side for now.

## Testing Instructions

 * Run `scripts/setup` so that the new `create_groups` command runs
 * Create a superuser if you haven't already: run `scripts/manage createsuperuser`
 * `scripts/server`
 * Navigate to `/admin` and log in
 * Click on "Groups" and confirm you see `Counselors` and `HouseSeekers`
 * Click on "Users" and create a new `Counselor` user and `HouseSeeker` user:
     * Add a new user and then in profile select "Counselor" in the groups section and save
     * Add a new user and then in profile select "HouseSeeker" in the groups section and save
     * Alternatively can create these users and assign to groups in the Django shell
* Click on the `Counselor` user you just created and confirm the `is_staff` checkbox has been automatically checked in the profile
* Logout and login as the `Counselor` user, confirm:
     * Can login to `/admin/` and see "Users" as well as "User profiles"
     * Only users visible are in `HouseSeeker` group
     * Can not add/delete a `User`
     * Can not add/delete a `UserProfile`
* Logout and try to login to `/admin/ `as a `HouseSeeker` and it should fail
* Run `./scripts/manage test` and confirm all pass

Resolves #427


ddohler:
Reassigning to @maurizi for review while I'm out.

scripts/test
```diff
@@ -20,6 +20,7 @@ then
     then
         usage
     else
+        echo "Running Taui tests..."
         docker-compose run --rm taui test
     fi
```
rachelekm, line 25: I originally added `./scripts/manage test` here but it's failing without collecting static files beforehand. We're [transitioning away from Travis CI](https://github.com/azavea/echo-locator/pull/438/files) so I'm taking it off for now until that's complete and then adding any necessary pre-test commands to the future `scripts/cibuild` file.

ktohalloran (commented):
Looks great! One note: I found the naming of `UserProfile` to be confusing, especially before we currently have anything in that table prior to the completion of [issue 428](https://github.com/azavea/echo-locator/issues/428).

I also wanted to ask a clarifying question of you @rachelekm and of @stephwall: I know the current model is that only counselors can create accounts but wanted to ask for a bit more clarity on the model we're moving towards. I know we want house seekers to be able to create their own profiles, too, but what is the ideal for the counselor's role in that--does it make sense for them to be able to add and delete as well? How interested would we be in giving counselors that ability and how hard would it be to do so?

maurizi:
> how hard would it be to do so?

I can't speak to the first half of the question, but I think we should be able to enable counselors creating housing seeker accounts in the Django admin fairly easily.

src/backend/users/management/commands/create_groups.py
```diff
@@ -0,0 +1,27 @@
...
+        for model in MODELS:
+            for permission in PERMISSIONS:
+                try:
+                    new_perm = Permission.objects.get(name=f'Can {permission} {model}')
+                except Permission.DoesNotExist:
+                    logging.warning(f'Permission not found with name {new_perm}')
+
+                counselor_group.permissions.add(new_perm)
+
+        self.stdout.write(self.style.SUCCESS('Successfully created Counselor and HouseSeeker groups and permissions'))
```
maurizi: This is a totally fine & tidy way to do things for our simple permissions setup.

I'm curious however if you considered using Django fixtures for this: https://docs.djangoproject.com/en/4.0/howto/initial-data/

src/backend/users/tests.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertNotIn("Add", content, "Got unexpected content: %s" % content)
+
+    def test_houseseeker_group_permissions(self):
+        response = self.houseseeker.get("/admin/")
+        # Expect redirect back to admin on failed login
+        self.assertEqual(
+            response.url,
+            "/admin/login/?next=/admin/",
+            f'Got unexpected response: {response.status_code}. {response.content}',
+        )
```
maurizi, line 77: Always great to see some tests 🎉

rachelekm:
> I know we want house seekers to be able to create their own profiles, too, but what is the ideal for the counselor's role in that--does it make sense for them to be able to add and delete as well? How interested would we be in giving counselors that ability and how hard would it be to do so?

From what I understand we don't want counselors to be able to add/delete house seekers. But this would be an easy fix, I could just add 'add' and 'delete' to the `PERMISSIONS` list in `create_groups`. I guess it's not very intuitive though, so if we anticipate changing the permissions of Counselors often maybe it would be worth it to just make a new User model with `PermissionsMixin` instead?

maurizi (approved):
Looks good :+1:

src/backend/users/management/commands/create_groups.py
rachelekm: Oh nice, this is definitely what I was looking for but couldn't find in my docs searching. I'd opt to switch over to fixtures so it's more standardized and will be much easier for future development vs a custom command.

ktohalloran (approved):

rachelekm:
@ktohalloran and @maurizi I just switched to setting permission through fixtures vs with a custom `create_groups` management command. Will you check to make sure this new commit works on your end before I merge?

src/backend/users/tests.py
```diff
@@ -7,7 +7,7 @@ class AdminSiteTest(TestCase):
     @classmethod
     def setUpClass(cls):
         cls.counselor, cls.houseseeker = Client(), Client()
-        call_command("create_groups")
+        call_command("loaddata", "users/fixtures/group_permissions")
```
maurizi: This can be even simpler now that you're using fixtures: https://docs.djangoproject.com/en/3.2/topics/testing/tools/#fixture-loading

rachelekm: Great, just fixed!

maurizi (approved):

PR #440, opened 2022-04-07
Replace Amplify authentication with anonymous-only login

### LINKED ISSUE #431: Re-enable anonymous login but without Amplify components
It's possible to use the site without logging in, but the feature is currently disabled by a feature flag in staging and production. When in anonymous mode, profile changes are saved to browser local storage rather than via API calls. Turning this functionality on simply requires setting `ALLOW_ANONYMOUS` to `true` in [env.yml](https://github.com/azavea/echo-locator/blob/develop/taui/configurations/default/env.yml.tmp).

As a first step toward shifting to Django for authentication, we should strip all Amplify-derived code from the frontend application, and replace the existing authentication page with a "login" page that simply repeats the existing anonymous login text:
> Don't have an account? [Continue without an account]()

The [auth diagrams](https://drive.google.com/drive/folders/1oA2b_KUvItY7Ecn9sWFQ6N1jFd7HgW_T?usp=sharing) will likely be helpful in figuring out where to cut.

It should look something like this:
Screen Shot 2022-03-29 at 16 00 34

ktohalloran:
## Overview

This PR completely removes Amplify and all of its related frontend authentication components and replaces the login flow with anonymous log in by replacing the `Authentication` HOC with a smaller `anonymousAuthenticator` HOC.

### Demo

login_AdobeCreativeCloudExpress

You are now only able to log in anonymously; doing so will take you directly to fill out your profile information.

### Notes

I had a tough time deciding the extend of what to cut out versus leave in for the sake of building from for Django authentication. Ultimately, along with removing anything that was explicitly related to Amplify's authentication process, I chose to remove places where currently useless checks for anonymity and/or authentication level impacted code readability prior to the addition of Django authentication, but left them in where they could be helpful for future authentication but not distracting. An example of this is removing the checks and related code for `isAnonymous` and `isCounselor` in `edit-profile.js`, as an anonymous user should never be presented with the option to re-send their invitation email, but leaving `clientAccountConfirmed`, `clientEmail`, and `clientInviteSent` as properties on the `AccountProfile` type. Following this rule, I have removed all checks for `isAnonymous`, `isCounselor`, and account validation, but have left `EMAIL_REGEX` in `constants.js` and email-related strings in the translation files.

One strange behavior I found following these changes was if you clicked "Continue without an account" and then clicked the browser's back button before completing the form, it would take you to a blank "search" page, rather than returning to the login page, unless you cleared your browser's local storage. This was compared to `develop`, where clicking back after selecting anonymous login takes you back to wherever you navigated to the login page from. I fixed this by replacing the redirect ternaries in `application.js` to redirect to '/profile' instead of '/search'. This fix is a little odd in the case of the 3rd path to '/profile', but it seemed a decent placeholder until we restore authentication.

## Testing Instructions

 * Run `./scripts/setup`
 * Run `./scripts/server` and navigate to **http://localhost:9966/**
 * Select "Continue without an account". It should take you to the profile page. 
 * Click your browser's back button and verify that it doesn't take you to a blank search page. ~~Then fill out the form and click save.~~
 * Click Cancel and verify that it takes you back to the login page. Select "Continue without an account" again and fill out the form and click save.
 * Verify that you can favorite neighborhoods, make changes to your profile, and that your profile is saved on browser refresh.

EDIT: Added testing instructions for re-route to login on cancel during profile edit.

Resolves #431


taui/src/components/anonymous-authenticator.js
```diff
@@ -0,0 +1,72 @@
...
+import CustomSignIn from './custom-sign-in'
+import CustomHeaderBar from './custom-header-bar'
+
+export default function anonymousAuthenticator (Comp) {
+  return class extends Component {
+    constructor (props) {
+      super(props)
+
+      this.handleAnonymousLogin = this.handleAnonymousLogin.bind(this)
+      this.handleProfileChange = this.handleProfileChange.bind(this)
```
BryanQuigley: I somewhat expected handleProfileChange to do the LoadProfile().  Any reason it doesn't?

BryanQuigley (approved):
The only small improvement I'd suggest is to just redirect back to the home page if someone presses Cancel and they haven't set up a profile yet.

taui/src/components/anonymous-authenticator.js
ktohalloran: I use `loadProfile()` as passed as a prop from `Application` on line 20--did you expect to see it used in a different way or as a method of `anonymousAuthenticator` rather than passed as a prop?

ktohalloran:
Thanks for reviewing @BryanQuigley! I agree that would definitely make the user experience better; I'll see what I can do. I think that might be a bit more complicated than I expect just because as soon as you click that "Continue without an account", an empty profile is stored in local storage; anytime you try to go to the login page from there, it currently re-routes to keep you within the app on seeing that you already have that profile. I'll see if I can find a better way to handle that than just sending you right back to "/profile", though.

taui/src/components/anonymous-authenticator.js
BryanQuigley: Conceptually I just expect handleProfileChange to "handle" loading the profile. It's fine as is, I was just curious.

ktohalloran: Ah, I see. That's a good point--I had made `handleProfileChange` to be a counterpart to Amplify's `withAuthenticator.changeUserProfile` which also doesn't use `loadProfile`, which makes sense within Amplify's structure but maybe not within this less intense `anonymousAuthenticator`. I'll explore that structure a bit more--thanks for pointing that out!

ddohler:
Reassigning to @rachelekm since I'll be out next week.

ktohalloran:
I've made the edit to the automatic re-routing so that pressing cancel before completing the edit profile form takes you back to login instead of the same blank profile page and have added tested instructions above to check that; however, when I tried moving `loadProfile` into `anonymousAuthenticator`'s `handleProfileChange` method, I got strange behavior on refresh as the app re-rendered without triggering `handleProfileChange`. The `loadProfile` method seems to be ok where it is only activating on app render and with local storage updating via `storeConfig`, so I'm going to leave it as is.

taui/src/components/anonymous-authenticator.js
```diff
@@ -0,0 +1,72 @@
...
+import { Component, Fragment } from 'react'
+import LogRocket from 'logrocket'
+
+import {ANONYMOUS_USERNAME, PROFILE_CONFIG_KEY} from '../constants'
+import {storeConfig} from '../config'
+
+import CustomSignIn from './custom-sign-in'
+import CustomHeaderBar from './custom-header-bar'
+
+export default function anonymousAuthenticator (Comp) {
```
rachelekm, line 10: Do we anticipate updating state here for any tokens/credentials when we shift to Django? Would it be better to rename to just `Authenticator` to keep its purpose open or is this just going to be used only for anonymous users?

taui/src/components/custom-sign-in.js
```diff
@@ -42,80 +24,23 @@ class SignInHeader extends React.Component {
...
-              {t('SignIn.AnonymousExplanation') + ' '}
-              <Link theme={theme} onClick={() => this.changeState('useAnonymous')}>
-                {t('SignIn.Anonymous')}
-              </Link>
-            </SectionFooterSecondaryContent>}
-          </SectionFooter>
-        </FormSection>
+        <div className='auth-screen__main auth-main'>
+          <div className='auth-main__anonymous-login'>
+            {t('SignIn.AnonymousExplanation') + ' '}
```
rachelekm:
```suggestion
            {`${t('SignIn.AnonymousExplanation') }`}
```
This doesn't really matter but I think the general pattern in this repo is to use template literals

taui/src/components/anonymous-authenticator.js
```diff
@@ -0,0 +1,72 @@
...
+      this.handleAnonymousLogin = this.handleAnonymousLogin.bind(this)
+      this.handleProfileChange = this.handleProfileChange.bind(this)
+
+      // Load the selected user profile from localStorage, if any
+      this.props.loadProfile()
+    }
+
+    handleProfileChange (profile: AccountProfile) {
+      storeConfig(PROFILE_CONFIG_KEY, profile)
+      this.props.store.dispatch({type: 'set profile', payload: profile})
```
rachelekm:
```suggestion
      this.props.setProfile(profile)
```
This prop calls both and has some error handling

taui/src/components/edit-profile.js
```diff
@@ -248,43 +191,8 @@ class EditProfile extends PureComponent<Props> {
...
-    } else {
-      // Do not attempt to write anonymous profile to S3
-      this.props.changeUserProfile(profile)
-      this.props.history.push('/map')
-    }
+    this.props.handleProfileChange(profile)
+    this.props.history.push('/map')
   }
 
   getUserDataVoucherNumber (userData: any): string {
```
rachelekm: I think we can get rid of `getUserDataVoucherNumber` now since it's not used by anything

rachelekm (commented):
Looks great! I left a couple small comments and then have one more broad thought on the `AnonymousAuthenticator` HOC. I see that at this point the user profile is dual-purpose: temporary profile information as well as acting as the “login” credentials for anonymous users. But I feel like that's causing the purposes of `handleProfileChange` and `handleAnonymousLogin` to blur together a little, and I agree that it would be nice to move `this.props.loadProfile()` (although I see your note above so maybe not this one if it's tricky) as well as the default `anonymousProfile` creation into `handleProfileChange` to better organize all profile changes into one function. That would make `handleAnonymousLogin` kind of obsolete, but I think it might be ok to get rid of it because it’s only being called once at this point — and even then it’s just to set the profile so it’s not really functionally “logging in” or “logging out”. Do you know if we’re going to eventually have some kind of token/credential stored in frontend state for user auth? Maybe in that instance it would be good to keep to handle whatever app-level auth state will exist besides the profile? What do you think?

ktohalloran:
@rachelekm thanks for reviewing! For your questions about how this will work when we're using Django, the answer is I don't know. I talked with @ddohler about this, asking if this card was about just putting something in place to replace Amplify or building a foundation for the Django authentication system and he said he "was envisioning more the first option, under the assumption that the anonymous login is kind of a weird separate thing from the rest of the auth system anyway, so it hopefully shouldn't interfere too much with building out the rest of the system later... it's expected that we won't know exactly what the Django-based auth will look like right now. I would say to try to generally structure things for easy alteration now, but the main goal here is to get rid of Amplify and put a placeholder there so that we still have site access while we work on the rest of the auth, so it's fine if things need to be adjusted later." So based on that, I'm assuming the point of this card is to make anonymous login work as best we can, assuming we can't fully anticipate what Django will need. To that end, I think leaving it named `AnonymousAuthenticator` makes sense, because that's all it may ever be designed to do. But I'll work on consolidating `handleProfileChange` and `handleAnonymousLogin`--you're right they both don't need to exist and doing so will improve readability and (in theory anyway) replaceability.

ktohalloran:
Ok, I've consolidated `handleProfileChange` and `handleAnonymousLogin` into one method that I've renamed to the more generic `handleAuthChange` and made the few other changes you suggested, so this should be ready for re-review.

taui/src/components/custom-header-bar.js
```diff
@@ -1,58 +1,26 @@
...
-  }
-
-  getUserName () {
-    const user = this.props.authData
-    // get name from attributes first
-    const nameFromAttr = user.attributes ? user.attributes.email : undefined
-    return nameFromAttr || user.name || user.username
+    this.setState({
+      userProfile: this.props.userProfile
+    })
```
rachelekm:
```suggestion
```
Since we're not using `userProfile` outside of `render()` and it's already being set to `this.props.userProfile` there, I don't think we need this

rachelekm (commented):
Thanks for clarifying how far to take the anonymous authentication, that makes sense and I agree sticking with the naming seems worth it. I also like the consolidated `handleAuthChange`. To better separate auth and profile info further, does it make sense to call the `setProfile` prop directly instead of `handleAuthChange` when setting favorites or making profile changes once "logged in" or is it better to keep the `handleAuthChange` calls there just so everything's consistent? Beyond that question I saw one small thing I missed earlier, but it looks great!

ktohalloran:
@rachelekm that's a good question that I've wavered on throughout this whole process. I can see how it would make sense to just call `setProfile` directly; however, I think it comes down to how which method of organization we think makes the most sense here. Making calls directly to `setProfile` outside of `anonymousAuthenticator` largely makes the authenticator HOC obsolete and feels like it would be worth taking it completely from the current method of using a HOC toward a more functional method. Choosing between object oriented vs. functional method of solving the problem was something I contemplated when starting the issue and I did initially try to solve the problem functionally; however, I ended up back at the HOC because I found it to be easier to mimic what Amplify's authenticator was doing. That being said, that was just what I found easiest and is not to say that wouldn't work or that using a HOC is the better approach. I think the question is whether we feel that this object oriented method should be replaced for the sake of readability/replaceability/performance.

rachelekm:
Ah cool, thanks for walking me through the process that's helpful context. I see how a HOC is a better fit for our needs and it makes sense to move forward that way 👍

rachelekm (approved):

PR #445, opened 2022-04-11
Upgrade Django image to latest

BryanQuigley:
- removes items that are installed already per new image (psql 13)
 - bumps workers to 2 to be more like what production/staging would look like

## Testing Instructions

 * Just check the admin site still works at localhost:8085/admin/


rachelekm (approved):
This works great. Image updated and I can access `/admin/` as expected 👍  was hoping this might solve my local performance issues but clicked around `/admin/` and eventually bumped into the same problem trying to load the page for `/admin/auth/group/`, see below. Regardless this looks good.
Screen Shot 2022-04-14 at 10 39 57 AM

PR #454, opened 2022-04-15
Serve static files at Django root

### LINKED ISSUE #424: Django root should serve static files
This did not get accomplished as part of #415.

In order to serve static files and the API on the same host, navigating to the root path `/` should cause the static site to be served. Currently it just returns a Django 404 page. Whitenoise and `django-spa` are supposed to be handling this, but clearly something's not working correctly.

rachelekm:
## Overview

We are currently using `django-spa` and a `mastarm` proxy on `start` to serve static files, however these static files were previously inaccessible to `django-spa`. This mounts the static assets created through `mastarm` to the django `/static` folder, allowing `django-spa` to find `index.html` on call to `/`.

### Notes

- The `index.html` file is not generated as part of the `mastarm` bundler and needed to be separately copied in, which is unideal. Ultimately, this serves as a temporary fix until we can move forward with a different bundler in #376.

## Testing Instructions

 * `scripts/update`
 * `scripts/server`
 * Navigate to `http://localhost:8085/` and confirm static page loads

Resolves #424

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,6 +28,7 @@ services:
     volumes:
       - ./src/backend:/usr/local/src/backend
       - $HOME/.aws:/root/.aws:ro
+      - ./taui/assets:/usr/local/src/backend/static/assets
     working_dir: /usr/local/src/backend
     depends_on:
       database:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,6 +29,7 @@ services:
       - ./src/backend:/usr/local/src/backend
       - $HOME/.aws:/root/.aws:ro
       - ./taui/assets:/usr/local/src/backend/static/assets
+      - ./taui/index.html:/usr/local/src/backend/static/index.html
     working_dir: /usr/local/src/backend
     depends_on:
       database:
--- a/src/backend/echo/settings.py
+++ b/src/backend/echo/settings.py
@@ -149,7 +149,6 @@ USE_TZ = True
 
 STATIC_URL = '/static/'
 STATIC_ROOT = os.path.join(BASE_DIR, 'static')
-STATICFILES_DIRS = ((os.path.join(STATIC_ROOT, 'assets')),)
 
 # Set the django-spa static file storage:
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -28,6 +28,7 @@ services:
     volumes:
       - ./src/backend:/usr/local/src/backend
       - $HOME/.aws:/root/.aws:ro
+      - ./taui/assets:/usr/local/src/backend/static/assets
```
ktohalloran, line 31:
```suggestion
      - ./taui/assets:/usr/local/src/backend/static/assets
      - ./taui/index.html:/usr/local/src/backend/static/index.html
```

Unfortunately, copying `index.html` as a command didn't work for me; however, mounting it as its own volume did. I'm not sure whether this is best practice but it makes sense to me that this file could persist across Docker containers like the rest of the stuff in `taui/assets`.

docker-compose.yml
```diff
@@ -40,6 +41,7 @@ services:
       - "--error-logfile=-"
       - "--log-level=debug"
       - "echo.wsgi"
+      - "cp ./taui/index.html /usr/local/src/backend/static"
```
ktohalloran:
```suggestion
```
If mounting `index.html` as its own volume works, this is no longer necessary.

ktohalloran (changes requested):
I learned a lot about Docker volumes from reviewing this so thanks for adding me as a reviewer! As you can see from my comments, I wasn't able to get it to run as is; however, this was a tricky one trying to work around `mastarm` and the frontend and backend not sharing a root the same way they do in other projects, so good job isolating and addressing the issue on this one!

docker-compose.yml
rachelekm: I'm thinking I must have already had the `index.html` file in my `/static` folder after trying to mount from before, so it was able to work for me locally but not actually in practice. This way makes sense to me and works well, so just pushed that change!

ktohalloran (approved):
Looks good! 👍

maurizi (approved):

Issue #452, opened 2022-04-14
Add django testing and linting to ./scripts/test


BryanQuigley:
Forked from #415

Requires a bigger discussion, but for linting we may want to consider using the GH super linter:
See https://github.com/azavea/operations/commit/b562dbb60659fc03b7fc7940410d0d5770ef54d7

BryanQuigley:
Needs to be added to CI as well - test currently depends on amplify.

PR #455, opened 2022-04-19
Add Django linting/formatting/testing to CI

### LINKED ISSUE #429: Add Python linting and autoformatting in CI
- Run Python linting and Black in check mode in CI
- Allow reformatting Python code by running `scripts/format`

rachelekm:
## Overview

Adds Python linting and formatting checks as well as Django tests to CI. Also allows testing linting/formatting/Django tests locally through `scripts/test` as well as reformatting locally through new `scripts/format` file.

### Notes

- This implements [GH Super-Linter ](https://github.com/azavea/operations/commit/b562dbb60659fc03b7fc7940410d0d5770ef54d7) for formatting and linting. Although it can format/lint the entire codebase, I set it to check python code in `src/backend/` exclusively. Right now, this checks every file (except `migrations/*` and `manage.py`) with: `gitleaks`, `jscpd`, `black`, `pylint`, `flake8`, and any `json` with `eslint`. I turned off formatting with `isort` since that threw a handful of errors and wasn't something we normally check, but that could easily be turned back on or conversely we could limit to just `black` and `flake8` for now.
- I debated adding `taui` files to the GH Super-Linter check, but it felt worth waiting until we switch to a different bundler and potentially moving frontend/static assets into `src/` directory.
- After adding `scripts/test` to CI, I originally received an error that Django was unable to find the `database` host to run tests. I brought up `database` in `cibuild` so it could be found by the github actions host, and I believe it should stop at the end of the workflow and doesn't affect staging since it's not creating a new tagged container image. However, this is one area where I still have a fuzzy understanding, so it may be worth editing if there are potential side effects.

## Testing Instructions

 * `scripts/update`
 * Create a formatting error somewhere in `src/backend/`
 * `scripts/test`
 * Confirm formatting fails
 * `scripts/format`
 * `scripts/test`
 * Confirm error is reformatted correctly and linting/formatting/testing passes locally
 * Confirm linting/formatting/testing is correctly running in CI through output

Resolves #429 , #452


scripts/test
```diff
@@ -22,5 +22,31 @@ then
...
+                -e RUN_LOCAL=true \
+                -e VALIDATE_ALL_CODEBASE=false \
+                -e IGNORE_GITIGNORED_FILES=true \
+                -e FILTER_REGEX_INCLUDE="src/backend/" \
+                -e FILTER_REGEX_EXCLUDE=".*(migrations/*|manage.py).*" \
+                -e LINTER_RULES_PATH="/" \
+                -e PYTHON_BLACK_CONFIG_FILE="src/backend/pyproject.toml" \
+                -e PYTHON_FLAKE8_CONFIG_FILE="src/backend/setup.cfg" \
+                -e VALIDATE_PYTHON_ISORT=false \
+                -v $PWD:/tmp/lint github/super-linter:slim-latest
```
ddohler: The GitHub action is pinned to the `v4` tag, so this should be pinned as well. I think this should be all that's required:
```suggestion
                -v $PWD:/tmp/lint github/super-linter:v4
```

src/backend/api/admin.py
```diff
@@ -1,3 +1 @@
-from django.contrib import admin
```
ddohler, line 1: It might be better to comment this out rather than deleting it; if we make changes to this file the first thing we'll have to do is re-import this (the same goes for the other default Django files).

src/backend/setup.cfg
```diff
@@ -0,0 +1,7 @@
+[flake8]
+max-line-length = 100
+exclude = .git,*migrations*,manage.py
+select = E,F,W,C90
+# E203, E501, and W503 directly or partially conflict with Black
+# E266 (Too many leading '#' for block comment) conflicts with old usage that's not worth changing
```
ddohler: I think that E266 can be re-enabled for this project; we don't have any legacy Django code here.

scripts/test
```diff
@@ -22,5 +22,31 @@ then
...
+            echo "Checking Django formatting and linting using GH Super-Linter..."
+            docker run --rm \
+                -e RUN_LOCAL=true \
+                -e VALIDATE_ALL_CODEBASE=false \
+                -e IGNORE_GITIGNORED_FILES=true \
+                -e FILTER_REGEX_INCLUDE="src/backend/" \
+                -e FILTER_REGEX_EXCLUDE=".*(migrations/*|manage.py).*" \
+                -e LINTER_RULES_PATH="/" \
+                -e PYTHON_BLACK_CONFIG_FILE="src/backend/pyproject.toml" \
+                -e PYTHON_FLAKE8_CONFIG_FILE="src/backend/setup.cfg" \
```
ddohler, line 47: I'm hesitating about adding additional linters / formatters here; we've had issues with Black and Flake8 occasionally getting into arguments, so I would expect that adding more tools would result in more conflicts.

`isort` seems to [explicitly want to play nicely with `black`](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html), so it might be worth trying that to see if that addresses the problems you encountered.

`pylint` I'm less sure about since it seems to overlap significantly with `flake8`. But it's relatively easy to turn on or off so I don't see any harm in leaving it on for now, and if it causes problems we can turn it off or add configuration later.

ddohler (approved):
Works great, thanks for putting this together! I love the easy integration of the super linter with GitHub Actions. I had a few minor comments but this is good to go after those are cleaned up!

scripts/test
rachelekm: Sounds good and makes sense, they _happened_ to play nice this time around but could see it being trouble later on. I'll take out everything except `black`, `flake8`, `isort`, and `pylint` but I'll add in a comment that if `pylint` causes any trouble just take it out.

PR #456, opened 2022-04-21
Add fields to UserProfile

### LINKED ISSUE #248: Replace GoSection8 with Apartments.com
Replace (unimplemented) link to GoSection8 in the neighborhood details with a deep link to Apartments.com.

Supersedes #133.

ktohalloran:
## Overview

This PR adds fields to the `UserProfile` model and also adds a `Destination` model to store destinations associated with each `UserProfile`. This enforces two rules: each profile must have at least one destination and exactly one primary destination.

### Demo
Screen Shot 2022-04-21 at 2 05 29 PM

Screen Shot 2022-04-21 at 2 17 52 PM

All references to "username" in both the login and create user forms have been renamed to "email".

Screen Shot 2022-04-21 at 2 17 06 PM

All fields, including a `TabularInline` with which to add and edit a profile's associated destinations are editable from the admin.

Screen Shot 2022-04-21 at 2 08 53 PM

Screen Shot 2022-04-21 at 2 09 10 PM

Users see an error if they either have no destinations or not exactly one primary destination.

### Notes

This PR has renamed the "username" field to "email" in and removed the actual email field from the admin UI so that users know what to enter and so that the signup experience from the frontend and backend will be the same, e.g. counselors won't have usernames while house seekers don't, etc. This does not change the sign up experience from the command line, where users are still prompted for both an email and username. 

As a note for future work connecting this to the frontend, I've set up a `full_name` field on the `UserProfile` model; however, this is different from the default `User` name fields, which has them as two separate fields. While we can squish first and last names together for storage, separating a full name into separate names can be tricky; this may be worth noting when connecting the form to user and profile creation.

## Testing Instructions

 * Run `./scripts/manage migrate` to update your database.
 * Run `./scripts/server` and navigate to **localhost:8085/admin/**. 
 * Start by creating a new user. Verify that if you enter an invalid email address in the email/username field, it will respond with an error.
 * Then go the User Profile model and add a new profile. 
 * First, test that it is properly enforcing the profile has at least one destination. Complete the form but don't add an address. Click "save" and verify that it responds with an error.
 * Next, text that it is properly enforcing that you have exactly one primary destination. Add an address but do not select that it is a primary destination. Click "save" and verify that it responds with an error.
 * Add another address and select that both are primary destinations. Click "save" and verify that it responds with an error.
 * Finally, remove one of the primary destination checks and click "save". Verify that it saves at it should.

Resolves #248


src/backend/users/admin.py
```diff
@@ -1,18 +1,91 @@
...
+    class Meta:
+        model = User
+        fields = '__all__'
+
+    def clean(self):
+        cleaned_data = super(CreateUserAdminForm, self).clean()
+        try:
+            validate_email(cleaned_data['username'])
+            return cleaned_data
+        except:
```
ddohler: In general, `except` should always include a specific error type that they're designed to handle; blanket `except`s like this can hide errors. For example, if `cleaned_data` didn't have a field named `username` for some reason, that would raise a `KeyError`, but it would still get handled by the `except` and converted into a `ValidationError`, which would provide a false trail for someone debugging the issue. This will get caught by linting once we've got it on this project (and also is an indication that you may want to add additional Python linting to your editor so that you can catch these issues regardless of how the project is configured).

However, in this particular case, `validate_email` raises a `ValidationError` already, so unless the default error message isn't descriptive enough, I think it should be possible to remove the try/except entirely here and just allow the default error to bubble up.

src/backend/users/admin.py
```diff
@@ -1,18 +1,91 @@
...
+        fields = '__all__'
+
+    def clean(self):
+        if self.is_valid():
+            count_primary_destinations = 0
+            for i in self.cleaned_data:
+                if i.get('primary_destination'):
+                    count_primary_destinations += 1
+            if count_primary_destinations != 1:
+                raise forms.ValidationError('Please select one primary address')
```
ddohler:
```suggestion
                raise forms.ValidationError('Please select exactly one primary address')
```

src/backend/users/admin.py
```diff
@@ -1,18 +1,91 @@
...
+
+class DestinationAdminForm(BaseInlineFormSet):
+    class Meta:
+        model = Destination
+        fields = '__all__'
+
+    def clean(self):
+        if self.is_valid():
+            count_primary_destinations = 0
+            for i in self.cleaned_data:
```
ddohler: This could be written somewhat more succinctly with a list comprehension and at least to my eye it's easier to understand, too:

```python
primary_destinations = [
    dest for dest in self.cleaned_data if dest.get("primary_destination")
]
if len(primary_destinations) != 1:
    raise forms.ValidationError("Please select exactly one primary address")
```

src/backend/users/models.py
```diff
@@ -2,4 +2,46 @@
 from django.contrib.auth.models import User
 
 class UserProfile(models.Model):
+    class TravelMode(models.TextChoices):
+        CAR = "CAR", "car"
+        BUSTRAIN = "BT", "bus/train"
```
ddohler:
```suggestion
        BUS_TRAIN = "BT", "bus/train"
```
I think this (and the others here) would be more readable with underscores added, though it's also fine the way it is if you don't want to have to change all the migrations.

src/backend/users/models.py
```diff
@@ -2,4 +2,46 @@
...
+    class TripPurpose(models.TextChoices):
+        WORK = "WK", "Work"
+        DAYCARE = "DC", "Day care"
+        FAMILY = "FA", "Family"
+        FRIENDS = "FR", "Friends"
+        WORSHIP = "WP", "Worship"
+        DOCTOR = "DR", "Doctor"
+        OTHER = "OT", "Other"
+    profile = models.ForeignKey(UserProfile, on_delete=models.CASCADE)
+    address = models.CharField(max_length=100)
```
ddohler: I think that 100 is almost certainly going to be enough, but there's essentially no cost to making the limit higher, and it would be quite disruptive to a user whose address happened to exceed the limit, so I'd suggest doubling this.

```suggestion
    address = models.CharField(max_length=200)
```

src/backend/users/models.py
```diff
@@ -2,4 +2,46 @@
...
+
     user = models.OneToOneField(User, on_delete=models.CASCADE)
+    username = models.EmailField(unique=True)
+    full_name = models.CharField(blank=True, max_length=60)
+    has_voucher = models.BooleanField(default=False)
+    voucher_number = models.CharField(blank=True, max_length=10)
+    voucher_bedrooms = models.IntegerField(blank=True, null=True)
+    rent_budget = models.IntegerField(blank=True, null=True)
+    desired_bedrooms = models.IntegerField(blank=True, null=True)
+    travel_mode = models.CharField(choices=TravelMode.choices, max_length=40, default=TravelMode.BUSTRAIN)
```
ddohler: When adding `choices` to a `CharField`, the data that gets stored in the database is the first item in the choices tuple (so in this case, `CAR`, `BT`, or `BTE`). The data being stored is what the `max_length` parameter applies to, so in this case that parameter only needs to be 3, not 40 (and similar for the choice fields, below). The second item in the tuple is just a human-readable label that gets displayed whenever the choices are rendered into a dropdown menu in the admin.

src/backend/users/models.py
```diff
@@ -2,4 +2,46 @@
...
+        BUSTRAINEXPRESS = "BTE", "bus/train/express"
+
+    class Priorities(models.TextChoices):
+        NOTIMPORTANT = "NI", "Not important"
+        SOMEWHATIMPORTANT = "SI", "Somewhat important"
+        IMPORTANT = "I", "Important"
+        VERYIMPORTANT = "VI", "Very important"
+
     user = models.OneToOneField(User, on_delete=models.CASCADE)
+    username = models.EmailField(unique=True)
```
ddohler: Sorry if this was confusing in the issue; this is duplicative of the field in the `User` model, which is required, so I don't think it's necessary to have it on this model as well.

src/backend/users/models.py
```diff
@@ -2,4 +2,46 @@
...
+    full_name = models.CharField(blank=True, max_length=60)
```
ddohler: This roughly duplicates the `full_name` and `last_name` fields on the base `User` model, but it's a trickier case than the username / email duplication because the existing profile editing screen only asks for a full name, not a first / last name. As you noted, it's easy to create a full name from a first/last name pair, but not so easy to go the other way, so we can't just go with the Django default fields without switching the profile editing screen too. 

For now, could you try hiding the first / last name fields on the user admin? They'll still be there in the database but at least we won't confuse any administrators with them. I think it would be also good to add a comment to this line as well noting that the first / last name fields are still present but unused.

ddohler: Also, a similar comment to the address field below: at small scales, there's very little downside to making this field overly large, and a significant downside to making it too small, so I'd err on the side of expansiveness even though I think 60 characters is likely to be enough for most names. I'd suggest bumping this to 100 or even higher.

ddohler (changes requested):
This works great! There are some Django / Python style things that need to be addressed but overall this is looking good, nice work!

ktohalloran:
@ddohler thanks for reviewing! I've made the changes as requested so this should be ready for re-review.

src/backend/users/models.py
```diff
@@ -2,4 +2,45 @@
...
+    voucher_bedrooms = models.IntegerField(blank=True, null=True)
+    rent_budget = models.IntegerField(blank=True, null=True)
+    desired_bedrooms = models.IntegerField(blank=True, null=True)
+    travel_mode = models.CharField(choices=TravelMode.choices, max_length=3, default=TravelMode.BUS_TRAIN)
+    commute_priority = models.CharField(choices=Priorities.choices, max_length=2, default=Priorities.NOT_IMPORTANT)
+    school_quality_priority = models.CharField(choices=Priorities.choices, max_length=2, default=Priorities.NOT_IMPORTANT)
+    public_safety_priority = models.CharField(choices=Priorities.choices, max_length=2, default=Priorities.SOMEWHAT_IMPORTANT)
+
+    def __str__(self):
+        return self.full_name
```
ddohler: I just noticed one minor thing: because `full_name` isn't required, this can result in the string representation being an empty string if that field's not available, e.g.:
Screen Shot 2022-04-27 at 08 37 21

I'd suggest using `self.user.email` instead since that's a required field.

ddohler (approved):
Looks good! I noticed one small additional thing, but this is good to go once that's addressed!

ddohler:
Sorry, forgot one other thing: this has some formatting / linting issues that will cause the build to break once #455 is merged, so you should coordinate with @rachelekm on merge order for the two PRs and then rebase one or the other and run `/scripts/format` to pick up the files that are being added here.

PR #457, opened 2022-04-21
Switch frontend app bundler

### LINKED ISSUE #376: Switch to a different bundler

aaronxsu:
## Overview

This PR has made the following changes as a result of switching away from `mastarm`:
- [X] move `taui` and rename to `src/frontend`
- [X] remove `mastarm` and its related config from `./taui/configurations`
- [X] switch to webpack + babel and add configs for them
- [X] update paths and volume mounting in `docker-compose` for `app` and `django` services
- [X] serve built frontend as static assets from django root
- [X] hard-code redux store init values from what used to be in `store.yml`
- [X] read from `.env` on frontend
- [X] update task runner scripts in `package.json`
- [X] update imports in JS due to changed directory structure
- [x] linter and formater
- [x] update CI/CD script if need be

### Demo

App served from Django root
Screen Shot 2022-04-26 at 6 18 40 PM

App in the map view
Screen Shot 2022-04-26 at 6 18 56 PM

### Notes
1. I tried to keep the formatter change and the linter change separate from the meaningful code changes
2. The following will be in their own issues:
- handle env-based dotenv files better, pull dotenv file in bootstrap script
- favicons

## Testing Instructions

 * Pull down this branch
 * At the root of the repo, run: `cp ./src/frontend/.env.template .env`. Then fill in the fields in `.env` by the one stored on lastpass
 * At the root of the repo, run `./scripts/update`, which will build the frontend and backend, install frontend deps, create a static build of the frontend to be used by django, and run django manage scripts
 * AT the root of the repo, run `./scripts/server`, which will bring both frontend and backend up and running. Navigate to `http://localhost:9966` to access app frontend. If you make changes to frontend code, module hot reloading should work. Navigate to `http://localhost:8085` to access the static frontend app served from django.

Resolves #376


docker-compose.yml
```diff
@@ -27,6 +29,7 @@ services:
       dockerfile: Dockerfile
     volumes:
       - ./src/backend:/usr/local/src/backend
+      - ./src/frontend/build:/usr/local/src/backend/static/
```
aaronxsu, line 33: Inspired by changes in this PR: https://github.com/azavea/echo-locator/pull/454. This way, the frontend static app should be served from django root.

src/backend/Dockerfile
```diff
@@ -4,7 +4,16 @@ RUN mkdir -p /usr/local/src/backend
 WORKDIR /usr/local/src/backend
 
 COPY requirements.txt /usr/local/src/backend/
-RUN pip install --no-cache-dir -r requirements.txt
+RUN set -ex \
```
aaronxsu: I am not sure why these were removed in a previous PR, since I think the postgres client should be needed for django to work with a postgres DB container. Any suggestions?

src/frontend/.eslintrc.json
```diff
@@ -28,11 +26,13 @@
...
     "flowtype/sort-keys": [
       "error",
-      "asc", {
-        "caseSensitive": true,
-        "natural": false
-      }
-    ]
+      "asc"
+    ],
+    "array-callback-return": 0,
```
aaronxsu, line 31: Rules from this line and below were added to make the linter less mad. I think we should open an issue to remove some or all of the following rules, and actually make the corresponding code changes, since I think some of these rules are actually legit and should not be ignored. Any suggestions?

src/frontend/package.json
```diff
@@ -0,0 +1,117 @@
+{
+  "name": "@azavea/echo-locator",
+  "description": "Accessibility analysis for housing",
+  "main": "src/index.js",
+  "scripts": {
```
aaronxsu, line 5: Task runner scripts are now updated so that `mastarm` is not used anywhere

src/frontend/package.json
```diff
@@ -0,0 +1,117 @@
...
+    "react-select": "~1.3.0",
+    "react-select-fast-filter-options": "~0.2.3",
+    "react-tooltip": "~3.11.2",
+    "react-virtualized-select": "~3.1.3",
+    "redux": "~4.0.5",
+    "redux-actions": "~2.6.5",
+    "redux-mock-store": "1.5.4",
+    "reselect": "~4.0.0"
+  },
+  "devDependencies": {
```
aaronxsu, line 71: Majority of the dev deps added here are due to the removal of `mastarm`.

src/frontend/webpack.config.js
```diff
@@ -0,0 +1,100 @@
+const path = require("path");
+const HtmlWebPackPlugin = require("html-webpack-plugin");
+const webpack = require("webpack");
+
+require("dotenv").config({ path: "./.env" });
```
aaronxsu: This enables webpack to pick up env vars from `.env`.

src/frontend/webpack.config.js
```diff
@@ -0,0 +1,100 @@
...
+
+module.exports = (env) => ({
```
aaronxsu: I added this `env` argument here, so that in the next issue/task, different `.env` files should be pulled depending on if it is in dev, stg, or prd- for example, we may have `.env.dev`, `.env.staging`, `.env.production`, etc. And then we should use this `env` argument to see which `.env.<something>` to use, before creating a build.

src/frontend/webpack.config.js
```diff
@@ -0,0 +1,100 @@
...
+                [
+                  "@babel/preset-env",
+                  {
+                    targets: "defaults",
+                  },
+                ],
+                [
+                  "@babel/preset-react",
+                  {
+                    runtime: "automatic",
```
aaronxsu, line 38: By default, it is `runtime: "classic"`, but because we do not `import React from "react"` everywhere, we need to tell it to babel. See more here: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#manual-babel-setup

src/frontend/webpack.config.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+    client: {
+      progress: true,
+    },
+    compress: true,
+    port: 9966,
+    hot: true,
+    open: true,
+    proxy: {
+      "/api": "http://django:8085",
```
aaronxsu, line 94: I think this should route `/api` calls to `http://django:8085/api` locally, if you run the frontend app from `http://localhost:9966`

src/frontend/package.json
```diff
@@ -0,0 +1,117 @@
...
+    "focus-visible": "~5.0.2",
+    "geojson-bbox": "~0.0.0",
+    "i18next": "^21.4.0",
+    "isomorphic-fetch": "~2.2.1",
+    "jsolines": "~2.1.2",
+    "leaflet": "~1.5.1",
+    "lodash": "~4.17.15",
+    "logrocket": "^1.0.6",
+    "logrocket-react": "~4.0.1",
+    "react": "~16.14.0",
```
aaronxsu, line 52: I updated the minor version of `react` and `react-dom` from `16.13.0` to `16.14.0`, because we do not use `import React to "react"` everywhere, and:

>[React 17 RC](https://reactjs.org/blog/2020/08/10/react-v17-rc.html) already includes support for the new transform, so go give it a try! To make it easier to adopt, we’ve also backported its support to React 16.14.0, React 15.7.0, and React 0.14.10. You can find the upgrade instructions for different tools [below](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#how-to-upgrade-to-the-new-jsx-transform).

See more here: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-a-jsx-transform

src/frontend/webpack.config.js
ddohler: I think there's a way to do this without adding this dependency:
- Place `.env` at the project root alongside docker-compose
- Specify `env_file` for `app` service pointing to `.env`.

Then this can be commented out and the `DefinePlugin` on line 76 of this file is enough to convert `process.env` into Webpack's `"process.env"`.

Then we can define environment variables by injecting them into the environment, which I believe all of our build / deployment platforms already support, without having to separately download and manage .env files for each deployment target.

src/frontend/webpack.config.js
ddohler: See the above comment; I think it might be possible to define different sets of environment variables directly in our build tools rather than managing separate files. Does that seem like a preferable approach to you?

src/frontend/src/index.js
```diff
@@ -27,9 +27,6 @@ import "./taui.css";
 // Set the title
 document.title = message("Title");
 
-// Opt-in to Webpack hot module replacement
-if (module.hot) module.hot.accept();
```
ddohler: Just curious, what's going on with this?

src/frontend/.eslintrc.json
ddohler: Yeah, let's open a tech-debt issue to address these (or maybe several issues if you think there are enough that it would be better to break them up).

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '2.1'
 services:
   app:
-    image: node:12-slim
+    image: node:12
```
ddohler: Can this commit be rebased out?

src/backend/Dockerfile
ddohler: Based on the [commit that removed them](https://github.com/azavea/echo-locator/commit/044c3cc49866b6a57bb8fadbdca6eddd8656c4ae) it looks like they were assumed to be already installed in the `azavea/django` container -- are you finding that they're missing?

src/frontend/Dockerfile
```diff
@@ -3,9 +3,9 @@ FROM node:12-slim
...
 RUN set -ex; \
   apt-get update && \
-        apt-get install -y git
+  apt-get install -y git
 
 COPY . /usr/local/src
 WORKDIR /usr/local/src
 
-RUN yarn install
+RUN yarn install --production false
```
ddohler: Should we be using `--frozen-lockfile` here?

ddohler (changes requested):
This is looking really good! I had a few comments, but the main thing I'm uncertain about is whether neighborhood images are working or not, and how to check. When I view the map page, I get several `404` errors in the console looking for `http://localhost:9966/assets/neighborhoods/02203_town_square.jpg` in order to display the neighborhood thumbnails. 

Under mastarm, the neighborhood thumbnails are downloaded by running `./neighborhood_data/update_data.sh` and then (I think?) manually copying the images to the assets folder. We are planning to migrate neighborhood info to the database so that it can be managed via the Django API, so I think we should avoid making major changes to the neighborhood data management process (because they'll just get thrown away in a few sprints), but we should make sure that we don't break the existing neighborhood images.

src/frontend/src/index.js
aaronxsu: Ah, I later found that this was not necessary, since `devServer.hot: true` does the job, which is in `webpack.config.js`.

src/frontend/Dockerfile
aaronxsu: Ah yes, good point. I will add it!

src/backend/Dockerfile
aaronxsu: Alright, I removed them locally and it seems fine. Lemme remove these changes!

src/frontend/webpack.config.js
aaronxsu: Yes. This sounds like a great approach to me!

src/frontend/webpack.config.js
aaronxsu: Sounds good. This approach, with some added configs, works well: https://github.com/azavea/echo-locator/pull/457/commits/a879f6bdb2a81eab6dc20343709a66451c8a7836

aaronxsu:
@ddohler good find on the bug that the neighborhood image thumbnails not showing up. I fixed it in this commit: https://github.com/azavea/echo-locator/pull/457/commits/16dfc1d865a5ad2c35a6fc8bee7d0a49a076bb1c . It was simply because `assets` subdirectory do not exist anymore under this new bundler configuration. Such assets should now be in `frontend/public`. 

If you still have those thumbnails lying around in the `neighborhood_data` directory, please feel free to do something similar to these changes: https://github.com/azavea/echo-locator/pull/457/commits/16dfc1d865a5ad2c35a6fc8bee7d0a49a076bb1c#diff-a7a011c1c5fb7bd090210ada6e19390c174069715fab4772fd4e614789a47e26R16 - just copy everything in `neighborhood_data/images` to `src/frontend/public/neighborhoods`. Then `./scripts/server` should run the frontend with those images. If you don't have the thumbnails anymore, re-running the `./update_data.sh` in `neighborhood_data` directory should do it automatically.

This gets me thinking:
1. If this step should be added in the `cibuild` before creating the static frontend bundle
2. Do these images ever change? Or are they different per environment? If they are really static, would it be easier, for both dev and for deployment, to upload them all to S3, and pull them down before bundling frontend? Since the current workflow seems to be running the `update_data` script first no matter what.
3. Could you share a bit more about your vision of having these images in DB? What would the schema be like for such a table? Does it mean all neighborhood JSON data will be in a new table too? Should they be stored as bytes or base64 encoded string? Because I think this might be a better approach than 1 and also 2 above.

aaronxsu:
I am aware of the merge conflicts, and I plan to address them before merging.

src/frontend/.eslintrc.json
aaronxsu: Agreed. I created an issue here: https://github.com/azavea/echo-locator/issues/483

src/frontend/webpack.config.js
```diff
@@ -0,0 +1,110 @@
...
+      "GOOGLE_ANALYTICS_ID",
+      "LOGROCKET_APP_ID",
+      "LOGROCKET_RELEASE",
+      "ALLOW_ANONYMOUS",
+      "LEAFLET_TILE_URL",
+      "LEAFLET_RETINA_URL",
+      "LABEL_URL",
+      "LABEL_RETINA_URL",
+      "LEAFLET_ATTRIBUTION",
+      "MAPBOX_ACCESS_TOKEN",
```
ddohler: I'm still a bit confused why this is necessary. I tried out the following for `plugins` and it seems to be working fine for me:

```javascript
    plugins: [                                                                                        
      new HtmlWebPackPlugin({                                                                         
        template: "./public/index.html",                                                              
      }),                                                                                             
      new webpack.HotModuleReplacementPlugin(),                                                       
      new webpack.DefinePlugin({                                                                      
        "process.env": JSON.stringify(process.env),                                                   
      }),                                                                                             
    ],                                                                                                
```

And here's my .env:
```
GOOGLE_ANALYTICS_ID=
... etc.
```

Is there something I'm missing that makes that approach not work?

src/frontend/webpack.config.js
```diff
@@ -0,0 +1,110 @@
+const path = require("path");
+const HtmlWebPackPlugin = require("html-webpack-plugin");
+const webpack = require("webpack");
+
+module.exports = (env) => ({
```
ddohler: I think this `env` variable is no longer used, unless I'm missing something?

ddohler (approved):
Looks good to me! I had one further question about how environment variables are being injected, but I think this is ready to go once we've got that sorted out and the merge conflicts are resolved.

> Could you share a bit more about your vision of having [neighborhood] images in DB? What would the schema be like for such a table? Does it mean all neighborhood JSON data will be in a new table too? Should they be stored as bytes or base64 encoded string? Because I think this might be a better approach than 1 and also 2 above.

Yes, we're going to migrate all the data that's currently in the `neighborhoods.json` and `neighborhood_bounds.json` files into database fields. For the neighborhood images specifically, I'm not exactly sure how we'll handle them -- I think storing them as binary blobs in the database could work since they're relatively small, or maybe it would be better to upload them to S3. But either way, we should allow admins to edit the images from the Django admin page, similar to this: https://github.com/azavea/cac-tripplanner/blob/develop/python/cac_tripplanner/cms/models.py#L96 . That way the admins can change the photos or add new neighborhoods if they need to. How does that sound?

scripts/cibuild
```diff
@@ -23,9 +23,22 @@ if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
...
         usage
     else
+        echo "Running update script"
         REACT_APP_GIT_COMMIT="${GIT_COMMIT}" \
             ./scripts/update
 
+        echo "Running tests"
+        # ./scripts/test
+
+        # TODO: set env vars for frontend
```
ddohler: Sorry, one other thing I missed -- this comment should have a link to #487 .

aaronxsu:
@ddohler, you are right about how env vars are read in through the way you suggested, I made the change here: https://github.com/azavea/echo-locator/pull/457/commits/14ca8204ca0db972c1366ad0a9a9772069790ad1

Regarding your vision of how neighborhood data will be stored and managed, yes it sounds good and makes sense to me.

Somehow I was seeing issues fetching json from those cloudfront URLs again. Did you see that from your end when doing the re-review?

ddohler:
> Somehow I was seeing issues fetching json from those cloudfront URLs again. Did you see that from your end when doing the re-review?

~Oh, good catch! I didn't actually check the console when I was reviewing, but I do see those errors now. But I'm also seeing this error and I think it may be the root cause: `Uncaught Error: Bounds are not valid.`~

~If the bounds aren't valid then it's similar to if you tried to use an address outside the Boston area, and you might see errors loading the network files. I suspect this indicates that `neighborhood_bounds.json` may not be being loaded properly. I don't see a request for it on this branch but on `develop` I do.~ As discussed in RRP, this seems to be happening across all environments, so it's unlikely to be something specific to this branch.

Issue #449, opened 2022-04-14
Bastion alternatives (or finish bastion)


BryanQuigley:
A bastion is being deployed (included in vpc terraform), but it's not fully setup.

This ticket is to evaluate if it's still needed as is, namely:

**Get along without one entirely?**
[Connect directly to fargate containers](https://aws.amazon.com/blogs/containers/new-using-amazon-ecs-exec-access-your-containers-fargate-ec2/)

And start up a CLI container when we actually need to do those kinds of actions (which is mostly DB related?).

Ideal - we only use resources when we need them and don't need project SSH keys. AWS accounts are the only way in (via SSM)

**Switch to using AWS SSM for access to bastion**
[Use SSM  to access bastion instead of SSH directly](https://aws.amazon.com/blogs/mt/replacing-a-bastion-host-with-amazon-ec2-systems-manager/)

OK - Still has deploy a bastion, but no external SSH access.  Don't need SSH keys for project.

If either move away from bastion is decided, we need:
* Move away from the vpc module (or evaluate updating that module)
* Write up new terraform to chosen path

PR #470, opened 2022-04-29
Enable Docker Exec

BryanQuigley:
## Overview
This enables the ability to exec into the running app container.

This removes access to the bastion but doesn't actually remove it
becuase that depends on reworking VPC terraform module or moving from it.

Related: #449

### Demo

$ aws ecs execute-command     --region us-east-1    --cluster ecsechostgdjangoCluster    --task <need_current_task_id>        --container django    --command "/bin/bash"    --interactive

The Session Manager plugin was installed successfully. Use the AWS CLI to start a session.

Starting session with SessionId: ecs-execute-command-0c855e0558b22fb37
root@ip-10-0-1-103:/usr/local/src/backend:

Confirm could get to psql.

### Notes

Future work:
 * Removing the bastion will actually depend on more work to separate from VPC module or update VPC to support bastion-less.
 * logging practices
 * process around using it in production (appcli, create separate task, etc?)

## Testing Instructions

1. Check local setup with https://github.com/aws-containers/amazon-ecs-exec-checker (need AWS with SSM)
2. Try demo above. 
3. For bonus connect to database.

### DIFF
--- a/deployment/terraform-django/app.tf
+++ b/deployment/terraform-django/app.tf
@@ -148,6 +148,7 @@ resource "aws_ecs_service" "app" {
 
   launch_type          = "FARGATE"
   platform_version     = var.fargate_platform_version
+  enable_execute_command = true
   force_new_deployment = true
 
   network_configuration {
--- a/deployment/terraform-django/dns.tf
+++ b/deployment/terraform-django/dns.tf
@@ -30,14 +30,6 @@ resource "aws_route53_zone" "external" {
   name = var.r53_public_hosted_zone
 }
 
-resource "aws_route53_record" "bastion" {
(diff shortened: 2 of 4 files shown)


rachelekm (approved):
Great! After pushing up yesterday's PRs in attempts to fix staging the current task no longer enables the `exec` command, so I can't test this now. However, I did test this command earlier shortly after the original task was running and was able to successfully start a session and get to psql 👍

ddohler (approved):
Awesome find! This is a lot easier than what we used to do and should cut down on costs as well. I wasn't able to fully test in staging but I did run the checker script (which is a pretty cool find itself) and also ran the `execute-command` command just to see what it would do. This would make a great TSM lightning talk or #architecture discussion.

Issue #473, opened 2022-05-02
Remove hardcoded debug 


BryanQuigley:
This should ease initial development on staging, but should be removed.

Issue #459, opened 2022-04-26
Fix login sever error on staging environment


rachelekm:
We can currently navigate to `/admin/` and the login page loads correctly on staging. However, when supplying a username and password and attempting to sign in, we receive a 500 server error. Other pages such as `password_reset` also do not load as expected and redirect back to `/admin/login/`. This issue is to investigate these errors and fix our staging environment. One note is our health-check is currently set to match 400 (see issue #441), so if our django app is healthy and pages are responding with 200 it may be causing a failure.

PR #475, opened 2022-05-02
Add django health-check

### LINKED ISSUE #441: Add health-check
Came out of #415:
- There is no django /health-check bit to check for so we are calling everything good if we get a 400 response. (TODO441)

We should wire up a health-check and then change matcher to 200.

rachelekm:
## Overview

This updates the `django-watchman` settings and creates a `/health-check` endpoint for django as well as updates the app resource `health_check` `matcher` to 200 in Terraform so that the instance can respond as healthy.

This is built off of PR #474 to bring together a multi-piece solution to resolving #459. We can't test health-check appropriately without the previous PR and the previous PR doesn't bring staging all the way to function properly without health-check and an update to allowed hosts.

### Notes

- `gazetteercache` was a custom check used in OAR that's not applicable to our `api` module, so I took it out.

## Testing Instructions

 * `scripts/server`
 * Navigate to `http://localhost:8085/health-check/` and confirm json response
 * Confirm https://stg.echosearch.org/ still works as expected
 * Login to AWS console and navigate to ECS --> ecsechostgdjangoCluster --> Tasks --> select current running task --> logs
 * Confirm `/health-check` called in the logs and responding with 200 status

Resolves #441, #459, #473

### DIFF
--- a/deployment/terraform-django/app.tf
+++ b/deployment/terraform-django/app.tf
@@ -43,8 +43,7 @@ resource "aws_lb_target_group" "app" {
   health_check {
     healthy_threshold   = "3"
     interval            = "30"
-    # TODO GH #441: Set this to 200 after we add a health check
-    matcher             = "400"
+    matcher             = "200"
     protocol            = "HTTP"
     timeout             = "3"
     path                = "/health-check/"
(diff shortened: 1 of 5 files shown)


BryanQuigley (approved):
Tested locally and on stg.  Confirmed staging is really running this code.
Locally also tested to make sure if db is down, health check reports bad.

Health-check is working perfectly.  LGTM!

ddohler (approved):
Works great! I verified that the Target Group on staging was using `200` as its success check and that the target was healthy.

PR #478, opened 2022-05-02
Add management command to migrate housing seeker data

### LINKED ISSUE #432: Create management command to migrate housing seeker data from S3
Create a management command that can migrate data from existing profiles into the Django database.

It should:
- Accept AWS credentials to use (I believe boto will automatically look for authentication credentials in the local environment, so simply using boto may be sufficient).
- Accept an S3 bucket as a parameter
- Read profile data from the target S3 bucket and create corresponding user profiles within the associated Django database.

Notes:
Scripts like this often need to be run multiple times with some user intervention in between to validate data, fix problems, etc., and that can be a struggle if the script's functionality doesn't lend itself to that kind of workflow. Some features that are often helpful in such cases:
  - Recovering from errors, printing a useful identifier so that the offending record can be tracked down later, and continuing to process further records.
  - Doing something intelligent when conflicts are discovered (usually that's either updating the record or skipping it as a duplicate)
  - Accepting a subset of IDs or otherwise enabling filtering on the source dataset so that problematic records can be targeted directly rather than having to re-run the import for the entire set.

You shouldn't necessarily implement all of these features; this is just an example of the kind of thing to be on the lookout for; if you find yourself wanting one of these at least once, then it's probably worth adding it because it'll likely be useful again.

aaronxsu:
## Overview

This PR:
- updates the `Destination` model and adds a migration so that a `PointField` of `location` replaces `address`
- adds `boto3` as a backend dependency
- exposes `AWS_PROFILE` env var for django service in `docker-compose`, which should make local development easier if the developer has multiple aws profiles. This should not matter for staging and prod
- adds a management command to migrate housing seeker data from S3 to the postgres db:
```
required arguments:
  -b BUCKET, --bucket BUCKET
                        S3 bucket to read housing seeker profile from
  -p PREFIX, --prefix PREFIX
                        S3 prefix under the bucket that stores housing seeker profile
  -vkts VOUCHER_KEYS_TO_SKIP [VOUCHER_KEYS_TO_SKIP ...], --voucher_keys_to_skip VOUCHER_KEYS_TO_SKIP [VOUCHER_KEYS_TO_SKIP ...]
                        Space separated list of double-quoted voucher numbers.
```

### Notes

See attached original issue for discussions about assumptions and edge case handling.

## Testing Instructions

 * `./scripts/update`
 * `AWS_PROFILE=<your echo profile name>./scripts/manage migrate_housing_seeker_data -b <staging profile bucket name> -p public -vkts ""`
 * The above should work and should insert related staging profiles as records in your local DB
 * Make sure it has an output of summary
 * `AWS_PROFILE=<your echo profile name>./scripts/manage migrate_housing_seeker_data -b <production profile bucket name> -p public -vkts ""`
 * The above should work and should insert related production profiles as records in your local DB
 * Make sure it has an output of summary

Closes #432


src/backend/users/models.py
```diff
@@ -51,7 +52,7 @@ class TripPurpose(models.TextChoices):
         OTHER = "OT", "Other"
 
     profile = models.ForeignKey(UserProfile, on_delete=models.CASCADE)
-    address = models.CharField(max_length=200)
```
ddohler, line 54: We'll need to continue to keep the address field as well so that users can read the street address for each destination.

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+        group.add_argument(
+            "-p",
+            "--prefix",
+            required=True,
+            help="S3 prefix under the bucket that stores housing seeker profile",
+        )
+        group.add_argument(
+            "-vkts",
+            "--voucher_keys_to_skip",
+            required=True,
```
ddohler: Given that this defaults to an empty list, could we make it `required=False`?

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def client(self):
+        return boto3.client("s3")
+
+    def list_s3_object_keys(self, bucket, prefix):
+        s3 = self.s3()
+        s3_bucket = s3.Bucket(bucket)
+        return s3_bucket.objects.filter(Prefix=prefix)
+
+    def get_s3_object_metadate(self, bucket, key):
```
ddohler:
```suggestion
    def get_s3_object_metadata(self, bucket, key):
```

src/backend/users/migrations/0004_add_location_to_destination.py
```diff
@@ -0,0 +1,24 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='destination',
+            name='address',
+        ),
+        migrations.AddField(
+            model_name='destination',
+            name='location',
+            field=django.contrib.gis.db.models.fields.PointField(help_text='The lat/lng point location of the destination', srid=4326),
```
ddohler: This migration failed for me because I already had Destinations that had been added to my development database. When the migration ran, those Destinations ended up with null locations, which violated the not-null constraint. There's not really a great solution here because we have no way of knowing what location the existing Destinations should have, but luckily this only impacts developers so I think it should be safe to simply use the coordinates for Boston:
```suggestion
            field=django.contrib.gis.db.models.fields.PointField(default=Point(-71.06, 42.36, srid=4326), help_text='The lat/lng point location of the destination', srid=4326),
```

(You'll also need to add `from django.contrib.gis.geos import Point` up top)

To test this you can migrate back to 0003, add a UserProfile with Destinations using the Django admin, and then attempt to migrate forward.

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+            reason += "NO PRIMARY; "
+
+        # when any of the required field is empty
+        # don't attach the corresponding destination to user profile
+        if has_location and has_purpose and has_primary:
+            lat = des[DestinationKeys.LOC][LocationKeys.POSITION][PositionKeys.LAT]
+            lon = des[DestinationKeys.LOC][LocationKeys.POSITION][PositionKeys.LON]
+            purpose = self.decode_purpose(des[DestinationKeys.PURPOSE])
+
+            destination = Destination(
```
ddohler: If you want to create and save in one step you can use `Model.objects.create()` instead of `obj = Model(); obj.save()`

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+            return False, reason
+        pos = loc[LocationKeys.POSITION]
+        if PositionKeys.LAT not in pos or PositionKeys.LON not in pos:
+            reason += "NO FULL COORDINATE PAIR; "
+            return False, reason
+
+        return True, reason
+
+    def decode_purpose(self, purpose):
+        if purpose.lower() == "friends":
```
ddohler: Not critical, but another way to set this up that I slightly prefer is something like this; it's slightly more compact but I don't think it sacrifices too much readability:

```python
PURPOSE_MAPPING = {
    "friends": Destination.TripPurpose.FRIENDS,
    "work": Destination.TripPurpose.WORK
    ...
}

def decode_purpose(self, purpose):
    return PURPOSE_MAPPING.get(purpose.lower(), Destination.TripPurpose.OTHER)
```

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+        if has_location and has_purpose and has_primary:
+            lat = des[DestinationKeys.LOC][LocationKeys.POSITION][PositionKeys.LAT]
+            lon = des[DestinationKeys.LOC][LocationKeys.POSITION][PositionKeys.LON]
+            purpose = self.decode_purpose(des[DestinationKeys.PURPOSE])
+
+            destination = Destination(
+                profile=user_profile,
+                location=Point(x=lat, y=lon),
+                purpose=purpose,
+                primary_destination=des[DestinationKeys.PRIMARY],
```
ddohler, line 392: Once we re-add the `address` field we'll need to update this so that it stores the `label` field into `address`. You could potentially also rename the Django field to be `label` if you coordinate with @ktohalloran on #497 .

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def save_user_record(self, bucket, key, profile):
+        user = None
+        has_email = self.is_key_val_nonempty(profile, ProfileKeys.EMAIL)
+
+        # only email field is really used to migrate user records
+        # headOfHousehold is used for names, but email is the final
+        # fallback if headOfHousehold is empty
+        if has_email:
+            first_name, last_name = self.get_first_last_names(profile)
```
ddohler, line 220: I don't think that `first_name` and `last_name` are required fields on `User`, and the existing Profile page only has a "Full Name" field, so I think it would be fine to skip these (unless you got errors leaving them out?).

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -0,0 +1,488 @@
...
+            last_modified_at = self.get_s3_object_metadate(bucket, key)[self.S3_OBJ_TIMESTAMP_KEY]
+            password = self.create_random_string()
```
ddohler, line 227: I'd add a comment here to note that users will not be using passwords to log in, and this is just to ensure that passwords are not blank and not all the same, but we don't expect to use it.

ddohler (changes requested):
This is looking really good! There are a couple small things that need to change here that I think may have been due to my poor communication of requirements (sorry!), but hopefully they should all be quick fixes!

src/backend/users/models.py
aaronxsu: Sounds good. So address is not readily available in user profile stored on S3. Do we want this field to be nullable? If not, I think we could reverse geo-code from lat,lon and use the address to populate.

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Oh, good point! This is actually very good to know and less verbose. I think the reason why I passed the result to `destination` is because I'd like to return it in the end. If the return is not `None`, then it won't be reported as a part of the summary in the end. What do you think?

ddohler: `Model.objects.create()` will return the created object so you should still be able to do `destination = Model.objects.create()` -- does that cover what you need? Sorry for the confusion!

src/backend/users/models.py
ddohler: There's a `label` field in each location that we should be able to use. If there are locations with lat/lon but no label then we can handle those however is easiest; making the field nullable sounds fine, I don't think it will be necessary to reverse-geocode them.

src/backend/users/migrations/0004_add_location_to_destination.py
aaronxsu: Ohh, oops, agreed. Lemme add it!

Update: addressed here: https://github.com/azavea/echo-locator/pull/478/commits/65d2a4f5074d2a03d551f6d00a96e6208465bcce

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Ah, I see. Alright.

Hey @ktohalloran , it looks like the linked PR uses `address` as the field in the `Destination` model, and reads in the value keyed under `label`. I am adding this field back in here, and am open to either names. I think `label` would be more consistent, and may better reflect what I have seen in the user profiles on S3, since those values don't look like full street addresses to me. So let's name this field `label`- what would you say?

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Hmmm, on my end, these two fields seem to be not nullable, or maybe I was looking at the wrong table...:

```
 Table "public.auth_user"
    Column    |           Type           | Collation | Nullable |                Default                
--------------+--------------------------+-----------+----------+---------------------------------------
 id           | integer                  |           | not null | nextval('auth_user_id_seq'::regclass)
 password     | character varying(128)   |           | not null | 
 last_login   | timestamp with time zone |           |          | 
 is_superuser | boolean                  |           | not null | 
 username     | character varying(150)   |           | not null | 
 first_name   | character varying(150)   |           | not null | 
 last_name    | character varying(150)   |           | not null | 
 email        | character varying(254)   |           | not null | 
 is_staff     | boolean                  |           | not null | 
 is_active    | boolean                  |           | not null | 
 date_joined  | timestamp with time zone |           | not null | 

```

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Addressed here: https://github.com/azavea/echo-locator/pull/478/commits/fb420c3cad82fa7a46b28f8bee9e46b40641fdd9

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Addressed here: https://github.com/azavea/echo-locator/pull/478/commits/fb420c3cad82fa7a46b28f8bee9e46b40641fdd9

src/backend/users/models.py
aaronxsu: Makes sense. I plan to handle this field after the discussion here: https://github.com/azavea/echo-locator/pull/478#discussion_r871839086, re: to decide on a field name

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: No worries! Addressed here: https://github.com/azavea/echo-locator/pull/478/commits/a36a8f148ac0ae9312f701d016466514c1a7c864

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Makes sense. Addressed here: https://github.com/azavea/echo-locator/pull/478/commits/a36a8f148ac0ae9312f701d016466514c1a7c864

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Good point. Addressed here: https://github.com/azavea/echo-locator/pull/478/commits/a36a8f148ac0ae9312f701d016466514c1a7c864

src/backend/users/management/commands/migrate_housing_seeker_data.py
ktohalloran: That sounds good to me! I'll make updates to my PR to reflect the switch once this PR is merged.

src/backend/users/management/commands/migrate_housing_seeker_data.py
ddohler: Ah, right, sorry -- by convention [Django uses empty string (`''`) as the null value for CharFields](https://docs.djangoproject.com/en/4.0/ref/models/fields/#null), so at the database level they're still not-nullable, but in application code they can be omitted and Django will fill the fields with empty-string before saving to the DB. So you should be able to simplify the code somewhat by leaving these out, but it also doesn't hurt anything, so it's fine to leave as-is if you'd prefer!

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: @ktohalloran that sounds perfect to me. Thanks!

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Ah, gotcha! I did a quick check on those user profiles, and they were populated fine using this `get_first_last_names ` method. Since we also implemented the profile editing page, it will be possible to edit if these names are not perfect I assume. So I think I will leave the name determination logic as is.

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Update: address this here: https://github.com/azavea/echo-locator/pull/478/commits/fcfafe622990d4fb1c326ba9c21cfafcd4b5b09a

aaronxsu:
@ddohler No worries. The suggestions are legit! I addressed them and linked the relevant commits. Please feel free to let me know if any changes needed.

ddohler (approved):
Works great! However, the field name switch from `address` -> `label` has broken a few things at the API level which it would be good to fix before merging so we don't temporarily break staging (any more than it already is due to the routing issue). I don't need to re-review after fixing these though, it should be a simple find-and-replace.

- [users/models.py](https://github.com/azavea/echo-locator/blob/develop/src/backend/users/models.py#L59)
- [users/views.py](https://github.com/azavea/echo-locator/blob/develop/src/backend/users/views.py#L82)
- [users/serializers.py](https://github.com/azavea/echo-locator/blob/develop/src/backend/users/serializers.py#L10)

PR #479, opened 2022-05-03
Change from latest to GIT_COMMIT

### LINKED ISSUE #476: Terraform doesn't always force the rebuild with the latest image like it should
This likely explains why it seemed to be working a while ago (and why it was really broken).

The other option is to just go with requiring the GIT_COMMIT workflow.  (The only real negative is it's harder to automatically clear old images)

Workaround: Manually stop/delete tasks/services in the cluster.

BryanQuigley:
This is a small change to stop relying on the latest docker image tag
and just have terraform use actual git commit shas.

Resolves: #476

### Demo
From ECS manager, showing django tasks

RUNNING 322268938932.dkr.ecr.us-east-1.amazonaws.com/echostgdjango:d95d488

INACTIVE RUNNING 322268938932.dkr.ecr.us-east-1.amazonaws.com/echostgdjango:5f23706

## Testing Instructions

 * See the Image SHA change like above

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -7,7 +7,7 @@ services:
       - $HOME/.aws:/root/.aws:ro
     environment:
       - AWS_PROFILE=echo-locator
-      - GIT_COMMIT=${GIT_COMMIT:-latest}
+      - GIT_COMMIT=${GIT_COMMIT}
       - AWS_ACCESS_KEY_ID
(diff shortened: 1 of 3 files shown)


ddohler (approved):
Looks good! I confirmed that the task is running with the commit from this PR:
Screen Shot 2022-05-03 at 13 52 03

Issue #465, opened 2022-04-27
Enable user signup


ddohler:
Because we're doing passwordless login, signing up will look pretty similar to signing in from a user's perspective, i.e. users will enter their email address and then receive an email with a signed link that will allow them to sign in to the app. However, handling will need to be different on the backend when the user doesn't already exist in the database. Tasks for this card:
- Create the necessary database objects when a user attempts to sign in who doesn't already exist in the system and send them an email
- Make adjustments to the frontend and/or email text to clarify to users that they are going through a signup process.

## Notes
- It will probably be helpful to be able to differentiate between users who have attempted to sign up but haven't actually clicked the email link, and users who have made it all the way to logging in to the app. This could be accomplished by not creating a `UserProfile` object until the user successfully logs in, or by storing one or more fields on the `UserProfile` (e.g. `is_verified` or `last_login_date`).

PR #482, opened 2022-05-05
Add password-less authentication flow for house seekers

### LINKED ISSUE #442: Set up housing seeker authentication (read-only)
Users should be able to:
- See a place to enter their email on the login screen (screenshot below)
- Receive a login email when they click it with an authorization link (this will just show up in the console for now)
- Clicking the link will take them back to the app and they'll be authenticated for 30 days.
- The app should reflect the user's profile.

Not included in this task:
- Editing the user's profile. The user should be able to see it but we don't need to add the update endpoints for now; we can just edit in the admin to test. If updates come along for the ride due to DRF built-ins including it, that's fine, but only the reading part needs to work correctly for this task.
- Signing up new users. We'll just use the admin to create users.
- Displaying error messages if the user isn't found. Because we're using passwordless sign-in, signing up and signing in are pretty similar from a user's perspective. Additionally, we don't want to let people discover whether someone is using the site just by entering their email address.
- Logging out.

In practice, I think this will require:
- Setting up an API endpoint to login which sends the email
- Figuring out how to generate the unique login link for the user and provide them with API credentials if they use a valid link (see discussion in #413 for some thoughts about this).
- Handling user authentication status on the frontend (where to store API credentials once acquired, and have the frontend state react appropriately to those credentials).
- Create an endpoint for user profile info and have the frontend use it.

Screen Shot 2022-04-07 at 14 38 25

ktohalloran:
## Overview

This PR gives house seekers the ability to log in to the app as an alternative to using it anonymously. It does so using a passwordless login flow in which the user enters their email and clicks sign in, receives an email with a sign in and link, and upon clicking it, is logged in and their profile loaded. The relationship between Django and React during this flow is illustrated below.

Django Magic Link Auth (2)

### Notes

**Fetch userProfile**: As seen in the above picture, this login system takes advantage of the existing `authenticator` component (previously named `anonymous-authenticator`) which renders either the sign in page or the application depending on whether a `userProfile` exists. On render, `authenticator` checks for the unique case in which the user has an `auth_token` cookie but not a user profile or an error `loginMessage` as generated on failed login (see more on that below). If this is the case, it then requests the user's profile from the backend and either saves the profile to state or shows the error message. I chose to put it in render here, as opposed to in `handleAuthChange`, because this part of the authentication needs to happen without interaction on the frontend, but rather as a response to the backend's redirect to the frontend. The check for `auth_token`, `userProfile`, and `loginMessage` should be enough to avoid infinite rendering.

**loginMessage:** As mentioned above, this PR adds a state-wide `loginMessage` that is displayed on the sign in page first following submission of an email, and potentially again if the user attempting to sign in does not have a `userProfile`. This should currently only be necessary when a Counselor or admin who has not made themselves an account attempts to log in, but may be useful when adding user sign up in issue #465. 

**Failed Test:** This PR uses Django Rest Framework's Token Authentication, which added a new `Token` table that changed the `pk`s of the entries in `auth.permissions` as used by `src/backend/users/fixtures/group_permissions.json` to set the Counselor group's permissions, resulting in a failed test. This PR includes an edit to that file to allow Counselors to view and change Destinations, as well as Users and their profiles, and updates the `pk`s in that fixture to reflect the new ones following the addition of `Token`.

 **Dummy Data:** In order to successfully bring each profile from the backend into the frontend, the data needed to match the `AccountProfile` type. As a result, `GetUserProfile` view includes a repackaging of data retrieved from the `UserProfile` and `Destination` tables to match AccountProfile. As described in PR [457](https://github.com/azavea/echo-locator/pull/457), Destination should but does not currently include longitude/latitude data. This PR uses dummy data when sending Destinations to the frontend, to be replaced with real data following the merge of 457.

**Why 8085?:** Try as I might, I could not get the `ObtainToken` view to work via port 9966. As a result, the magic link directs to 8085 and everything past that occurs on 8085 to avoid CSRF issues.

## Testing Instructions

 * Run `./scripts/update` to migrate the new `Token` table and to apply the new Counselor group permissions
 * Start the server and navigate to **localhost:9966**. 
 * First, check that signing in with an invalid email address fails silently. Enter an email that you _do not_ have an account associated with and verify 1) you still see the success message on the sign in page, 2) you do not see an "email" in your terminal, and 3) that you do not see any 500 errors in your console or network tab of the dev tools.
 * Next, check that signing in as a counselor/admin displays the correct error message. Enter an email for which you _do_ have a user but _not_ a user profile and click "Sign in".
 * You should see an "email" in your terminal. Copy and paste the link into your browser. It should display a message on the screen explaining that you do not have a profile.
 * Check that login works for users with an associated profile. Enter an email for which you have an associated user and userProfile and click "Sign in".
 * You should see an "email" in your terminal. Copy and paste the link into your browser. It should take you into the app. 
 * Verify that you see the account's addresses in the dropdown menu of the dock (as a reminder, the routing is currently using dummy lon/lat data for each address, but you should still be able to see the proper addresses and labels in the dock).
 * Finally, run `./scripts/test` and verify that the fixture fix worked as expected.

Resolves #442


src/backend/users/views.py
```diff
@@ -1,3 +1,102 @@
...
+            <p>Thanks for using ECHO! Here is your <a href="{}">link to login</a>. </p>
+            <p>BHA</p>
+            """.format(
+                login_link
+            )
+
+            send_mail(
+                "Your ECHO Login Link",
+                html_message,
+                "admin@domain.com",  # This will need to be replaced with actual values
```
ddohler: We should create an issue for this and restructure this comment as a `TODO` with a link to the issue. This will need to be parameterized based on environment (so that the emails come from `@stg.echosearch.org` in staging and from the production domain in production).

src/backend/users/views.py
```diff
@@ -1,3 +1,102 @@
...
 
-# Create your views here.
+from .serializers import UserSerializer
+
+
+class LoginPage(APIView):
+    def post(self, request, **kwargs):
+        email = request.data["email"]
+        # try to find the User, but for privacy reasons, do not send 500 error back if not found
+        try:
```
ddohler, line 18: We usually write this type of pattern more like the following:

```
try:
    user = User.objects.get(username=email)
    # other email-sending stuff
    ...
except ObjectDoesNotExist:
    pass
return Response(...)
```

It's a stylistic thing because it doesn't really save any lines, but I think it makes the intent slightly clearer because `pass` is a strong expression of "I don't care about this".

Separately, Django makes [model-specific not-found exceptions](https://docs.djangoproject.com/en/4.0/ref/models/class/#django.db.models.Model.DoesNotExist) available, so you can tighten down this `except` clause even further with `except User.ObjectDoesNotExist:`.

src/backend/users/views.py
```diff
@@ -1,3 +1,102 @@
...
+        formatted_destinations = [
+            {
+                "location": {
+                    "label": i["address"],
+                    "position": {"lat": 42.351550, "lon": -71.084753},
+                },
+                "primary": i["primary_destination"],
+                "purpose": map_purposes[i["purpose"]],
+            }
+            for i in user_profile["destinations"]
```
ddohler: Customarily, using a plain `i` as a loop variable is usually done when the loop variable denotes a simple counter ("repeat this action X times"). In cases where you're iterating over a list of things, giving the loop variable a more descriptive name is usually easier for future readers to parse. I'd suggest `prof` here given that this isn't a very dense loop, but you could go with `profile` or with `p` depending on how terse you prefer to be.

src/backend/users/views.py
```diff
@@ -1,3 +1,102 @@
...
+
+
+class LoginPage(APIView):
+    def post(self, request, **kwargs):
+        email = request.data["email"]
+        # try to find the User, but for privacy reasons, do not send 500 error back if not found
+        try:
+            user = User.objects.get(username=email)
+            login_token = utils.get_query_string(user)
+            login_link = "http://localhost:8085/api/auth/login/{}".format(login_token)
```
ddohler: This should [pull the host from the request](https://docs.djangoproject.com/en/4.0/ref/request-response/#django.http.HttpRequest.get_host), otherwise these links won't be valid on staging or production (because the staging domain, for example, is `stg.echosearch.org` rather than `localhost:8085`). The issue with 9966 being broken on the token-generation side makes it harder to handle that, but I think maybe we just need to access the login form via port `8085` for the time being -- see the discussion of that issue below.

ddohler: Additionally, this should use [`Django's reverse() method`](https://docs.djangoproject.com/en/4.0/ref/urlresolvers/#reverse) to construct the URL path from the view name.

src/backend/users/views.py
```diff
@@ -1,3 +1,102 @@
...
+
+
+class GetUserProfile(APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, **kwargs):
+        user = User.objects.get(username=request.user)
+        serializer = UserSerializer(user)
+        user_profile = serializer.data["userprofile"]
+        map_prefs_to_nums = {"NI": 1, "SI": 2, "I": 3, "VI": 4}
```
ddohler:
```suggestion
        map_priorities_to_nums = {"NI": 1, "SI": 2, "I": 3, "VI": 4}
```

Just to keep things consistent because this is called a "priority" scale in other places.

taui/src/components/custom-sign-in.js
```diff
@@ -25,12 +25,42 @@ class SignInHeader extends React.Component {
...
   render () {
     const { t } = this.props
     return (
       <div className='auth-screen'>
         <SignInHeaderTranslated />
         <div className='auth-screen__main auth-main'>
+          <form onSubmit={this.handleSubmit}>
+            <fieldset>
+              <label>
+                <p>Email address</p>
```
ddohler: We should wrap the text strings in this component in `t()` calls so that they're translatable now that we're building this component ourselves.

taui/src/components/custom-sign-in.js
```diff
@@ -25,12 +25,42 @@ class SignInHeader extends React.Component {
...
+    this.handleSubmit = this.handleSubmit.bind(this)
+  }
+
+  handleChange (e) {
+    this.setState({value: e.target.value})
+  }
+
+  handleSubmit (e) {
+    e.preventDefault()
+    this.props.setLoginMessage('Thank you! Please check your email for a link to sign into ECHO.')
```
ddohler: I'd suggest moving this into the `sendLoginLink` action; that way, we can wait to display the "Thank you!" message until we're certain the request has succeeded or failed.

taui/src/actions/profile.js
```diff
@@ -26,3 +28,37 @@ export const setProfile = (profile) => (dispatch, getState) => {
...
+      'Authorization': `Token ${authToken}`
+    }
+  })
+    .then((response) => {
+      dispatch({type: 'set auth token', payload: authToken})
+      dispatch({type: 'set profile', payload: response.data})
+    })
+    .catch((error) => {
+      console.error('Error fetching user profile', error)
+      dispatch({type: 'set login message', payload: 'You do not have a user profile. Please create one and try again.'})
```
ddohler: These strings should also be made translatable.

src/backend/users/views.py
```diff
@@ -1,3 +1,102 @@
...
+            "FA": "Family",
+            "FR": "Friends",
+            "WP": "Worship",
+            "DR": "Doctor",
+            "OT": "Other",
+        }
+
+        # repackage destinations to match frontend AccountAddress
+        # currently uses dummy lat & lon prior to addition of Point in PR 478
+        # https://github.com/azavea/echo-locator/pull/478/
```
ddohler: Let's make this a TODO and create an issue to update this once #478 goes in.

taui/src/index.js
```diff
@@ -76,6 +76,7 @@ function mapStateToProps (state, ownProps) {
     activeListing: get(state, 'data.activeListing'),
     activeNetworkIndex: select.activeNetworkIndex(state, ownProps),
     activeNeighborhoodBounds: select.activeNeighborhoodBounds(state, ownProps),
+    authToken: select.authToken(state, ownProps),
```
ddohler: These selectors are so simple that I'd suggest just inlining the `get()` call right here, like is done on line 76, above:
```suggestion
    authToken: get(state, 'authToken'),
```

That should then let us get rid of the separate files for these selectors.

src/backend/users/views.py
ddohler: Also, something very strange is going on when trying to hit this endpoint via port `9966`.

Here are the logs I get when I access this via `curl`:
```
django_1    | [2022-05-05 20:45:36 +0000] [7] [DEBUG] GET /api/auth/login/
django_1    | 172.28.0.3 - - [05/May/2022:20:45:36 +0000] "GET /api/auth/login/?echo_auth=AAAAAwRoTvOSkbCDhf8E9VyW HTTP/1.1" 302 0 "-" "curl/7.82.0" <-- Successfully returns a 302
django_1    | [2022-05-05 20:45:36 +0000] [7] [DEBUG] GET /
django_1    | 172.28.0.3 - - [05/May/2022:20:45:36 +0000] "GET / HTTP/1.1" 200 0 "http://django:8085/api/auth/login/?echo_auth=AAAAAwRoTvOSkbCDhf8E9VyW" "curl/7.82.0"
```

And here's what curl sees:
```
$ curl -v http://localhost:9966/api/auth/login/?echo_auth=AAAAAwRoTvOSkbCDhf8E9VyW
*   Trying 127.0.0.1:9966...
* Connected to localhost (127.0.0.1) port 9966 (#0)
> GET /api/auth/login/?echo_auth=AAAAAwRoTvOSkbCDhf8E9VyW HTTP/1.1
> Host: localhost:9966
> User-Agent: curl/7.82.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: Cache-Control, Pragma, Origin, Authorization, Content-Type, X-Requested-With
< Access-Control-Allow-Methods: GET, PUT, POST
< access-control-allow-origin: *
< server: gunicorn
< date: Thu, 05 May 2022 20:49:47 GMT
< connection: close
< content-type: text/html; charset="utf-8"
< cache-control: max-age=0, public
< last-modified: Fri, 29 Apr 2022 18:35:02 GMT
< etag: "626c2fd6-2c8"
< content-length: 712
< x-content-type-options: nosniff
< referrer-policy: same-origin
< 
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0">
    <link href="https://cdnjs.cloudflare.com/ajax/libs/normalize/8.0.0/normalize.min.css" rel="stylesheet">
    <link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
    <link href="https://api.mapbox.com/mapbox.js/v3.1.1/mapbox.css" rel="stylesheet">
    <link rel="shortcut icon" href="https://www.bostonhousing.org/favicon.ico" type="image/x-icon">
  </head>
  <body>
    <div id="root" class="app-root"></div>
    <script src="assets/index.js"></script>
  </body>
</html>
* Closing connection 0
```

That looks like Django received at least 3 requests for what should have been only one, and _something_ spookily followed the redirect without ever letting `curl` see it (`curl` doesn't follow redirects via default, so if the redirect had been returned to `curl` it would have just quit at that point).

My best guess for what that spooky something might be is mastarm. When we access Django via port `9966` we're getting routed through mastarm's proxy feature, and the proxy misbehaving could certainly explain Django seeing mysterious requests that don't seem to come from anywhere 👻 . What it looks like may be happening is that the "inside" of the mastarm proxy is following the redirect _on its own_, rather than passing it back to the client that made the initial request. I don't think there's any reason to dig in further given that mastarm will be removed soon, so let's open an issue to re-investigate this once #457 is merged and call it good for the time being.

ddohler (approved):
This was a great PR to review -- the diagram made it really easy to understand what to expect, the notes and description added helpful context, and the code is well organized and clean. I also tried a few different ways of confusing the logic (like making the token expire after only a minute) but it was rock solid. Bravo! 👏 

There is one comment that I do think needs to be changed before this gets merged (relating to how we generate the host in the magic link URLs), but that should be a pretty simple fix. The rest of the comments are very minor, so I don't need to re-review before merging unless you want me to. There are a number of follow-on issues that need to get opened for this; if any of the comments beside the host thing turn out to be not-so-simple to implement, feel free to make issues and punt on those as well.

PR #484, opened 2022-05-06
Fix static assets on staging

### LINKED ISSUE #471: Static files should be served on staging
Discovered as part of #459 , we expect to serve our static files and the API on the same host, but in the staging env we receive a 404 on `/` endpoint and `whitenoise` errors within the task logs. This was addressed in #424, but only works locally. Additionally, when `docker exec` into the django instance there is no `/static` dir as expected. In the github action logs `collectstatic` is running correctly so it's possible volume mounting the `taui` assets into `/static` is not as expected.

rachelekm:
## Overview

This allows for the frontend static assets bundled by mastarm to be available on staging as part of the django container image build.

### Notes
- Ideally there should be one fix for both deployed and local environments. I found that after changing the context to include `/taui`, copying assets via the Dockerfile, and taking the volume mounts out of the docker-compose yaml, it worked on deployed environments but no longer locally. I could make larger changes so that the assets could be built/bundled before the django container build in `scripts/update` (so that it better mirrored `scripts/cibuild`), but knowing this will eventually need to be rebased and altered with the new bundler PR #457, I opted to leave the volume/bind mounting in for local dev for now since it will be easier to take out.
- On testing staging, the address field in user profile is not functioning. It looks like the access token we use for mapbox geocoding the address field in user profile is undefined and now we can't submit destinations. This has been addressed in issue #487

## Testing Instructions

 * Delete any already built assets and their directory folder from your local environment: `taui/assets` and `src/backend/static`
 * _optional_ Run `docker system prune`
 * Run` scripts/update`
 * Run `scripts/server` and test static files locally: http://localhost:8085/
 * Navigate to https://stg.echosearch.org/ and confirm frontend displays and you can reach https://stg.echosearch.org/profile
 
Resolves #471, #459

### DIFF
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -26,6 +26,11 @@ if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
         REACT_APP_GIT_COMMIT="${GIT_COMMIT}" \
             ./scripts/update
 
+        docker-compose \
+            -f docker-compose.yml \
+            run --rm --no-deps --entrypoint "bash -c" taui \
+            "yarn install && yarn build"
+
         # Build tagged container images
         GIT_COMMIT="${GIT_COMMIT}" docker-compose \
             -f docker-compose.yml \
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -26,6 +26,12 @@ if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
         REACT_APP_GIT_COMMIT="${GIT_COMMIT}" \
             ./scripts/update
 
+        # Build static asset bundle
+        docker-compose \
+            -f docker-compose.yml \
+            run --rm --no-deps --entrypoint "bash -c" taui \
+            "yarn install && yarn build"
+
         # Build tagged container images
         GIT_COMMIT="${GIT_COMMIT}" docker-compose \
             -f docker-compose.yml \
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -26,12 +26,6 @@ if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
         REACT_APP_GIT_COMMIT="${GIT_COMMIT}" \
             ./scripts/update
 
-        # Build static asset bundle
-        docker-compose \
-            -f docker-compose.yml \
-            run --rm --no-deps --entrypoint "bash -c" taui \
-            "yarn install && yarn build"
-
         # Build tagged container images
         GIT_COMMIT="${GIT_COMMIT}" docker-compose \
             -f docker-compose.yml \
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -26,11 +26,6 @@ if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
         REACT_APP_GIT_COMMIT="${GIT_COMMIT}" \
             ./scripts/update
 
-        docker-compose \
-            -f docker-compose.yml \
-            run --rm --no-deps --entrypoint "bash -c" taui \
-            "yarn install && yarn build"
-
         # Build tagged container images
         GIT_COMMIT="${GIT_COMMIT}" docker-compose \
             -f docker-compose.yml \
--- a/scripts/update
+++ b/scripts/update
@@ -26,6 +26,12 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Ensure container images are current
         docker-compose build
 
+        # Build static asset bundle
+        docker-compose \
+            -f docker-compose.yml \
+            run --rm --no-deps --entrypoint "bash -c" taui \
+            "yarn install && yarn build"
+
         # Bring up PostgreSQL and Django in a way that respects
         # configured service health checks.
         docker-compose up -d database django
--- a/scripts/update
+++ b/scripts/update
@@ -24,8 +24,14 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         usage
     else
         # Ensure container images are current
-        # Separated to build taui static assets before copied to django container
         docker-compose build taui
+
+        # Separated to build taui static assets before copied to django container
+        docker-compose \
+            -f docker-compose.yml \
+            run --rm --no-deps --entrypoint "bash -c" taui \
+            "yarn install && yarn build"
+
         docker-compose build database django
 
         # Bring up PostgreSQL and Django in a way that respects
--- a/scripts/update
+++ b/scripts/update
@@ -24,13 +24,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         usage
     else
         # Ensure container images are current
-        docker-compose build
-
-        # Build static asset bundle
-        docker-compose \
-            -f docker-compose.yml \
-            run --rm --no-deps --entrypoint "bash -c" taui \
-            "yarn install && yarn build"
+        # Separated to build taui static assets before copied to django container
+        docker-compose build taui
+        docker-compose build database django
 
         # Bring up PostgreSQL and Django in a way that respects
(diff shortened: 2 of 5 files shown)


src/backend/Dockerfile
```diff
@@ -3,10 +3,13 @@ FROM quay.io/azavea/django:3.2-python3.10-slim
 RUN mkdir -p /usr/local/src/backend
 WORKDIR /usr/local/src/backend
 
-COPY requirements.txt /usr/local/src/backend/
+COPY /src/backend/requirements.txt /usr/local/src/backend/
```
ddohler:
```suggestion
COPY ./src/backend/requirements.txt .
```

I think it's possible to use `.` for the destination because we've specified `WORKDIR /usr/local/src/backend` up above.

scripts/update
```diff
@@ -26,6 +26,12 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Ensure container images are current
         docker-compose build
```
ddohler: This command fails on a fresh build (when `taui/assets` is empty) because it tries to copy from `taui/assets` into the Django container, but fails because the folder is empty. I was able to resolve the issue by running the `yarn build` command below and then re-running `scripts/update`, so I think either we need to have two separate `docker-compose build` commands so that `docker-compose build taui` comes first and then `docker-compose build database django` comes after line 34, or we need to execute `yarn build` as part of the taui Dockerfile so that the static assets exist by the time the Django image is being built.

src/backend/Dockerfile
```diff
@@ -3,10 +3,13 @@ FROM quay.io/azavea/django:3.2-python3.10-slim
...
 RUN pip install --no-cache-dir -r requirements.txt
 
-COPY . /usr/local/src/backend
+COPY ./src/backend /usr/local/src/backend
+
+COPY ./taui/assets*/ /usr/local/src/backend/static/assets
```
ddohler: Can this be just `./taui/assets/`?

scripts/cibuild
```diff
@@ -26,6 +26,12 @@ if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
         REACT_APP_GIT_COMMIT="${GIT_COMMIT}" \
             ./scripts/update
 
+        # Build static asset bundle
+        docker-compose \
+            -f docker-compose.yml \
+            run --rm --no-deps --entrypoint "bash -c" taui \
+            "yarn install && yarn build"
```
ddohler: Is this necessary if it's also happening in `scripts/update`?

ddohler (changes requested):
Following the testing instructions didn't quite work for me because the Django Dockerfile assumes that the `taui/assets/` folder isn't empty, but I think that should be relatively straightforward to fix.

src/backend/Dockerfile
rachelekm: This worked! Will edit.

scripts/cibuild
rachelekm: Ah, nope! Will take this out and leave the `yarn install` and `build` in `scripts/update` to cover asset bundling for both local and deployed setup.

scripts/update
rachelekm: We talked about this in a separate discussion, but I set this up under the assumption that the directories `taui/assets/ `and `src/backend/static/` themselves would not exist at the start of a fresh build—which is not always true, as evidenced here! I implemented the two separate builds as suggested so that regardless of a user's local environment the assets are made available to the Django container at volume mount regardless of whether the directory originally existed, didn't exist, or is empty.

src/backend/Dockerfile
rachelekm: I tested this out and it does work without the asterisk now that we include `yarn build` as part of the `taui` docker build! But, I also had a scenario where I got rid of the `taui/assets` directory, ran `scripts/update`, and it skipped over the build since it was already cached and then failed because there was nothing there. I can't imagine this happening very often if ever in a local dev, but it offers a little extra resiliency when there may not be a `taui/assets` so I decided to leave it. Let me know if you think it will be cleaner to just take out!

scripts/cibuild
rachelekm: Alas, I had to bring this back in. I thought I could still leave this out since `yarn build` was now included in the `taui` docker build, but it didn't build and bundle the assets as expected.

rachelekm:
Thanks so much for your edits and help better understanding the this build process. I implemented some of the changes you suggested/added some context in the replies from our separate discussion and hoping it should be more resilient now.

src/backend/Dockerfile
ddohler: Hmm, so interestingly enough I can't get the build to work (at all) with the asterisk there. Even if `taui/assets/` exists and has files, I always get
```
ddohler@lima:~/Development/projects/echo-locator$ ls taui/assets/
config.json      echo_combined_logo_fullcolor.svg      echo_combined_logo_STACKED_fullcolor.svg      index.js      neighborhood_bounds.json      neighborhoods.json
config.json.map  echo_combined_logo_fullcolor.svg.map  echo_combined_logo_STACKED_fullcolor.svg.map  index.js.map  neighborhood_bounds.json.map  neighborhoods.json.map
ddohler@lima:~/Development/projects/echo-locator$ docker-compose build django
Building django
Sending build context to Docker daemon    1.2GB
Step 1/9 : FROM quay.io/azavea/django:3.2-python3.10-slim
 ---> d158d6f22ceb
Step 2/9 : RUN mkdir -p /usr/local/src/backend
 ---> Using cache
 ---> 77de371f9bb5
Step 3/9 : WORKDIR /usr/local/src/backend
 ---> Using cache
 ---> 7dbd4a9ca59c
Step 4/9 : COPY ./src/backend/requirements.txt .
 ---> Using cache
 ---> 005fb203af59
Step 5/9 : RUN pip install --no-cache-dir -r requirements.txt
 ---> Using cache
 ---> d60532b194f2
Step 6/9 : COPY ./src/backend /usr/local/src/backend
 ---> Using cache
 ---> 42cc518b80e4
Step 7/9 : COPY ./taui/assets*/ /usr/local/src/backend/static/assets
COPY failed: no source files were specified
ERROR: Service 'django' failed to build : Build failed
```

scripts/cibuild
ddohler: Ah okay, I think that having this in the `taui` Dockerfile vs. in `scripts/update` itself may be the source of the problem -- putting it in the Dockerfile means it'll be subject to Docker's rules about when to re-run a build step or when to use the cached layer, so sometimes it may get skipped. But since we're relying on this step to run in order to populate files on the host file system so that they can make it into another image, I think we want it to always run, no matter what. The only way to accomplish that is to put it in `scripts/update` itself, between the two `docker-compose build` calls, rather than in `taui/Dockerfile`. I think doing that should cause `taui/assets` to be filled with files in all cases when `scripts/update` is run, so then it should be possible to remove it from `scripts/cibuild`.

rachelekm: OH! Thank you for clarifying, that's really helpful. Ok I'll take that back out of the `Dockerfile` and into `scripts/update` and hopefully that does the trick!

rachelekm: This worked for me! This is back to being out of `scripts/cibuild` 🎉

src/backend/Dockerfile
rachelekm: I wasn't able to replicate this on my end _but_, as you suggested below, I think we can ditch the asterisk now that the `taui` `yarn build` command is in the Dockerfile vs directly in `scripts/update`. After making the change I was able to get this to succeed without the asterisk, so it's officially out and hopefully it doesn't cause trouble!

ddohler (approved):
Looks good! I didn't test on staging but building locally works successfully for me and I can open up the Django image without a volume mount and confirm that the taui assets show up inside.

PR #490, opened 2022-05-08
Inline VPC to drop bastion

### LINKED ISSUE #449: Bastion alternatives (or finish bastion)

BryanQuigley:
Adds about 130 lines to network but may be able to reduce that further.
Drops dashboard as it is mostly covered by automatic dashboards
A few other changes were due to terraform fmt.

Resolves: #449

### Notes
This still uses NAT gateways and should be equivalent to the VPC module without the bastion.

### Demo

No bastion deployed and staging works as before.

## Testing Instructions

 * Confirm staging worked just as well as before.


ddohler (approved):
I confirmed that I can tunnel in and run management commends. I'm not really a terraform expert, but I read through that code and it looks good to me.

PR #497, opened 2022-05-11
Enable user profile editing

### LINKED ISSUE #443: Enable user profile editing
Users should be able to use the Profile page to edit their preferences and have those preferences persisted to the database. This may require creating an API endpoint and will require wiring up the frontend to use it when the user saves.

ktohalloran:
## Overview

This PR allows users to edit their profiles from the frontend, rather than only in the admin. It adds a `saveProfile` action which sends a `PUT` request to the `UserProfileView` (previously named `GetUserProfile`) on save to update existing profiles and destinations and add new destinations. It also makes a few small tweaks, such as changing "Day care" (as it was saved in the backend) to "Daycare" (as it was used in the frontend) to fix potential key errors and adds the user's email (as read-only) to the `EditProfile` component.

### Notes

While I found that I could use the `UserProfileView` to update destinations and profiles and create new destinations, deleting destinations via that endpoint was harder and messier. So I've added a separate action and endpoint for deleting destinations that is triggered directly by the delete button next to each destination, as opposed to the profile/destination updates, which are sent to the backend all at once on profile save.

The hardest part of this card was state management, especially between the `EditProfile` component state, the app state, and the database. I've tried to minimize the places where these 3 can diverge from each other by only updating database state on profile save or destination delete. Initially, I had planned to let the existing `handleAuthChange` flow handle updating `userProfile` in app state; however, I changed it to update inside the `saveProfile` action after a successful database update instead in order to prevent app and database state from diverging. I also added an error message so that when updating the database fails, it at least notifies the user rather than just not saving their changes; I just couldn't figure out how to make that messaging genuinely helpful and say something other than "try again."

## Testing Instructions

 * Run `./scripts/update` to make the new "day care" to "daycare" migration, then start the server.
 * Login and select "Edit Profile". Verify that your profile looks right based on what's in the backend.
 * Delete a destination and verify that the destination no longer appears on your profile in the backend.
 * Edit one your existing destinations and add a new one; check that both appear as they should in the backend.
 * Make edits to the rest of your profile and verify that they show up as expected in the backend.

Resolves #443


ktohalloran:
Seems the error messaging I added on profile save failure is no longer working as expected. Troubleshooting now.

ktohalloran:
Ok, my initial method of adding an error message on profile save failure isn't working and I can't think of another way to do it without creating an app-wide prop like `loginMessage`. @ddohler do you think I should try adding that? Or do you have any other recommendations on how to handle error messaging int his case? I don't think failing silently is a great way to handle it, but we have validation checks in the frontend so in theory the only times it should fail are if the database is offline or some other fluke an error message wouldn't help with, anyway.

src/backend/users/views.py
```diff
@@ -105,4 +109,76 @@ def get(self, request, **kwargs):
...
+        serializer = UserSerializer(User.objects.get(username=request.user))
+        content = self.repackage_for_frontend(serializer.data)
+        return Response(content)
+
+
+class DeleteDestinationView(APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def delete(self, request, *args, **kwargs):
+        deleted_destination = Destination.objects.get(id=request.data["destination"]["id"])
```
ddohler: This view may not stick around depending on how you approach the comments on the UserProfile view, but regardless, we shouldn't allow authenticated users to delete other users' destinations, so if we keep this, we should make the `get()` call more like `get(id=..., profile__user=request.user)` (not tested, but it should be close). That will restrict users to deleting only their own destinations. It would also be good to handle the case where the `id` doesn't exist, in which case I think we can just do nothing.

src/backend/users/views.py
```diff
@@ -105,4 +109,76 @@ def get(self, request, **kwargs):
...
+        content = self.repackage_for_frontend(serializer.data)
+
         return Response(content)
+
+    def put(self, request, *args, **kwargs):
+        data = request.data
+        updated_profile = UserProfile.objects.get(user=User.objects.get(username=request.user))
+
+        # save new/changed destinations
+        for destination in data["destinations"]:
```
ddohler: Okay, yeah, this is a bit of a tricky logic spot because you need to:
- Create Destinations that don't exist
- Update Destinations that do exist
- Delete Destinations that have been deleted (but it's not easy to tell).

The structure you've chosen is to handle deletions separately so that it's clear when they should happen, but one problem I see with that is that it makes the "Cancel" button not work as well as it used to, because there's no way to get a Destination back once it has been deleted.

I'd like to propose a couple of alternative structures that you could consider for this logic; let me know if you think one of these would be preferable:

1. Delete and re-create every Destination on every PUT. This would look something like (without handling the map_purposes stuff):
```python
Destination.objects.filter(user_profile=updated_profile).delete()
destinations = [Destination(profile=updated_profile, **dest) for dest in data["destinations"]]
Destination.objects.bulk_create(destinations)
```

2. Instead of not returning a Destination that you want to delete, mark it with a special flag or include a list of Destinations that you want to delete. For example, you might include a field in the payload that looked something like `destinations_to_delete: [1, 5]`. Or, you could add a temporary field to each destination, e.g. `should_delete: false`. Then, your code could do something like this:
```python
Destination.objects.filter(id__in=data["destinations_to_delete"]).delete()
```
or
```python
Destination.objects.filter(id__in=[d["id"] for d in data["destinations"] if d["should_delete"]).delete()
```

In terms of drawbacks / benefits to each one, the first option is very terse, and simplifies the update logic too, which is nice. It should also remove a bit of complexity on the frontend because it no longer has to worry about the `id` field because it'll be new each time. However, it'll make the update queries a bit heavier, so it'll place more load on the database. I'm not too worried about the database load issue because the number of users is expected to remain low and we're only ever doing 3 destinations at a time.

The second option would require fewer alterations to what you've currently got, however, and would have lower database load because it would only update what was needed.

What do you think?

src/backend/users/views.py
```diff
@@ -105,4 +109,76 @@ def get(self, request, **kwargs):
...
+            try:
+                updated_destination = Destination.objects.get(id=destination["id"])
```
ddohler: It looks like the frontend is specifying `id=0` for newly-created Destinations. I think a better approach would be to leave off the `id` field entirely for new Destinations. That's the convention that Django uses, so then you should be able to do something more like:

```python
dest = Destination(**destination) # Ignoring map_purposes for the sake of example
dest.save() # Will update if `id` exists, or create if it doesn't.
```

That should allow you to cut this `try`/`except` roughly in half because everything will collapse down to one case (though obviously none of this will be relevant if you choose to take the "delete everything and recreate it" approach).

src/backend/users/views.py
```diff
@@ -105,4 +109,76 @@ def get(self, request, **kwargs):
...
+
+    def get(self, request, **kwargs):
+        user = User.objects.get(username=request.user)
+        serializer = UserSerializer(user)
+
+        content = self.repackage_for_frontend(serializer.data)
+
         return Response(content)
+
+    def put(self, request, *args, **kwargs):
```
ddohler, line 123: For operations that potentially affect multiple objects it's good practice to use [`@transaction.atomic`](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#django.db.transaction.atomic) so that if something fails halfway through all the changes will either succeed or fail as one.

src/backend/users/views.py
```diff
@@ -105,4 +109,76 @@ def get(self, request, **kwargs):
...
+        data = request.data
+        updated_profile = UserProfile.objects.get(user=User.objects.get(username=request.user))
```
ddohler: For this pattern of
1. Grab a thing
2. Do some processing with the thing
3. Save the results back to the thing

it's best to use [`select_for_update`](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#select-for-update) -- that way, if multiple requests altering the same object come in at nearly the same time, they won't get into race conditions with each other.

In this particular case I actually don't think it's super important because the usage will be low (which makes a race condition less likely to occur) and, more importantly, this endpoint doesn't use any of the fields on the existing profile, it just writes to them with whatever's in the payload, so even if there is a "race" it'll still do the right thing in the end. But `select_for_update()` is a good habit to establish and it won't hurt anything.

src/backend/users/views.py
```diff
@@ -105,4 +109,76 @@ def get(self, request, **kwargs):
...
+                    address=destination["location"]["label"],
+                    primary_destination=destination["primary"],
+                    purpose=list(self.map_purposes.keys())[
+                        list(self.map_purposes.values()).index(destination["purpose"])
+                    ],
+                )
+                new_destination.save()
+
+        # determine user's mode of travel
+        if data["hasVehicle"]:
```
ddohler, line 146: I'm assuming this field manipulation (here through line 170) is in the view rather than in a serializer for the same reason that the `get` method has some parts that aren't in a serializer. One strategy that can be helpful in this type of situation is to use two different serializers, one for reading and one for writing. I don't think we should spend a lot of extra time on this now that it's working and there are other comments that are more pressing, but for future reference, that's a technique that I've found to work in the past.

taui/src/components/edit-profile.js
```diff
@@ -175,27 +178,32 @@ class EditProfile extends PureComponent<Props> {
...
+    } else {
+      this.props.handleAuthChange(profile)
+    }
     this.props.history.push('/map')
   }
 
   addAddress () {
     const destinations = this.state.destinations.slice()
     const newAddress: AccountAddress = {
+      id: 0,
```
ddohler: As noted above, I think it would be best to leave this off to better match Django's convention for identifying what's a new object.

taui/src/components/edit-profile.js
```diff
@@ -175,27 +178,32 @@ class EditProfile extends PureComponent<Props> {
...
       this.setState({errorMessage: t('Profile.AddressMissing')})
       return
     } else {
       this.setState({errorMessage: ''})
     }
 
-    this.props.handleAuthChange(profile)
+    if (!this.state.isAnonymous) {
+      if (!this.props.saveProfile(profile, this.props.data.authToken)) {
+        this.setState({errorMessage: t('Profile.SaveError')})
```
ddohler: This isn't working quite right for me -- I always get an error message, even if the profile does in fact save correctly. I think this is because `saveProfile()` always returns `undefined`, which will evaluate to `false` in the `if()` statement and cause this error message to get shown.

I think there are two potential paths forward:
1. Handle `setState()` and `this.props.history.push('/map')` within `saveProfile()` itself.
2. Have `saveProfile()` return a Promise and then switch this to
```javascript
this.props.saveProfile(profile, this.props.data.authToken)
  .then(/* handle success */)
  .catch(/* handle failure */)
```

taui/src/actions/profile.js
```diff
@@ -29,6 +29,35 @@ export const setProfile = (profile) => (dispatch, getState) => {
   }
 }
 
+export const saveProfile = (profile, authToken) => (dispatch, getState) => {
+  axios.put(`/api/user/`, profile, {
```
ddohler: To expand on the earlier comment about this not returning anything, this function can be summarized as:
```javascript
saveProfile = (profile, authToken) => (dispatch, getState) => {
  axios.put().then().catch()
}
```
Which shows that it's not returning anything because it only contains one expression, and that expression isn't returned. The `return` inside the `catch()` function will only affect the value of the Promise that is returned by `catch()` -- [`catch()` always returns a Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch), and the return value of any function that has been passed to `catch()` as the `onRejected` parameter is used to resolve the Promise.

ddohler (changes requested):
This is a good start -- I left some suggestions that I think could simplify the logic and potentially make things easier to reason about. I left one comment to address an issue I was seeing with the error message, but I'm not sure if that's the issue you were asking about. If that doesn't give you a path forward let me know and we can discuss it further!

src/backend/users/views.py
ktohalloran: Thanks for reviewing! I'm inclined towards the first option, especially because we're not expecting a lot of activity. It just feels cleaner and easy to read.

ktohalloran:
Thanks for reviewing this! I've made the changes to the `UserProfileView` to just delete and replace all destinations instead of using a separate endpoint and have also fixed the error messaging on profile save failure. I have also rebased onto `develop`, so the latest commits include necessary fixes for the bundler and management command commits. I did not feel that I had the time to reorganize the existing serializers into write and read serializers, especially due to the time I spent on rebasing, but please let me know if you'd like me to fix that before merging.

src/backend/users/views.py
```diff
@@ -105,4 +108,64 @@ def get(self, request, **kwargs):
...
+            updated_profile.travel_mode = "CA"
+        elif data["useCommuterRail"]:
+            updated_profile.travel_mode = "BTE"
+        else:
+            updated_profile.travel_mode = "BT"
+
+        updated_profile.full_name = data["headOfHousehold"]
+        updated_profile.commute_priority = list(self.map_priorities_to_nums.keys())[
```
ddohler: I totally see why you chose this path, because it feels wasteful to declare another variable when there's already one that has all the necessary data, but I think it would be preferable to declare a reverse mapping and use that. Doing it this way repeats the conversion logic three times, which creates openings for bugs to sneak in during future changes if someone messes up a copy-paste or forgets that this logic relies on the key ordering of `self.map_priorities_to_nums`.

It's possible to derive the reverse mapping from the forward mapping in order to maintain a single source for the data:
```
map_nums_to_priorities = {value: key for key, value in map_priorities_to_nums.items()}
...
updated_profile.commute_priority = self.map_nums_to_priorities[data["importanceAccessibility"]]
```

ddohler (approved):
Works great! I was able to update my profile and confirmed that if I changed the endpoint to return an error, I received an error message in the frontend. I had one small stylistic comment but this is good to go after that's resolved.

src/backend/users/views.py
ktohalloran: Thanks for illustrating how to do this--I'd never heard of reverse mapping before, so this was helpful! I've made this change and will merge now.

PR #498, opened 2022-05-11
Drop NAT Gateways

BryanQuigley:
Drop private routing as well
And unused aws_vpc_endpoint s3 (could see this being used in other
projects)

### Notes
I am somewhat surprised this appears to just work. My guess is the NAT gateways were actually used for the bastions primarily.
Internet access did not work from container before this - but if we need to make it work, we should just use an internet gateway (and add firewall rules if needed).

Low priority - we can do this change any time - it depends on nothing else AFAICT.

## Testing Instructions

 * Confirm everything works just as well as before.

### DIFF
--- a/deployment/terraform-django/network.tf
+++ b/deployment/terraform-django/network.tf
@@ -23,26 +23,6 @@ resource "aws_internet_gateway" "default" {
   }
 }
 
-resource "aws_route_table" "private" {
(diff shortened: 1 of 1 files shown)


BryanQuigley:
This needs more work as the app works fine, but it breaks ecr access...

PR #499, opened 2022-05-12
Implement user signup

### LINKED ISSUE #465: Enable user signup

rachelekm:
## Overview

This PR allows for users to sign up for an account with their email address and, on form submit, receive an emailed link to create a profile and sign in. Below the surface, when a user clicks the “create account” button: their email is validated, a new `User` is created, the new `User` is added to the `HouseSeeker` group, and an empty related `UserProfile` is created before sending the email. Appropriate frontend changes were made, including:
- Adding a clickable prompt on the “sign in” page to create an account (and from the “sign up” page a prompt to toggle back to “sign in”). It should be noted that these are not two separate pages/components, but just toggles between appropriate text/actions.
- Updated `loginMessage` payloads to walk through the signup process or give context to any signup errors.

### Notes

- One tricky part of this PR was figuring out how to let a new `User` create their initial `UserProfile`, even though the authorization to reach the `EditProfile` component relies on the presence of a `UserProfile`. The way I chose to get around this, while also differentiating `Users` who have attempted to sign up but haven't actually clicked the email link or who haven't create a `UserProfile`, was to ~~create an intermediary `AccountProfile` type: `{unverifiedUserProfile: boolean}`. Responding with this object from the backend and setting it as the profile seemed simpler than creating a field on the `UserProfile` model and allows `profile !== undefined` while also differentiating between a complete `UserProfile` and `isAnonymous`~~ create an empty `UserProfile` on signup. This allows `profile !== undefined` to pass authentication and will populate/view a default `EditProfile` component as if `isAnonymous` without actually being `isAnonymous`.
- I made a couple adjustments to the `AccountProfile` type as well as some of the conditionals in `EditProfile` to better fit when a new `User` _has_ a profile, but doesn't actually have anything in it to validate. This is namely getting rid of the `key` property since this is a leftover of our Amplify/S3 profile process (I couldn't find any active purpose to it) as well as making `voucherNumber` optional. These changes were also made as part of the open PR #497 so I tried to stay as close as possible to those changes to avoid any conflicts.
- Full account creation depends on the ability to edit a `UserProfile`. However, since there’s an open PR #497 for editing `UserProfile` from the frontend I decided not to touch any of the functionality. This PR simply creates an empty `UserProfile` for new `Users` and, once we have the ability to save edits to `UserProfile`, the absence of the `unverifiedUserProfile` property will signify account creation is "complete".
- I left the email body copy the same as login since it seemed broad enough to cover signup as well. Any extra signup context is included as part of the `loginMessage` after submitting email.

## Testing Instructions

 * `scripts/update` 
 * `scripts/manage test users.tests` to confirm tests function
 * `scripts/server`
 * Navigate to `localhost:9966/`
 * Click the "create an account" text to toggle to the signup form
 * Try and submit an invalid email and confirm account creation fails and error response message appears
 * Try and submit a valid email and confirm email link prints in the console
 * Try and submit the same email again and confirm duplicate account creation fails
 * Navigate to the email link and confirm it loads `/profile` page, and all fields are present but empty
 * Confirm you can't navigate beyond `/profile` without setting required fields
 * Create a new `User` through the admin, and don't create a related `UserProfile`. Login with the new `User` email and confirm still authorized to navigate to the `/profile` page.

Resolves #465

### DIFF
--- a/src/backend/users/serializers.py
+++ b/src/backend/users/serializers.py
@@ -1,4 +1,4 @@
-from django.contrib.auth.models import User
+from django.contrib.auth.models import Group, User
 from rest_framework import serializers
 
 from .models import Destination, UserProfile
@@ -37,3 +37,19 @@ class UserSerializer(serializers.ModelSerializer):
         model = User
         fields = ["userprofile", "username"]
         read_only_fields = ["username"]
+
+
+class HouseSeekerSignUpSerializer(serializers.ModelSerializer):
+    username = serializers.CharField(required=True, max_length=150)
+
+    class Meta:
+        model = User
+        fields = ["username"]
+
+    # Override to create User without UserProfile and add to HouseSeeker group
+    def create(self, validated_data):
+        user = User.objects.create(**validated_data)
+        houseseeker_group = Group.objects.get(name="HouseSeeker")
+        user.groups.add(houseseeker_group)
+        user.save()
+        return user
--- a/src/backend/users/serializers.py
+++ b/src/backend/users/serializers.py
@@ -40,7 +40,7 @@ class UserSerializer(serializers.ModelSerializer):
 
 
 class HouseSeekerSignUpSerializer(serializers.ModelSerializer):
-    username = serializers.CharField(required=True, max_length=150)
+    username = serializers.EmailField(required=True, max_length=150)
 
     class Meta:
         model = User
--- a/src/backend/users/serializers.py
+++ b/src/backend/users/serializers.py
@@ -46,10 +46,13 @@ class HouseSeekerSignUpSerializer(serializers.ModelSerializer):
         model = User
         fields = ["username"]
 
-    # Override to create User without UserProfile and add to HouseSeeker group
+    # Override to create User with empty UserProfile and add to HouseSeeker group
     def create(self, validated_data):
         user = User.objects.create(**validated_data)
         houseseeker_group = Group.objects.get(name="HouseSeeker")
         user.groups.add(houseseeker_group)
         user.save()
+
+        UserProfile.objects.create(user=user)
+
         return user
--- a/src/backend/users/serializers.py
+++ b/src/backend/users/serializers.py
@@ -1,4 +1,5 @@
 from django.contrib.auth.models import Group, User
+from django.db import transaction
 from rest_framework import serializers
 
 from .models import Destination, UserProfile
@@ -47,6 +48,7 @@ class HouseSeekerSignUpSerializer(serializers.ModelSerializer):
         fields = ["username"]
 
     # Override to create User with empty UserProfile and add to HouseSeeker group
+    @transaction.atomic
     def create(self, validated_data):
         user = User.objects.create(**validated_data)
         houseseeker_group = Group.objects.get(name="HouseSeeker")
--- a/src/backend/users/tests.py
+++ b/src/backend/users/tests.py
@@ -1,6 +1,8 @@
 from django.contrib.auth.models import Group, User
+from django.core import mail
 from django.test import Client, TestCase
 from users.models import UserProfile
+from users.serializers import HouseSeekerSignUpSerializer
 
 
 class AdminSiteTest(TestCase):
@@ -90,6 +92,49 @@ class AdminSiteTest(TestCase):
         )
 
 
+class HouseSeekerLoginTest(TestCase):
+    """Check User login process."""
+
+    fixtures = ["group_permissions"]
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.houseseeker = Client()
+        user_serializer = HouseSeekerSignUpSerializer(
+            data={"username": "testechouserexists@azavea.com"}
+        )
+        user_serializer.is_valid(raise_exception=True)
+        user_serializer.save()
+
+    def test_login_endpoint_response(self):
+        """
+        Test login endpoint responds with 200 for valid and invalid emails,
+        but only sends one login email for valid User.
+        """
+        first_response = self.houseseeker.post(
+            "/api/login/",
+            {"email": "testechouserexists@azavea.com"},
+            content_type="application/json",
+        )
+        self.assertEqual(
+            first_response.status_code,
+            200,
+            f"Expected 200, got {first_response.status_code}. {first_response.content}",
+        )
+        second_response = self.houseseeker.post(
+            "/api/login/",
+            {"username": "testechouserdoesntexist@azavea.com"},
+            content_type="application/json",
+        )
+        self.assertEqual(
+            second_response.status_code,
+            200,
+            f"Expected 200, got {second_response.status_code}. {second_response.content}",
+        )
+        num_emails = len(mail.outbox)
+        self.assertEqual(num_emails, 1, "Fail: expected one email to be sent, not %d." % num_emails)
+
+
 class HouseSeekerSignUpTest(TestCase):
     """Check User sign up process."""
 
@@ -138,3 +183,15 @@ class HouseSeekerSignUpTest(TestCase):
         self.assertEqual(houseseeker_user.first().groups.first(), self.test_houseseeker_group)
         houseseeker_user_profile = UserProfile.objects.filter(user=houseseeker_user.first())
         self.assertIsNone(houseseeker_user_profile.first())
+
+    def test_signup_email_link(self):
+        """
+        Test email is sent on User sign up and link in email responds with 200.
+        """
+        self.houseseeker.post(
+            "/api/signup/",
+            {"username": "testechoemail1@azavea.com"},
+            content_type="application/json",
+        )
+        num_emails = len(mail.outbox)
+        self.assertEqual(num_emails, 1, "Fail: expected one email to be sent, not %d." % num_emails)
--- a/src/backend/users/tests.py
+++ b/src/backend/users/tests.py
@@ -172,7 +172,7 @@ class HouseSeekerSignUpTest(TestCase):
         self.assertContains(first_response, "created your account", status_code=200)
         self.assertContains(second_response, "already have an account", status_code=200)
 
-    def test_user_and_no_profile_created_on_signup(self):
+    def test_user_and_empty_profile_created_on_signup(self):
         self.houseseeker.post(
             "/api/signup/",
             {"username": "testechoemail1@azavea.com"},
@@ -182,7 +182,7 @@ class HouseSeekerSignUpTest(TestCase):
         self.assertEqual(len(houseseeker_user), 1)
         self.assertEqual(houseseeker_user.first().groups.first(), self.test_houseseeker_group)
         houseseeker_user_profile = UserProfile.objects.filter(user=houseseeker_user.first())
-        self.assertIsNone(houseseeker_user_profile.first())
+        self.assertEqual(len(houseseeker_user_profile), 1)
 
     def test_signup_email_link(self):
         """
--- a/src/backend/users/tests.py
+++ b/src/backend/users/tests.py
@@ -100,9 +100,7 @@ class HouseSeekerLoginTest(TestCase):
     @classmethod
     def setUpTestData(cls):
         cls.houseseeker = Client()
-        user_serializer = HouseSeekerSignUpSerializer(
-            data={"username": "exists@azavea.com"}
-        )
+        user_serializer = HouseSeekerSignUpSerializer(data={"username": "exists@azavea.com"})
         user_serializer.is_valid(raise_exception=True)
         user_serializer.save()
 
@@ -131,7 +129,9 @@ class HouseSeekerLoginTest(TestCase):
             200,
             f"Expected 200, got {second_response.status_code}. {second_response.content}",
         )
-        self.assertEqual(len(mail.outbox), 1, "Fail: expected one email to be sent, not %d." % len(mail.outbox))
+        self.assertEqual(
+            len(mail.outbox), 1, "Fail: expected one email to be sent, not %d." % len(mail.outbox)
+        )
 
 
 class HouseSeekerSignUpTest(TestCase):
@@ -203,4 +203,6 @@ class HouseSeekerSignUpTest(TestCase):
             {"username": "testechoemail1@azavea.com"},
             content_type="application/json",
         )
-        self.assertEqual(len(mail.outbox), 1, "Fail: expected one email to be sent, not %d." % len(mail.outbox))
+        self.assertEqual(
+            len(mail.outbox), 1, "Fail: expected one email to be sent, not %d." % len(mail.outbox)
+        )
--- a/src/backend/users/tests.py
+++ b/src/backend/users/tests.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.models import Group, User
 from django.test import Client, TestCase
+from users.models import UserProfile
 
 
 class AdminSiteTest(TestCase):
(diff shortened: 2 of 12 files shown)


src/backend/users/views.py
```diff
@@ -106,3 +124,22 @@ def get(self, request, **kwargs):
             "voucherNumber": user_profile["voucher_number"],
         }
         return Response(content)
+
+
+class SignUpPage(APIView):
+    def post(self, request, **kwargs):
+        signup_message = "Thank you! You'll receive an email shortly with a link to complete your account. Click the link to create your profile and get started with ECHO."
+        try:
+            user_serializer = HouseSeekerSignUpSerializer(data=request.data)
```
rachelekm, line 207:
```suggestion
        try:
            with transaction.atomic():
                user_serializer = HouseSeekerSignUpSerializer(data=request.data)
```

I'm realizing after the fact I don't have a good way to rollback any database transactions in the event something happens between User creation --> add to HouseSeekers --> UserProfile creation within the serializer. Does adding something like the above make sense here?

rachelekm, line 133:
```suggestion
        try:
            with transaction.atomic():
                user_serializer = HouseSeekerSignUpSerializer(data=request.data)
```

I'm realizing after the fact I don't have a good way to rollback any database transactions in the event something happens between User creation --> add to HouseSeekers --> UserProfile creation within the serializer. Does adding something like the above make sense here?

src/backend/users/serializers.py
```diff
@@ -37,3 +37,22 @@ class Meta:
...
+
+class HouseSeekerSignUpSerializer(serializers.ModelSerializer):
+    username = serializers.EmailField(required=True, max_length=150)
+
+    class Meta:
+        model = User
+        fields = ["username"]
+
+    # Override to create User with empty UserProfile and add to HouseSeeker group
+    def create(self, validated_data):
```
ddohler, line 52: For operations that affect multiple objects at once, it's good practice to use [`@transaction.atomic`](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#django.db.transaction.atomic) in order to ensure that things don't get left in a half-completed state if something goes wrong partway through.

src/backend/users/tests.py
```diff
@@ -87,3 +90,120 @@ def test_houseseeker_group_permissions(self):
...
+            content_type="application/json",
+        )
+        self.assertEqual(
+            first_response.status_code,
+            200,
+            f"Expected 200, got {first_response.status_code}. {first_response.content}",
+        )
+        second_response = self.houseseeker.post(
+            "/api/login/",
+            {"username": "testechouserdoesntexist@azavea.com"},
```
ddohler: It took me a bit to figure out what was going on here, I think because the `doesntexist` is kind of hidden in the middle of this email, maybe these could get changed to `exists@azavea.com` and `doesntexist@azavea.com`?

src/backend/users/tests.py
```diff
@@ -87,3 +90,120 @@ def test_houseseeker_group_permissions(self):
...
+            content_type="application/json",
+        )
+        self.assertEqual(
+            second_response.status_code,
+            200,
+            f"Expected 200, got {second_response.status_code}. {second_response.content}",
+        )
+        num_emails = len(mail.outbox)
```
ddohler: `len(mail.outbox)` is short enough and descriptive enough on its own that I think you could probably just inline it in the next line and avoid creating a separate variable (here and in other places).

src/backend/users/views.py
ddohler: Hah, yes! I made a comment about this in another spot before seeing this.

src/backend/users/views.py
```diff
@@ -1,46 +1,60 @@
...
-from .serializers import UserSerializer
+from .serializers import HouseSeekerSignUpSerializer, UserSerializer
+
+
+def send_login_link(request):
+    """
+    Send email with passwordless login link to User on Login/Sign Up
+    """
+    try:
+        # Request.data key could be "username" or "email", so turn into list to be safe
```
ddohler: TODO: What's up with this?

src/backend/users/views.py
```diff
@@ -1,46 +1,60 @@
...
+        send_mail(
+            "Your ECHO Login Link",
+            html_message,
+            "admin@domain.com",
+            [email],
+            fail_silently=False,
+            html_message=html_message,
+        )
+    except User.DoesNotExist:
+        raise Exception("User does not exist. Login email not sent.")
```
ddohler: This will have the effect of hiding some information because it'll transform the `User.DoesNotExist` exception (which is pretty descriptive) into a generic `Exception` (albeit one with a descriptive message). I think it'd be preferable to allow the User.DoesNotExist to bubble up so that it can be handled appropriately by the calling function. So I'd remove these last two lines and add a docstring noting that this function will raise an exception if the user doesn't exist, and then alter the login and signup views appropriately.

src/backend/users/views.py
```diff
@@ -1,46 +1,60 @@
...
-                "admin@domain.com",
-                [email],
-                fail_silently=False,
-                html_message=html_message,
-            )
-        except User.DoesNotExist:
+            # Will throw exception if cannot find the User
+            # For privacy reasons, do not send 500 error back if not found
+            send_login_link(request)
+        except Exception:
```
ddohler:
```python
except Exception:
    pass
```
Can be kind of a dangerous thing to have in a file because it'll catch and attempt to ignore pretty much any conceivable problem that could occur in the `try` block. This can make it harder to learn about and diagnose issues because there won't be any logging information available. For this reason I'd revert this to `except User.DoesNotExist` so that only that error is ignored and allow any other exceptions to bubble up so that it yells at us if there are other problems beyond what we specifically know we want to ignore.

src/backend/users/views.py
```diff
@@ -106,3 +124,22 @@ def get(self, request, **kwargs):
...
+
+class SignUpPage(APIView):
+    def post(self, request, **kwargs):
+        signup_message = "Thank you! You'll receive an email shortly with a link to complete your account. Click the link to create your profile and get started with ECHO."
+        try:
+            user_serializer = HouseSeekerSignUpSerializer(data=request.data)
+            user_serializer.is_valid(raise_exception=True)
+            user_serializer.save()
+            send_login_link(request)
+        except Exception as e:
```
ddohler: It's possible to have multiple `except` clauses, so this can be restructured as
```python
except IntegrityError:
    ...
except ValidationError:
    ...
except Exception:
    ...
```

src/backend/users/views.py
```diff
@@ -106,3 +124,22 @@ def get(self, request, **kwargs):
...
+            if type(e) == IntegrityError:
+                signup_message = "It looks like we already have an account with that email. Sign in by clicking the link below!"
+            if type(e) == ValidationError:
+                signup_message = "Please try again with a valid email address."
+            if type(e) == Exception:
```
ddohler: I would leave this catch-all error handling to the frontend; this catch-all will have the effect of converting all errors into 400 errors, which could be misleading to both users and ourselves. It'll also prevent stack traces from getting printed to logs which will make it harder for us to diagnose an error in production.
```suggestion

```

This means we'll need to attempt to parse the error message on the frontend, but if it doesn't exist we'll have to fall back to printing a generic error there.

src/backend/users/views.py
```diff
@@ -73,6 +87,10 @@ def get(self, request, **kwargs):
             "OT": "Other",
         }
 
+        # On initial sign up a UserProfile will be empty
+        if user_profile is None:
```
ddohler: I suspect you may have inadvertently solved this problem twice, because it doesn't look to me like this can possibly happen. For users who come in through the ordinary flow, `HouseSeekerSignUpSerializer` will ensure that they get a `UserProfile`, so they shouldn't ever encounter this case. When I sign up with a new email address, on logging in for the first time I get a full-fledged (albeit empty) Profile object, rather than this response. Please let me know if I'm missing something here!

taui/src/components/edit-profile.js
```diff
@@ -64,7 +64,7 @@ class EditProfile extends PureComponent<Props> {
   componentWillReceiveProps (nextProps) {
     // Listen for when profile to appear on props, because it is not present
     // on initial load. Only load once by checking state.
-    if (!nextProps.isLoading && nextProps.userProfile && !this.state.key) {
+    if (!nextProps.isLoading && nextProps.userProfile && !nextProps.userProfile.unverifiedUserProfile) {
```
ddohler: As noted above, I suspect this may no longer be necessary.

src/backend/users/views.py
```diff
@@ -1,46 +1,60 @@
...
+
+
+def send_login_link(request):
+    """
+    Send email with passwordless login link to User on Login/Sign Up
+    """
+    try:
+        # Request.data key could be "username" or "email", so turn into list to be safe
+        email = list(request.data.values())[0]
+        user = User.objects.get(username=email)
```
ddohler: I just had a thought: could we check for a user's profile and give them a blank one if it doesn't exist at this point? You're already doing that in the signup serializer, and I think it's a good idea to continue doing that, but doing it here would address the issue where someone who has been signed up manually using the admin or CLI may not have a profile set. There's no reason that we wouldn't want those people to be able to log in normally, so we might as well just give them a blank profile if we detect they're missing one.

ddohler (changes requested):
There are some code style things that I think should be addressed before merging, but the overall structure of this is looking good! Everything worked as expected in my testing.

src/backend/users/views.py
rachelekm: This is a great point! Actually, what do you think about creating a custom `User` manager class that creates an empty `UserProfile` instance with the `User` instance so everything's good to go on `create`? It might mean we have a build up of empty `UserProfile` instances, but like you said it would allow every `User` to be authorized to view the site by default once they have the link. I _think_ `Counselors` can only view/edit a `HouseSeeker` `UserProfile` in admin, but if not I can open an issue so the empty `UserProfiles` don't show up for them. Either way, I think a check for profile before sending the link is good here and will pop that in!

src/backend/users/views.py
rachelekm: Awesome! I see your comment below and will edit this to use the @transaction.atomic decorator

src/backend/users/views.py
rachelekm: Is this in reference to the TODO below or the comment on line 21? I added the comment to add some clarification behind how we get `email` from the `request.data`, but I'm realizing that may be more confusing/unnecessary than helpful! I'll take it out, if so.

src/backend/users/views.py
rachelekm: Oh jeez, ok will definitely revert this back.

src/backend/users/views.py
rachelekm: Ah, ok got it will do!

src/backend/users/views.py
rachelekm: Another "oh jeez", ok I'll get rid of this catch-all and make sure I update the frontend appropriately.

src/backend/users/views.py
rachelekm: Oh, huh! Nope you're right, I thought I was covering my bases but this isn't needed! I'll take this out along with the rest of the references to `unverifiedUserProfile`.

src/backend/users/views.py
ddohler: Ugh, this was a TODO for myself to look into this closer, which I lost track of before submitting the review. Sorry! Anyway, this piqued my interest because it made me wonder whether the request payload could be different depending on whether this was a signup or a signin situation, and whether maybe we should find a way to unify them to make sure it's always consistent. But I didn't actually check into any of those things, so I don't know whether any of those concerns are valid.

src/backend/users/views.py
ddohler: That's a good thought! If doing that, we'd need to keep in mind that there are several different ways to create objects in Django and some of them [bypass the usual methods](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#bulk-create). We've put a fair bit of effort on past projects into ensuring that related objects always, _always_ get created at the same time, and it can be a fair bit of work to cover all the bases. We don't have much more work to do on this phase of the project, so I'm thinking that, for now, it's probably safe to do this in the few places where we know we'll need it, but if we add more User-related features later or find that we're encountering bugs related to profiles being missing then we should revisit and consider doing something like a custom User manager.

src/backend/users/views.py
rachelekm: I took another look back and I think unifying the payloads is doable! When creating a new `User` it expects to find the `username` field, but I don't think there's any reason not to update the payload key from `email` to `username` when logging in. I gave it a try and logging in and signing up both are functioning as expected so far.

src/backend/users/views.py
rachelekm: Ok, got it! For now I'll implement the check/creating an empty `UserProfile` here and will keep an eye out for any funny business testing with admin/CLI.

rachelekm:
Thanks @ddohler! I implemented the changes from our discussion above and updated the testing instructions so this should be good to go for a second review.

src/backend/users/serializers.py
ddohler: Looks like this still needs the `@transaction.atomic` decorator.

src/backend/users/views.py
```diff
@@ -183,3 +198,19 @@ def put(self, request, *args, **kwargs):
         serializer = UserSerializer(User.objects.get(username=request.user))
         content = self.repackage_for_frontend(serializer.data)
         return Response(content)
+
+
+class SignUpPage(APIView):
+    def post(self, request, **kwargs):
+        signup_message = "Thank you! You'll receive an email shortly with a link to complete your account. Click the link to create your profile and get started with ECHO."
```
ddohler, line 205: This can create challenges for translation because these are user-visible strings that don't pass through the frontend translation framework. I wouldn't change anything now in the interest of time, but for the future, if there's translation happening client-side, I'd have the backend send back message keys, maybe something like `'SignIn.AccountExists'` or `'SignIn.InvalidEmail'`, and then use those to determine what to display to the user client-side.

ddohler (approved):
Works great! I think one of the comments from the previous review round slipped through without getting addressed, but aside from that this is good to go!

I did notice one problem with logging out that may have been obscured by the earlier problems with loading routing networks, or it may have been something introduced by a merge conflict. If I try to log out from `/map` I don't make it back to the login screen, and instead there's a console error saying `Uncaught (in promise) TypeError: profile is null`. If you're able to replicate this can you open an issue for it (I have been having some troubles with cached assets acting strangely, I think. I'm fairly certain this behavior is real but want someone else to double-check)?

rachelekm:
Thanks @ddohler! I tested it locally and noticed the same problem logging out, I just opened an issue for it #511.

> one of the comments from the previous review round slipped through without getting addressed, but aside from that this is good to go!
>
^ For the `@transaction.atomic` decorator I added `with transaction.atomic():` to the `try` block in the `SignUpPage` view to allow us to customize the signup message sent back in the response if there is an `IntegrityError`. Should this instead be a decorator on the serializer's `create`?

ddohler:
> For the @transaction.atomic decorator I added `with transaction.atomic()`: to the try block in the SignUpPage view to allow us to customize the signup message sent back in the response if there is an IntegrityError. Should this instead be a decorator on the serializer's create?

Oh, sorry, I missed that! (Though I thought I searched for it, but anyway 🤷 ) But yes, it should be moved to the serializer's `create()` method -- you want the `transaction.atomic` to be as close to any database-altering calls (`.save()`, `Model.objects.create()`, etc.) as possible so that it's guaranteed to wrap them no matter what happens outside the function. Similarly, because you [shouldn't catch exceptions inside a `transaction.atomic()`](https://docs.djangoproject.com/en/4.0/topics/db/transactions/#django.db.transaction.atomic), it's also a good idea to keep the `transaction.atomic()` as tight around the affected code as possible, which may mean using a `with transaction.atomic():` instead of the decorator if the function is larger. But in this case, you need to cover pretty much the whole `create()` function (except the `return`), so the decorator is appropriate.

rachelekm:
Ah, ok thank you! I understand why the decorator on the serializer works better here and that change has been addressed.

Issue #487, opened 2022-05-06
Fix api access tokens on staging env


rachelekm:
When starting/editing a user profile on staging the address field no longer searches destinations for a user to select. Instead, it stalls on loading (we are unable to manually type in an address to continue as well) and we cannot move on to saving the profile:
<img width="650" alt="Screen Shot 2022-05-06 at 10 01 23 AM" src="https://user-images.githubusercontent.com/36374797/167149040-a878126b-3bd1-4f30-b506-876687857c6a.png">
Taking a look at the network tab it appears we have no access token available for the mapbox geocoding API, so we should make that available and any other environment variables needed for the app to function on staging:
<img width="941" alt="Screen Shot 2022-05-06 at 10 01 58 AM" src="https://user-images.githubusercontent.com/36374797/167149521-9834aa0f-2156-4e17-a3a9-e9154e720f78.png">
This is related to the [discussion](https://github.com/azavea/echo-locator/pull/419#discussion_r827235296) from PR #419.

PR #500, opened 2022-05-12
Load dotenv file in CI/CD

aaronxsu:
## Overview

This PR:
- updates some frontend app code based on linter/formatter results
- updates the format script to also format frontend app code
- updates the bootstrap script to pull down `.env` file
- updates the `ci.yml` recipe to conditionally run `bootstrap` script

### Notes
1. I created a dev config bucket `echo-locator-devdjango-config-us-east-1 ` and uploaded the dotenv file for dev there.

## Testing Instructions

 * `AWS_PROFILE=<echo profile> ECHOLOCATOR_SETTINGS_BUCKET=echo-locator-devdjango-config-us-east-1 ./scripts/setup`
 * `./scripts/server`
 * The app frontend and backend should be built and run successfully
 * CI should also be happy

Closes https://github.com/azavea/echo-locator/issues/487


.github/workflows/ci.yml
```diff
@@ -23,6 +23,11 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: "us-east-1"
 
+      - run: ./scripts/bootstrap
+        if: github.ref != 'refs/heads/master'
```
aaronxsu, line 27: For non-production environment, we want to run this step in order to use the corresponding config bucket to pull down the dotenv file. This is to prepare for the `cibuild` step.

I was not so sure about the deployment story of production, so I skipped uploading a dotenv file to the production bucket, and I also did not add an action here for prod either.

scripts/bootstrap
```diff
@@ -15,13 +15,15 @@ Pull down secrets from S3.
...
 function pull_env() {
     pushd "${DIR}"
 
-    echo "Pulling .env from ${1}"
-    # aws s3 cp "s3://${1}/.env" ".env"
+    if aws s3 cp "s3://${ECHOLOCATOR_SETTINGS_BUCKET}/.env" ".env"; then
+        echo "Updated environment variables file from ${ECHOLOCATOR_SETTINGS_BUCKET}"
+    elif [[ ! -f ".env" ]]; then
+        cp .env.sample .env
+        echo "Updated environment variables file from .env.sample"
```
BryanQuigley: Maybe make this error make it clear that it's ok if this fails and you want to do local development.

aaronxsu: Sounds good. How about

>"Updated environment variables file from .env.sample for local development. Please populate the fields in this file manually."

BryanQuigley (approved):
LGTM

aaronxsu:
Update: 
- I pushed this commit https://github.com/azavea/echo-locator/pull/500/commits/be8c7a5efae3f74e4383c9a4c70de158c54e1b10 up to fix the build context issue
- the idea was borrowed from @rachelekm 's work https://github.com/azavea/echo-locator/pull/484 
- I pushed a `test/` branch up and made sure that the env was read correctly as the geocoding worked as expected: https://azavea.slack.com/archives/CELFP4B4N/p1652461701619289

scripts/bootstrap
aaronxsu: Updated here: https://github.com/azavea/echo-locator/pull/500/commits/689495395de9c42c5a485d38b62819572c2a5eab

.github/workflows/ci.yml
ddohler: We don't have a Phase II deployment task yet -- could you create one and add this as one of the required tasks?

scripts/bootstrap
```diff
@@ -15,13 +15,15 @@ Pull down secrets from S3.
...
 function pull_env() {
     pushd "${DIR}"
 
-    echo "Pulling .env from ${1}"
-    # aws s3 cp "s3://${1}/.env" ".env"
+    if aws s3 cp "s3://${ECHOLOCATOR_SETTINGS_BUCKET}/.env" ".env"; then
+        echo "Updated environment variables file from ${ECHOLOCATOR_SETTINGS_BUCKET}"
+    elif [[ ! -f ".env" ]]; then
+        cp .env.sample .env
+        echo "Updated environment variables file from .env.sample for local development. Please populate the fields in this file manually."
```
ddohler:
```suggestion
        echo "Copied environment variables file from .env.sample to .env for local development. Please populate the fields in this file manually."
```

src/frontend/src/actions/profile.js
```diff
@@ -31,9 +31,7 @@ export const setProfile = (profile) => (dispatch, getState) => {
 
 export const sendLoginLink = (email) => (dispatch, getState) => {
   addActionLogItem(`Sending login link to ${email}`);
-  axios.post("/api/login/", {
-    email: email,
-  });
+  axios.post("/api/login/", { email });
```
ddohler, line 34: Just double-checking, things like this will cause CI to fail once #457 is merged, right?

scripts/bootstrap
```diff
@@ -15,13 +15,15 @@ Pull down secrets from S3.
...
 }
 
-# TODO: potentially as a part of this issue: https://github.com/azavea/echo-locator/issues/376
-# after switching to a different bundler, we may want to copy and populate some env vars
 function pull_env() {
     pushd "${DIR}"
 
-    echo "Pulling .env from ${1}"
-    # aws s3 cp "s3://${1}/.env" ".env"
+    if aws s3 cp "s3://${ECHOLOCATOR_SETTINGS_BUCKET}/.env" ".env"; then
```
ddohler: If I run this with a bare `./scripts/setup` I get the following output:
```
fatal error: Parameter validation failed:
Invalid bucket name "": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$" or be an ARN matching the regex "^arn:(aws).*:(s3|s3-object-lambda):[a-z\-0-9]*:[0-9]{12}:accesspoint[/:][a-zA-Z0-9\-.]{1,63}$|^arn:(aws).*:s3-outposts:[a-z\-0-9]+:[0-9]{12}:outpost[/:][a-zA-Z0-9\-]{1,63}[/:]accesspoint[/:][a-zA-Z0-9\-]{1,63}$"
```

I think maybe this just needs to check for the presence of the `ECHOLOCATOR_SETTINGS_BUCKET` environment variable before attempting to download? And then the `.env` check should work fine as-is so that if someone already has a `.env` file it won't do anything and if they don't it'll copy them one from the sample.

ddohler (approved):
I did find one very small thing that I think needs to change but overall this does what it's supposed to! Thanks for getting this figured out!

src/frontend/src/actions/profile.js
aaronxsu: Good question. The short answer is no for the existing app code, but yes for the future frontend app code. I updated the `format` script, and such things should be autocorrected by that script. For example, this change was made automatically by the format script. I already ran the formatting and linting command with fixes in the bundler PR. 

So, how about I add a checkbox item in the PR template to suggest running the `./scripts/format`?

ddohler: That sounds good to me! It'll help people to remember to run `scripts/format` even if CI is going to catch any problems, so it may save some time.

aaronxsu: Agreed!

.github/workflows/ci.yml
aaronxsu: Good point. Done! https://github.com/azavea/echo-locator/issues/502

src/frontend/src/actions/profile.js
aaronxsu: Addressed here: https://github.com/azavea/echo-locator/pull/500/commits/262686ecae83a0af62538305cebe8a917f020884

scripts/bootstrap
aaronxsu: That is a great point. I modified the code block in there so that:
- if there is already a dotenv file, do not bother do anything extra
- otherwise:
      - if no S3 bucket provided as the env var, copy from the sample dotenv
      - if has S3 bucket provided as env var, pull from S3

This is reflected here: https://github.com/azavea/echo-locator/pull/500/commits/5cd4b0c4e4a282717610e36c8b4a32f16f5a2124

ddohler (approved):
I didn't re-test, but this looks good!

PR #503, opened 2022-05-13
Replace docker CMDs with gunicorn config

BryanQuigley:
This replaces two docker CMD lines - one was actually always overridden
with one gunicorn config file.

This also allows programmatically different config for development vs staging/production, namely:
 - development:reload enabled for reloading workers when code changes
 - prod/stg: preload enabled possibly saving memory

## Overview

The CMD in dockerfile was doing nothing, so I initially started with just removing that. But reload is not recommended in production so I wanted to vary it by that.

### Notes

Other options I considered and decided not to go with:
 - Some gunicorn settings can be set as environment variables, but this a config file seems cleaner
 - It's also possible to define some things in the gunicorn config and still run the CMD in docker, but why when it can all be in one place.

For future consideration:
It's also possible to have items like workers be generated on the fly based on the number of cpus - but we don't performance constraints for this project so not planning to optimize.

## Testing Instructions

 * Ensure works as before
 * Confirm in debug output that reload is set in development and on deployment that reload is off, and preload is on.

### DIFF
--- a/src/backend/gunicorn.conf.py
+++ b/src/backend/gunicorn.conf.py
@@ -1,0 +1,16 @@
+import os
+
+bind = ":8085"
+accesslog = "-"
+errorlog = "-"
+workers = 3
+loglevel = "Debug"
+
+ENVIRONMENT = os.getenv("DJANGO_ENV", "Development")
+
+if ENVIRONMENT == "Development":
+    reload = True
+else:
+    preload = True
+
+wsgi_app = "echo.wsgi"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,14 +37,6 @@ services:
     depends_on:
       database:
         condition: service_healthy
-    command:
-      - "-b :8085"
-      - "--reload"
-      - "--timeout=90"
-      - "--access-logfile=-"
-      - "--error-logfile=-"
-      - "--log-level=debug"
-      - "echo.wsgi"
     ports:
       - "8085:8085"
     cpus: 2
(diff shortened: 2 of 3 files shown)


aaronxsu (approved):
Tested locally and on staging by a `test` branch. Worked as advertised. I think the items under the `Notes` section in this PR sound legit, and we may create tech debt issues if need be- up to you!

Update: discussed on Slack- the updated notes section clarifies the previous confusions. This is good to go 🚀

src/backend/gunicorn.conf.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+bind = ":8085"
+accesslog = "-"
+errorlog = "-"
+workers = 3
+loglevel = "Debug"
```
aaronxsu, line 7: I think debug level logging makes a lot of sense for local dev. What is our usual practice for staging and production? Should it be dependent on the env if they differ?

BryanQuigley: We can certainly vary this - and now programmatically based on environment.  But AFAICT all Gunicorn does for debug is spit out all the values it's been set with. ANd this is just debug for gunicorn not Django.

aaronxsu: Got it. If this is the case, then I think keeping this as is seems fine 👍🏼

rachelekm (approved):
I tested this locally and with a test branch. App functions as expected and everything logs as normal locally and I don't see anything abnormal on staging logs — I can see in the task logs that reload is set to `False` as expected. I think this looks good to go!

PR #507, opened 2022-05-16
Handle voucher/no voucher cases in EditProfile

### LINKED ISSUE #461: Profile editing: Handle voucher / no-voucher cases
- On the profile screen, users should see a checkbox allowing them to specify whether they have a voucher or not.
- The checkbox should default to unchecked.
- If the checkbox is unchecked, they should see 2 fields
  1. Rent budget
  2. Desired number of bedrooms
- If the checkbox is checked, they should see 3 fields
  1. Voucher max rent
  2. Number of bedrooms on voucher
  3. Voucher number
- None of the fields should be required
- There should be appropriate translations for any new text

## Notes
- The demo site (demo.echosearch.org) has some of this functionality implemented already (or maybe all of it). I wouldn't necessarily copy any code from the demo branch, but you can at least consult the demo site for layout and text suggestions.

ktohalloran:
## Overview

This PR adds conditional logic to display different questions in the `EditProfile` component based on whether or not the user has a voucher.

### Checklist

### Demo

Screen Shot 2022-05-16 at 1 47 08 PM

If a user _does_ have a voucher, they will see the "voucher number" and "number of bedrooms on your voucher" fields.

Screen Shot 2022-05-16 at 1 47 00 PM

 If they _do not_ have a voucher, they will see the "budget for apartment" and "desired number of bedrooms" fields.

### Notes

I initially decided to use radio buttons as opposed to a checkbox for the "Do you have a voucher" question because that was what the demo site uses. I also spoke with @gatesgodin, who verified that the radio buttons were appropriate here.

Currently, if a user switches from being a voucher user to not (or vice versa), the database will still keep the previously entered values, e.g. if a user who initially had a voucher filled out their voucher number and number of bedrooms then switched to not having a voucher and filled out their desired number of bedrooms and rent budget, the database would keep all 4 values and would not clear the voucher-related values following the switch.

## Testing Instructions

 * Run `./scripts/server`
 * Navigate to **localhost:9966** and login
 * Click "Edit profile" and verify that changing your answer to "Do you have a voucher?" changes the fields as illustrated above
 * Make a change to one of the altered fields and verify that it successfully saves to the database

Resolves #461


ddohler (approved):
Looks good, works great!

Issue #501, opened 2022-05-13
Application no longer routes to neighborhoods


ddohler:
The cause is that the Cloudfront domain which used to host the routing files has disappeared (see #495 ). This issue is to track resolution.

The following screenshots show the issue on production, but it is consistent across all environments.

<img width="1512" alt="Screen Shot 2022-05-13 at 08 43 37" src="https://user-images.githubusercontent.com/447977/168285886-7209b41d-08dd-4415-ba5b-0fc4d38b3712.png">
<img width="1512" alt="Screen Shot 2022-05-13 at 08 44 29" src="https://user-images.githubusercontent.com/447977/168285892-5362853b-f2ca-4284-83b4-bf929afbb9fc.png">
<img width="1512" alt="Screen Shot 2022-05-13 at 08 44 44" src="https://user-images.githubusercontent.com/447977/168285893-be367121-33e6-428a-a334-d70a66560a30.png">

ddohler:
Conveyal has reactivated the Cloudfront endpoint and routing works once again 🎉 . If we decide to take any other action I'll open a separate issue.

PR #509, opened 2022-05-16
Enable user logout

### LINKED ISSUE #444: Enable logging out
It should clear any values stored in cookies or local storage and direct the user to the login page. This should work for both authenticated and anonymous logins.

rachelekm:
## Overview

This PR enables users to log out of their current session, including: clearing authentication tokens stored in cookies, profile information in app state, and any local storage used for authenticated and anonymous use.

### Checklist

### Notes

- Update: This has been resolved, see discussion below. On logout, django responds with a redirect to "/". The `auth_token` cookie is no longer on the response, but I did notice that the `Authorization` header was still unchanged on the redirect "/" request. I could not seem to get this to clear for the redirect. However, if I prompt a page load the `Authorization` header no longer exists. I'm thinking that because the action dispatch to clear `authToken` only happens after the action promise resolves it still exists in time for the redirect. I ended up leaving out the page reload since while the `Authorization` header is visible it can't be used to successfully make calls anymore, and it looks a little funny to be brought back to the login page and then suddenly have it reload to the exact same page—but maybe a loading spinner here would work as well if the page reload seems necessary? All other requests do not include the `Authorization` header or cookie until logging back in.
- Since there are some issues with neighborhood routing (#501 ) it's tricky to test some of the features we would expect to reset after logout—mostly, any map/neighborhood preferences like displaying listings. I've updated the 'clear data' action which is dispatched on main page initialization so I'm hoping that's all that's needed until we can further test.

## Testing Instructions

 * `scripts/server`
 * `localhost:9966/`
 * Submit an email for a verified `User` with a `UserProfile` and follow the "emailed" link
 * Confirm the logout button appears and, when clicked, you're brought back to the login screen
 * In Chrome Dev Tools --> Networks tab select the most recent request for `localhost` and confirm the `auth_token` cookie is no longer there
 * Select the option to continue as an anonymous user and confirm logout button exists and also takes you back to the login page

Resolves #444

### DIFF
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -57,6 +57,16 @@ class ObtainToken(APIView):
         return response
 
 
+class DeleteToken(APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, **kwargs):
+        response = HttpResponseRedirect("/")
+        response.delete_cookie("auth_token")
+        Token.objects.get(key=request.auth.key).delete()
+        return response
+
+
 class UserProfileView(APIView):
     permission_classes = (IsAuthenticated,)
     map_priorities_to_nums = {"NI": 1, "SI": 2, "I": 3, "VI": 4}
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -60,8 +60,8 @@ class ObtainToken(APIView):
 class DeleteToken(APIView):
     permission_classes = (IsAuthenticated,)
 
-    def get(self, request, **kwargs):
-        response = HttpResponseRedirect("/")
+    def post(self, request, **kwargs):
+        response = Response(status=200)
         response.delete_cookie("auth_token")
         Token.objects.get(key=request.auth.key).delete()
         return response
--- a/src/frontend/src/actions/profile.js
+++ b/src/frontend/src/actions/profile.js
@@ -55,7 +55,7 @@ export const sendLoginLink = (email) => (dispatch, getState) => {
 export const setLogout = (authToken) => (dispatch, getState) => {
   addActionLogItem(`Deleting token cookie`);
   axios
-    .get("/api/logout/", {
+    .post("/api/logout/", null, {
       headers: {
         Authorization: `Token ${authToken}`,
       },
--- a/src/frontend/src/actions/profile.js
+++ b/src/frontend/src/actions/profile.js
@@ -53,17 +53,18 @@ export const sendLoginLink = (email) => (dispatch, getState) => {
 };
 
 export const setLogout = (authToken) => (dispatch, getState) => {
-  addActionLogItem(`Deleting token cookie`)
-  axios.get('/api/logout/', {
-    headers: {
-      'Authorization': `Token ${authToken}`
-    }
-  })
-    .then(() => {
-      dispatch({type: 'set auth token', payload: null});
-      dispatch({type: 'set profile', payload: null});
+  addActionLogItem(`Deleting token cookie`);
+  axios
+    .get("/api/logout/", {
+      headers: {
+        Authorization: `Token ${authToken}`,
+      },
     })
-}
+    .then(() => {
+      dispatch({ type: "set auth token", payload: null });
+      dispatch({ type: "set profile", payload: null });
+    });
+};
 
 export const setAuthToken = (authToken) => (dispatch, getState) => {
   addActionLogItem("Updating authToken");
--- a/src/frontend/src/actions/profile.js
+++ b/src/frontend/src/actions/profile.js
@@ -52,6 +52,19 @@ export const sendLoginLink = (email) => (dispatch, getState) => {
   dispatch({ type: "set login message", payload: "SignIn.LoginLinkSent" });
 };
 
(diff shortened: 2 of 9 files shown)


src/backend/users/views.py
```diff
@@ -55,6 +55,16 @@ def get(self, request, **kwargs):
         return response
 
 
+class DeleteToken(APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, **kwargs):
```
ddohler: I think this would be more natural as a `POST` -- `GET`s are usually more about retrieving information, whereas a `POST` is more of a "do-something" verb.

src/frontend/src/components/custom-header-bar.js
```diff
@@ -54,6 +54,13 @@ class CustomHeaderBar extends React.Component {
             中文
           </button>
         </div>
+        <div className="app-header__logout">
+          <span className="app-header__button">
+            <button className="app-header__button" onClick={() => this.props.logout()}>
+              Logout
```
ddohler: This should be wrapped in `t()` so that it can be translated later.

src/frontend/src/components/edit-profile.js
```diff
@@ -531,168 +530,166 @@ class EditProfile extends PureComponent<Props> {
...
         {errorMessage && <p className="account-profile__error">{errorMessage}</p>}
         <div className="form-screen__main">
-          {key && (
-            <div className="account-profile">
-              {!isAnonymous && (
-                <div className="account-profile__field">
-                  <label className="account-profile__label" htmlFor="headOfHousehold">
-                    {t("Accounts.Name")}
-                  </label>
+          <div className="account-profile">
```
ddohler: This is a minor thing, but when a formatter unexpectedly introduces changes like this that don't relate to the functional changes you're making, it's best to split out those changes into a separate commit as much as possible. This can make things easier on reviewers and later travelers who may be looking through commits to track down the source of a bug. I usually do this by using the "patch" feature of interactive-add: `git add -i`. It's not always easy to completely separate formatting from logic changes, but if you can get the bulk of it into a separate commit that's still helpful.

No need to go back and split these though, that's more trouble than it's worth at this point!

src/backend/users/views.py
```diff
@@ -55,6 +55,16 @@ def get(self, request, **kwargs):
...
+        response = HttpResponseRedirect("/")
```
ddohler: I don't think the redirect is necessary here--we can simply return a `200` status of some kind. It was more important on logging in because the initial request was coming from a user clicking an email link, so the Javascript part of the app hadn't loaded yet. When we're logging out, we don't have that problem, so we can just rely on the frontend to navigate back to `/` (and in fact, if we leave this out it still redirects back to `/` correctly).

src/frontend/src/actions/profile.js
```diff
@@ -35,6 +35,20 @@ export const sendLoginLink = (email) => (dispatch, getState) => {
   dispatch({ type: "set login message", payload: "SignIn.LoginLinkSent" });
 };
 
+export const setLogout = (authToken) => (dispatch, getState) => {
+  addActionLogItem(`Deleting token cookie`);
+  axios
+    .get("/api/logout/", {
+      headers: {
+        Authorization: `Token ${authToken}`,
```
ddohler, line 60: Not redirecting from Django will also fix the issue with the second request to `/` being made with the authorization token header still attached because it's Axios that is following the redirect using the parameters that were initially passed to this `get()` call.

ddohler (approved):
Works as expected! There are some cleanup items to address but this is good to go after that! I should note that I wasn't able to edit my profile, but I wasn't expecting to be able to do that since #497 hasn't been merged yet.

src/backend/users/views.py
rachelekm: Got it, will change!

src/frontend/src/components/edit-profile.js
rachelekm: Whoops, will start doing this more thoughtfully—I haven't used patch much so that's helpful to know when I need to separate out formatting.

src/frontend/src/actions/profile.js
rachelekm: Ah ha! Thank you, will implement and update my comment! This is irrelevant now but for my own learning: after trying the delete the header in the response in `views`, I also tried creating an `axios` instance that would be conditionally configured with an authorization token header if `authToken` existed (and the header deleted when not), taking some inspiration from [these lines in treetective](https://github.com/azavea/treetective/blob/develop/client/src/api.ts#L59-L70). This obviously didn't work out so I dropped it, but I initially thought that was because of the state of `authToken`—is it because a redirect will always use the same `axios` instance as the original request so it's not actually creating a "new" instance?

ddohler: > is it because a redirect will always use the same axios instance as the original request so it's not actually creating a "new" instance?

Yup, I think that's exactly what's happening! It's transparently making multiple requests under the hood, but on the "outside" you only see the one Promise with its eventual result.

rachelekm: Ah got it, thanks for clarifying!

Issue #502, opened 2022-05-13
Phase II Deployment


aaronxsu:
The production application should be deployed to app.echosearch.org

Below is a list of known one-off items that need to be done in order to complete the deployment.

- [x] One task we are now sure to be on the to-do list is to plan and implement the handling of env vars for the frontend for production deployment, as discussed in here: https://github.com/azavea/echo-locator/pull/500#discussion_r872656289
- [x] Importing Users
   1. Add AmazonS3ReadOnlyAccess to ecsecho*TaskRole
   2. Enter container via `aws ecs execute-command`
   3. Run ./manage.py loaddata users/fixtures/group_permissions from (should this be done elsewhere?)
   4. `./manage.py migrate_housing_seeker_data -b <bucket-name> -p public`
- [x] Add `default_from_email` and `aws_s3_photo_bucket` variables to `terraform.tfvars` in S3, like we have for staging
- [x] Import neighborhoods
  1. Locally, run `./update_data.sh` in `neighborhood_data` directory, then run `./scripts/imagepublish production` to push the neighborhood images to production assets bucket
  2. Run `./manage.py import_neighborhood_data s3://echolocator-production-data-us-east-1/neighborhoods.json s3://echolocator-production-data-us-east-1/neighborhood_bounds.json s3://echo-locator-production-site-us-east-1/assets/neighborhoods/` via `execute-command` as with the user import.
- [x] Complete the SES setup as described in #491
- [x] Send BHA links for the Login and Signup buttons on https://echosearch.org/

Note that this task can be started before #563 is done, but can't be finished until #563 is because it needs to include the update data

ddohler:
Still waiting on client feedback; we may need to swap this out if there are critical issues that are discovered.

Issue #448, opened 2022-04-14
Remove old amplify style deployment


BryanQuigley:
Once we have switched over production to django based back-end we should:
- Tear down the old terraform infra and delete the /terraform directory (or additionally mv terraform-django/ terraform/ ?)
- Delete or rename the non-Django Deployment README
- Cleanup STRTA to make sure we don't have old scripts
- Other README cleanups as necessary
- Remove Add AmazonS3ReadOnlyAccess to instance (or further restrict if we actually do need S3 access for something else)
- Do something to migrate S3 photo bucket from old terraform to new created in old, but defined to pull from in new (See [comment]https://github.com/azavea/echo-locator/pull/561#pullrequestreview-1017177127)  )

BryanQuigley:
S3 access is needed.   And I've killed the old terraform config so the production-site bucket shouldn't be taken down.

PR #512, opened 2022-05-18
Add s3readonly and terraform version bump

BryanQuigley:
## Overview

s3 read only is needed to import the users from amplify
terraform has a 1.0 promise so upgrading is easy.

### Notes

We likely should remove the S3 permissions once we have are done with migration - will add to #448

## Testing Instructions

 * Test #502 Importing Users without doing the permission add.


aaronxsu (approved):
Tested it by pushing up a `test` branch. Saw that the policy was attached to the task in the staging cluster. Successfully ran the housing seeker migration script without boto3 complaining about S3 access. Good to go 🚀

PR #514, opened 2022-05-19
Fix logout error

### LINKED ISSUE #511: Fix redirect on logout
As discussed in PR #499, the app does not display the login/signup component after logging out from either `/map` or `/profile`. Instead, we remain on the page but just see a grey screen since we are no longer authorized to view the app components. This was likely obscured by the earlier problems loading routing networks as the below error logs to the console from the network-neighborhood-route selector after logout:
Screen Shot 2022-05-18 at 3 51 09 PM
We should investigate this issue and resolve it so that a `User` is redirected back to the login page after logout.

rachelekm:
## Overview

This PR addresses the logout error that appeared after fixing network routing. Now, when a logged-in `User` selects logout, they are directed back to the login/signup page. Since there is an opportunity for a `User` to be authorized to use the site, but may not have a `userProfile`, we needed to add a conditional in the networks selector to return empty if a profile doesn't exist.

### Checklist

## Testing Instructions

 * `scripts/server`
 * `http://localhost:9966/`
 * Login, click around (or not), and then click logout
 * Confirm you are logged out and the signup/logout component displays
 * Try logging out from both the `/profile` and `/map` page

Resolves #511

### DIFF
--- a/src/frontend/src/components/neighborhood-details.js
+++ b/src/frontend/src/components/neighborhood-details.js
@@ -33,7 +33,7 @@ class NeighborhoodDetails extends PureComponent<Props> {
 
     this.state = {
       isFavorite:
-        props.userProfile && props.neighborhood
+        props.userProfile && props.neighborhood && props.userProfile.favorites
           ? props.userProfile.favorites.indexOf(props.neighborhood.properties.id) !== -1
           : false,
     };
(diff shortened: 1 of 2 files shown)


ddohler (approved):
Works great! And this made me realize that we still need to implement the Favorites feature, which I had totally forgotten about 🤦 . So I opened #515 .

rachelekm:
Ah, that makes sense! I'll add a comment to the issue that those two conditional checks for `props.userProfile.favorites` in the `NeighborhoodDetails` can be taken out once the `favorites` field is back in.

PR #516, opened 2022-05-19
Include destination coordinates in API

### LINKED ISSUE #486: Replace dummy Point values in Profile as sent by backend
Following completion of [442](https://github.com/azavea/echo-locator/issues/442), the backend is currently sending dummy values in the place of actual latitude and longitude data. This should be replaced following the merge of [PR 478](https://github.com/azavea/echo-locator/pull/478) once location has been added as a field on the `Destination` model.

ddohler:
## Overview

Previously we were using placeholders for destination locations; this replaces the placeholders with values from the database.

### Checklist

### Demo

Screen Shot 2022-05-19 at 15 30 10
Screen Shot 2022-05-19 at 15 30 20

## Testing Instructions

 * Open up an existing profile and edit some locations so that they get re-geocoded (geocoding doesn't happen unless you type in the address box, so if locations have the default coordinates they'll keep them until they are changed).
 * Save the profile. Use the network inspector to view the request and confirm that the request / response payloads show the same values for `lat` and `lon` for each destination (i.e. no coordinates are getting switched between request and response).
 * View the map and choose different locations. Confirm that the start marker moves around on the map, and if you want, drop one of the addresses into another mapping service like Google Maps and confirm that we agree with them on where it is.

Resolves #486

### DIFF
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -1,4 +1,5 @@
 from django.contrib.auth.models import User
+from django.contrib.gis.geos import Point
 from django.core.mail import send_mail
 from django.db import transaction
 from django.db.utils import IntegrityError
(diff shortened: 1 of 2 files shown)


src/backend/users/views.py
```diff
@@ -98,19 +99,21 @@ class UserProfileView(APIView):
 
     def repackage_for_frontend(self, serialized_data):
         # repackage destinations to match frontend AccountAddress
-        # TODO replace dummy Point data issue 486
-        # (https://github.com/azavea/echo-locator/issues/486)
         user_profile = serialized_data["userprofile"]
+        print(user_profile["destinations"])
```
ktohalloran: Did you mean to leave this in here?

ktohalloran (approved):
Looks and works great! One small potential clean up as mentioned below, but good to go after that.

src/backend/users/views.py
ddohler: Oops, sure didn't! Thanks for catching that!

PR #517, opened 2022-05-20
Favorites support

### LINKED ISSUE #515: Add Favorites support to User Profile
1. The UserProfile model should have a `favorites` field which should be an [array of strings](https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/fields/#arrayfield) (it should default to empty: `list`). The values stored will be zip codes.
2. The user import script should populate this field when loading profiles
3. The API should include this field in GET requests and accept it on POST / PUT requests
4. The frontend should use this field to filter the `Saved` section of the dock (it already has logic to do this, so this is just making sure it uses the right field)
Screen Shot 2022-05-19 at 13 41 26
5. Clicking the star next to a neighborhood should trigger an API request to save the favorite to the user's profile.

ddohler:
## Overview

The existing site has a "favorites" function that allows users to keep a list of neighborhoods that they like. This implements the same functionality using the Django API.

### Checklist

### Demo

Screen Shot 2022-05-20 at 13 20 49

### Notes

I opted not to remove the checks for the presence of `.favorites` mentioned in https://github.com/azavea/echo-locator/issues/515 -- they should now be superfluous but they're not hurting anything and could help the app be more resilient to failures in the future.

## Testing Instructions

 * Run `./scripts/update` and confirm migrations apply cleanly.
 * Log in to the admin site and confirm that you see a field in the `UserProfile` editor for favorites
 * If you've imported profiles from S3 to your local environment, use the admin to delete all of them
 * Run `AWS_PROFILE=echo-locator ./scripts/manage migrate_housing_seeker_data -b echolocator-staging-profiles122738-staging -p public` to import staging profiles to your local instance.
 * Click into a few of the newly imported files in the admin and confirm that some of them have favorites set.
 * Log in to the frontend
 * Use the ⭐  buttons next to a neighborhood card to add / remove some favorites. Edit other parts of your profile too. Reload the admin page for your profile and confirm that the favorites get persisted.
 * Log out and log back in and confirm that your favorites are saved.
 * Confirm that clicking "Saved" shows only your favorites.
 * Log out and browse anonymously and confirm that you can save favorites, and that they persist if you close the browser window and come back but are cleared if you logout.

Resolves #515


src/frontend/src/components/dock.js
```diff
@@ -122,8 +123,22 @@ class Dock extends PureComponent<Props> {
       }
     }
 
-    profile.favorites = favorites;
-    handleAuthChange(profile);
+    // Save updated profile
+    const updatedProfile = { ...profile, favorites };
```
ddohler, line 126: This makes a copy of the profile so that the original object isn't changed by this function. This prevents the favorite from showing up until it's saved successfully.

src/frontend/src/components/dock.js
```diff
@@ -122,8 +123,22 @@ class Dock extends PureComponent<Props> {
...
+      // Despite the name, this saves the profile to local storage in anonymous mode.
+      this.props.handleAuthChange(updatedProfile);
+    } else {
+      // Save to API
+      this.props
+        .saveProfile(updatedProfile, this.props.authToken)
+        .then((newProfile) => {
+          this.props.setProfile(newProfile);
+        })
+        .catch((err) => {
```
ddohler, line 137: This isn't very friendly if it fails but I wasn't sure of the best way to visually communicate a failure here, so I opted to simply have it not work, and in most cases we'll get a stack trace in the Django logs to let us know something is wrong.

rachelekm: Since there's Django logs I think that's the right decision. The star remaining unfilled after selection feels like an immediate/straightforward enough indicator to either try again or signal it's not working as it's supposed to, so I think it covers our bases enough in that respect.

rachelekm (approved):
Looks great! Went through the testing instructions and admin/frontend functionality for both verified `User` and `IsAnonymous` work as expected, `favorites` persist as intended for each, and can successfully migrate profiles with `favorites` from S3.

Issue #522, opened 2022-05-23
Can't remove profile rent budget from either front or backend


ktohalloran:
The backend is not letting users erase their rent budget, either on the front or backend, even though the model allows that field to be both blank and null. It saves as expected if you never give it a value, but if you have entered a value and attempt to erase it from the frontend, the backend will respond with the following error:
![Screen Shot 2022-05-23 at 9 04 33 AM](https://user-images.githubusercontent.com/77936689/169826611-c84ed454-2736-4e2b-b2e8-ac180e4d9f07.png)

Similarly, if you try to erase it from the backend, it will respond with a vague reference to an error that is not reflected anywhere on the page:
<img width="1111" alt="Screen Shot 2022-05-23 at 9 10 54 AM" src="https://user-images.githubusercontent.com/77936689/169826956-6aeb635d-6a7c-4ab2-a127-07277419bc32.png">

PR #523, opened 2022-05-23
Feature/kob/fix neighborhood details

### LINKED ISSUE #513: Fix neighborhood detail and listing rooms/budget
When selecting a neighborhood as a logged in user, the details sidebar does not show the updated budget or number of rooms and listings no longer appear. In addition, our neighborhood detail links do not filter by rooms/budget anymore. I suspect this is because these still rely on the old `userProfile` key properties `rooms` and `estMaxRent` which no longer exist. We should update these as well as add conditionals where appropriate for when voucher/no voucher.
Screen Shot 2022-05-19 at 12 24 08 PM

ktohalloran:
## Overview

Following the addition of the no voucher use case, the properties on `AccountProfile` that had been used to populate the neighborhood detail view, namely `rooms` and `estMaxRent`, needed to be reworked. This PR adds logic to populate `rooms` (and therefore `estMaxRent`) based on whether the user has a voucher or not. In cases when the user does not have a voucher, it also changes the external housing app searches to use the user's set budget, rather than the neighborhood's max rent.

### Checklist

### Demo

Screen Shot 2022-05-23 at 8 46 27 AM
When a user does not have a voucher and has set a budget, it will display the neighborhood's max rent in the "Est. Max Rent" section of the neighborhood detail view, but the user's set budget in the "Search for apartments" headline of the external search tools, as well as filling their budget into those external search tools' queries.

### Notes

This PR includes a few nested ternaries to prevent funkiness from happening if a user does not fill out a preferred number of rooms, a budget, or both. While Zillow and Hotpads' queries were smart enough to use null values, GoSection8 and Craigslist links both needed default values set, as described below:
* If a user does not have a voucher and also has not set a budget, it will display the neighborhood's max rent in the "Search for apartments" headline.
* If a user had not filled out their preferred budget and clicked on GoSection8, the link was completely broken. As a result, that query will default to using the neighborhood's max rent if no budget has been set.
* If a user had not provided a number of rooms and clicked on Craigslist, the query would fail while attempting to look for a null value of bedrooms. It will now default to 0 if no value for rooms has been provided (this is what the demo site does).

While testing this, I discovered that users will get an error from the backend if they attempt to erase their budget, a bug that has been spun out into [issue 522](https://github.com/azavea/echo-locator/issues/522). As a related note, the current behavior of not allowing a null value for budget is also true on the demo site.

While testing this, I also discovered an issue with how priorities were being saved to the backend. This PR also adds a fix for that in commit [5b4c06c](https://github.com/azavea/echo-locator/commit/5b4c06cf347d63c78377f60dc5f258171648568f).

## Testing Instructions

 * Start the server, navigate to **localhost:9966**, and log in
 * Start by verifying that changes to priorities saves as expected
 * Then select a neighborhood and verify that the details fill in as expected:
   * If your profile _does_ have a voucher, check that the "Est. Max Rent" and "Search for apartments" headlines both display your profile's set number of bedrooms and the same max rent
   * Check that if you select that you _do not_ have a voucher and also do not fill out that profile's budget, that the "Search for apartments" headline autofills with the neighborhood's max rent and that the external tools (especially GoSection8) successfully filter
   * Check that if you do not fill out your profile's number of bedrooms, that the external tools (especially Craigslist) successfully filter
   * If your profile _does not_ have a voucher, check that the "Est. Max Rent" and "Search for apartments" headlines both display your profile's set number of bedrooms, but that the "Search for apartments" headline displays your set budget, not the neighborhood's max rent
   * In both cases, check that the external search tools filter for the right number of bedrooms and rent

Resolves #513

### DIFF
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -186,10 +186,10 @@ class UserProfileView(APIView):
             int(data["importanceAccessibility"])
         ]
         updated_profile.school_quality_priority = self.map_nums_to_priorities[
-            int(data["importanceAccessibility"])
+            int(data["importanceSchools"])
         ]
         updated_profile.public_safety_priority = self.map_nums_to_priorities[
-            int(data["importanceAccessibility"])
+            int(data["importanceViolentCrime"])
         ]
         updated_profile.has_voucher = data["hasVoucher"]
         updated_profile.voucher_bedrooms = data["voucherRooms"]
--- a/src/frontend/src/components/top-bar.js
+++ b/src/frontend/src/components/top-bar.js
@@ -46,7 +46,9 @@ class TopBar extends PureComponent<Props, State> {
     const query = {
       zipcode: this.props.neighborhood.properties.id,
       budget: this.props.estMaxRent,
-      rooms: this.props.userProfile.rooms,
+      rooms: this.props.userProfile.hasVoucher
+        ? this.props.userProfile.voucherRooms
+        : this.props.userProfile.nonVoucherRooms,
     };
     switch (type) {
       case "BHA":
--- a/src/frontend/src/components/neighborhood-details.js
+++ b/src/frontend/src/components/neighborhood-details.js
@@ -187,7 +187,9 @@ class NeighborhoodDetails extends PureComponent<Props> {
           {t("NeighborhoodDetails.BedroomAbbr")} ($
           {userProfile.hasVoucher
             ? estMaxRent.toLocaleString(i18n.language)
-            : userProfile.nonVoucherBudget.toLocaleString(i18n.language)}{" "}
+            : userProfile.nonVoucherBudget
+            ? userProfile.nonVoucherBudget.toLocaleString(i18n.language)
+            : estMaxRent.toLocaleString(i18n.language)}{" "}
           {t("NeighborhoodDetails.MaxRentSearch")})
         </h6>
         <div className="neighborhood-details__links">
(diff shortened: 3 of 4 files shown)


src/frontend/src/components/top-bar.js
```diff
@@ -46,7 +46,9 @@ class TopBar extends PureComponent<Props, State> {
     const query = {
       zipcode: this.props.neighborhood.properties.id,
       budget: this.props.estMaxRent,
-      rooms: this.props.userProfile.rooms,
+      rooms: this.props.userProfile.hasVoucher
```
ddohler, line 49: There are enough of these ternaries that it might be worth adding a utility function so that the logic only exists in one place. However, it feels like a borderline case to me; the logic is really simple, and I suspect that there would be a readability cost to adding a function because it would likely end up being called something like `getProfileRooms()`, which could obscure what's going on. I also don't think it would be too hard to grep for all the places we use `hasVoucher` and change them if we need to.

Anyway, that's all to say that if you haven't thought about putting these into a function (or two, because there is the budget to handle as well) it'd be worth thinking about for a few minutes, but I don't have a strong preference either way (and in fact I may have talked myself around to leaning toward keeping things as-is).

src/backend/users/views.py
```diff
@@ -186,10 +186,10 @@ def put(self, request, *args, **kwargs):
             int(data["importanceAccessibility"])
         ]
         updated_profile.school_quality_priority = self.map_nums_to_priorities[
-            int(data["importanceAccessibility"])
+            int(data["importanceSchools"])
```
ddohler, line 189: Dang, I think I missed this in code review -- sorry!

ddohler (approved):
Works great!

src/backend/users/views.py
ktohalloran: No, this was my poor copy and pasting during a small fix before merging. Can't believe I missed it 🤦‍♀️

src/frontend/src/components/top-bar.js
ktohalloran: Thanks for reviewing and for pointing this out--it hadn't occurred to me to put them in functions but I wanted to give it a try because it did feel cleaner. However, after trying it, I found that attempting to apply the same rules wherever a ternary is currently being used (if the user has no voucher or budget set, use neighborhood max rent; if the user has no rooms set, use 0) broke the external links like Zillow that were smart enough to use null values. As a result, I'm inclined to keep it as is, because it seems like I would need to add more logic to each query anyway to make it behave the way it currently does. So I'm going to leave it and merge now but, as you said, we can always replace them if we decide that makes more sense later.

PR #526, opened 2022-05-23
Conditionally show voucher question on profile page

### LINKED ISSUE #519: Fix 'invalid prop type' console error on edit profile
Currently, if you select "Edit Profile" while in the app, an error in the console appears as included below:

Screen Shot 2022-05-20 at 2 12 09 PM

We should figure out why this error is occurring and fix it.

### LINKED ISSUE #524: "Do you have a voucher" profile question should be hidden in anonymous mode
We use `voucherNumber === 'ANONYMOUS'` to detect anonymous logins, so if an anonymous user says they have a voucher they can see that value and then potentially change it, which will cause the site to get confused.

Asking for a voucher number isn't relevant for anonymous users anyway because the information will never get communicated to BHA, which is the main reason to ask the question. So we should hide the radio buttons if the user is browsing anonymously.

aaronxsu:
## Overview

This PR updates the `/profile` UI so that:
- it shows the "Do you have a voucher" question and selection radio buttons only when in anonymous mode
- a warning message won't show due to a styling related prop

### Checklist

## Testing Instructions

 * On the login UI, click on "Continue without an account"
 * Make sure on the profile form, the "Do you have a voucher" question and selection radio buttons do not show up
 * Fill out the form, save, then click on "Edit profile" from top bar to go back to the profile edit page
 * Make sure the "Do you have a voucher" question and selection radio buttons do not show up
 * Log out and log in with a non-anonymous account
 * Go to the profile edit page
 * Make sure the above question shows up with the radio buttons

Closes #524 
Closes https://github.com/azavea/echo-locator/issues/519

### DIFF
--- a/src/frontend/src/components/edit-profile.js
+++ b/src/frontend/src/components/edit-profile.js
@@ -79,7 +79,6 @@ class EditProfile extends PureComponent<Props> {
   }
 
   getDefaultState(profile: AccountProfile) {
-    console.log("profile", profile);
     if (profile) {
       // Read profile into an object for initial component state
       return {
@@ -403,7 +402,7 @@ class EditProfile extends PureComponent<Props> {
             html
             effect="solid"
             place="top"
-            offset='{"top": -10}'
+            offset={{ top: -10 }}
             isCapture
             delayHide={TOOLTIP_HIDE_DELAY_MS}
             className="map-sidebar__tooltip"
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/edit-profile.js
```diff
@@ -402,7 +402,7 @@ class EditProfile extends PureComponent<Props> {
             html
             effect="solid"
             place="top"
-            offset='{"top": -10}'
+            offset={{ top: -10 }}
```
ktohalloran, line 405: Thanks for fixing this! It looks like this fixes issue #519 so we should add a comment to that issue and drag it into review.

aaronxsu: Good call. Thanks!

ktohalloran (approved):
Looks great! 👍

Issue #525, opened 2022-05-23
Sidebar "from" destination should update on change


rachelekm:
On the main page in the sidebar, our "from" location displays the first address in our destinations array instead of our primary destination (it should be noted the origin point on our map and its pop-up is appropriately updating). In addition, when we try and make a selection to a different "from" destination that rest of the app updates, except for our "from" address. This is functioning on the demo site which also uses the `Select` component from the "react-virtualized-select" dependency (we are using the same package version).

A couple solutions were attempted as part of PR #527, including:
- setting the `value` as the `destination` obj itself instead of `destination.purpose` so that there would be unique values between `Select` options (However, this gets rid of the purpose prefix and is not translatable)
-  Pulling out the labelling logic from the `locationsWithLabels` fn into a separate `labelDestination` fn to call for the `Select` value so that the value is formatted as the 'prefix: address' label

However, this didn't solve the issue completely. While the "from" value updates to the new primary address and we can select new options in the "from" dropdown with a mouse, we can't with arrow keys and the option highlight doesn't move with each new selection. In addition, when a new option is selected, the label now appears as 'prefix: prefix: address'.
![bha_issue_525_example](https://user-images.githubusercontent.com/36374797/170505880-43313790-9874-4487-bf5a-eae7d18e0ff7.gif)

PR #527, opened 2022-05-23
Fix adding addresses in anonymous mode

### LINKED ISSUE #521: Can't add more than one address in anonymous mode
Steps to reproduce:
- Log in using "anonymous mode" ("Continue without an account")
- Add one destination to your profile
- Try to add another destination by clicking "Add another destination" and then typing in an address

Expected:
- Adding up to three addresses should work

Actual:
- When you type the first character of the address, the entire destination disappears

rachelekm:
## Overview

This PR adjusts updating `userProfile` when adding destinations so that the destination fields no longer disappear when trying to add more than one in anonymous mode. In addition, I noticed that after selecting a new primary destination, sometimes the origin on the main page updated but not the "from" field in the sidebar. I created issue #525  for it ~~and resolved it here, as well~~.

### Checklist

### Notes

- I noticed if I had more than one destination with the same type of purpose (multiple "work" destinations, for example) the "from" address on the main page sidebar would not update to the right primary address and I could not select a different destination either. The `value` for the `Select` component was set to the destination's `purpose`, which wouldn't update with multiple "work" addresses because the `value` wasn't unique between these selections. Instead, I set the `destination` obj itself to `value` to resolve this issue. UPDATE: see discussion below this was moved to a separate issue.

## Testing Instructions

 * `scripts/server`
 * Login to the site as an anonymous user
 * Confirm you can add more than one address to profile and it persists until logout

Resolves #521

### DIFF
--- a/src/frontend/src/components/edit-profile.js
+++ b/src/frontend/src/components/edit-profile.js
@@ -231,12 +231,12 @@ class EditProfile extends PureComponent<Props> {
       purpose: DEFAULT_PROFILE_DESTINATION_TYPE,
     };
     this.setState({ destinations: [...destinations, newAddress] });
-    if (!this.state.isAnonymous) {
-      this.props.setProfile({
-        ...this.getProfileFromState(),
-        destinations: [...destinations, newAddress],
-      });
-    }
+    // Despite the name, this saves the profile to local storage in anonymous mode
+    // before calling setProfile
+    this.props.handleAuthChange({
+      ...this.getProfileFromState(),
+      destinations: [...destinations, newAddress],
+    });
   }
 
   deleteAddress(index: number, event) {


ddohler (changes requested):
This works but it seems to introduce a change in the way the `from:` menu behaves after saving the user's profile.

On `develop`, after saving the user's profile, the primary destination is displayed with its Purpose prefix:
Screen Shot 2022-05-25 at 10 07 12

But on this branch, after saving the profile, the primary address is missing the purpose prefix:
Screen Shot 2022-05-25 at 10 06 55

Changing the dropdown to something else does work, and the prefix is displayed after doing so. I think we need to be using something different as the value, either the destination's index within the destination array, or its full `${purpose}: ${label}` string.

rachelekm:
Uh oh, the dropdown options showed the purpose and I totally missed that the actual value didn't include it. To keep the 'prefix: address' label consistent and translatable, I ended up pulling out the labelling logic from the `locationsWithLabels` fn into a separate `labelDestination` fn to call for the `Select` value, too. It should function as `develop` does now.

ddohler:
Hmm, so translation and the labeling work correctly now, but now it seems like we're maybe back to where we started because if I give two or more destinations the same purpose, I can only select one of them, which I think is the issue that motivated #525 in the first place? If this seems like it's going to blow up in complexity, we could potentially kick #525 back to To-Do (and increase its estimate) and just merge the address-adding fixes here.

rachelekm:
Oh interesting, thank you for catching this. If I give two or more destinations with the same purpose, the "from" updates to the new primary destination and I'm able to select the other destination with a mouse but not through arrow keys and the label includes the purpose twice for me: 
bha_issue_525_example

I agree this is less of a one-line simple fix than I thought at the beginning so I'll undo my last two commits related to solving 525, adjusting 525 as you just described, and just merge the address-adding fixes here.

ddohler:
Oh that's interesting! I didn't realize you could actually click on it--I just noticed that the highlight doesn't move over it and assumed that meant that clicking wouldn't work, but apparently it does. So lots going on with this issue 🙂 -- I agree that working on it separately is the right move.

rachelekm:
Alas, I forgot to test with arrow keys! I'll be sure to update the issue to note the highlight isn't adjusting to the new destination value as part of selecting a new option, too. Thanks!

ddohler (approved):

PR #528, opened 2022-05-24
Update integer fields handling in the user profile PUT endpoint

### LINKED ISSUE #522: Can't remove profile rent budget from either front or backend

aaronxsu:
## Overview

This PR updates the handling of the integer fields in the update endpoint of user profile.

### Checklist

## Testing Instructions

 * Spin up local app and server
 *  Log in with your email account
 * Go to profile edit page
 * Populate the integer fields, make sure submitting the form works
 * Go to profile edit page again
 * Remove values from the integer fields, make sure submitting the form still works
 * Go to the admin site, and sign in as an admin
 * Go to the edit page of a user's profile
 * Populate the integer fields, make sure submitting the form works
 * Go to the edit page of a user's profile again
 * Remove values from the integer fields, make sure submitting the form still works

Resolves #522

### DIFF
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -150,7 +150,7 @@ class UserProfileView(APIView):
     def process_nullable_int(self, data, field):
         try:
             result = int(data[field])
-        except ValueError:
+        except (ValueError, TypeError):
             # field is something that cannot be converted to
             # a number. It could be an empty string, or a
             # string like 'hello'
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -146,6 +146,17 @@ class UserProfileView(APIView):
         content = self.repackage_for_frontend(serializer.data)
 
         return Response(content)
+    
+    def process_nullable_int(self, data, field):
+        try:
+            result = int(data[field])
+        except ValueError:
+            # field is something that cannot be converted to
+            # a number. It could be an empty string, or a
+            # string like 'hello'
+            # insert null in this case
+            result = None
+        return result
 
     @transaction.atomic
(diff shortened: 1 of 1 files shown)


src/backend/users/views.py
```diff
@@ -147,6 +147,17 @@ def get(self, request, **kwargs):
 
         return Response(content)
 
+    def process_nullable_int(self, data, field):
+        try:
+            result = int(data[field])
+        except ValueError:
```
ktohalloran:
```suggestion
        except (ValueError, TypeError):
```

Catching the `ValueError` helped part of the time for me, but anytime I tried to add `nonVoucherBudget` but not both `voucherRooms` and `nonVoucherRooms` or vice versa, it failed with a `TypeError` and the message "int() argument must be a string, a bytes-like object or a real number, not 'NoneType'".

ktohalloran (approved):
Looks and works great! Just one issue as described below, but can be merged once you've addressed it.

src/backend/users/views.py
aaronxsu: Ah good call. Thanks!

Issue #529, opened 2022-05-24
Get signup / login fully working in staging


ddohler:
Once #485 is finished, we should expect that emails will successfully be sent from staging to allow people to log in, but it's likely that there may be other things that need to be done to get it fully working. This is a card to cover any such tasks.

PR #533, opened 2022-05-26
Paramaterize magic link email 'from' domain

### LINKED ISSUE #485: Parameterize email domain in magic link email
Following completion of #442,  the magic link emails come from a dummy domain. This needs to be replaced and parameterized based on the environment from which the email is coming, e.g. emails come from @stg.echosearch.org in staging and from the production domain in production.

ktohalloran:
## Overview

This PR changes the domain from which the magic link email is sent based on the environment by setting it as an environment variable in terraform which is passed down from `terraform.tfvars` in S3.

### Checklist

## Testing Instructions
_Because staging login isn't sending emails yet, this PR was a hard one to test (and write testing instructions for)._ 

 * Test the the default behavior is working as expected by starting your server locally and clicking sign in
 * Verify that the email it sends is coming from "noreply@stg.echosearch.org"
 * Now checkout a test branch based on this branch in order change a few things to test on staging. In `users/views.py`, add the following under except `User.DoesNotExist: pass` on lines 59-60:
```     
except Exception:
        return Response(data=settings.DEFAULT_FROM_EMAIL)
```
This will trigger with whatever error is currently happening on staging when attempting to access the `LoginPage` view, allowing us to see what value terraform is passing down without needing to see the email itself.
 * Remove the default value from `DEFAULT_FROM_EMAIL` in `settings.py` so that you can check the value is being correctly handed down from `terraform.tfvars`
 * Push your test branch up and once it has built, open your Network tab and then click sign in. No email will be sent, but your login request should display "noreply@stg.echosearch.org" as its response

Resolves #485

### DIFF
--- a/deployment/terraform-django/app.tf
+++ b/deployment/terraform-django/app.tf
@@ -116,6 +116,7 @@ resource "aws_ecs_task_definition" "app" {
     gunicorn_workers = ceil((2 * (var.fargate_app_cpu / 1024)) + 1)
 
     django_secret_key      = var.django_secret_key
+    default_from_email     = var.default_from_email
     django_log_level       = var.django_log_level
     r53_public_hosted_zone = var.r53_public_hosted_zone
 
@@ -187,6 +188,7 @@ resource "aws_ecs_task_definition" "app_cli" {
     postgres_db       = var.rds_database_name
 
     django_secret_key      = var.django_secret_key
+    default_from_email     = var.default_from_email
     django_log_level       = var.django_log_level
     r53_public_hosted_zone = var.r53_public_hosted_zone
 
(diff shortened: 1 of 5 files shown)


ddohler (approved):
Works great! I didn't test all the way through onto staging because I didn't want to interrupt work on #529, but I did test adding a different `DEFAULT_FROM_EMAIL` to my `docker-compose.yml` and confirmed that it was used correctly.

Now that we've gotten emails working from staging (I just got one 🎉 ), an easy way to confirm that this is working would be to deploy a test branch that tells it to send from some other email address and confirm that it either sends correctly or breaks (more likely), either one of which would confirm that the terraform variable is getting passed through.

ktohalloran:
Ok, just tested and if you give it a non-verified 'from' address, the server responds with 500 due to the email not being verified, so it looks like terraform and SES are working together properly 🎉  Merging now!

PR #535, opened 2022-05-26
bug fix swap lat lon in user profile migration script

### LINKED ISSUE #510: Profile import swaps lat/lon for points
It looks like the latitude / longitude values are getting swapped for locations imported from S3 profiles:

Screen Shot 2022-05-18 at 10 59 34

lukemckinstry:
## Overview

This PR adjusts the management script  `migrate_housing_seeker_data` used to import user profiles from S3. In the initial migration the latitude / longitude values for locations were swapped. This script has already been run to replace the user profiles on staging.

### Checklist

### Demo

echo-issue-510-fix

## Testing Instructions

- `./scripts/update`
- `AWS_PROFILE=<your echo profile name>./scripts/manage migrate_housing_seeker_data -b <staging profile bucket name> -p public -vkts ""`
- The above should work and should insert related staging profiles as records in your local DB. 
- Make sure it has an output of summary. User profiles with locations should be rendered on map correctly in admin.
- `AWS_PROFILE=<your echo profile name>./scripts/manage migrate_housing_seeker_data -b <production profile bucket name> -p public -vkts ""`
- The above should work and should insert related production profiles as records in your local DB
- Make sure it has an output of summary. User profiles with locations should be rendered on map correctly in admin.

Resolves https://github.com/azavea/echo-locator/issues/510

### DIFF
--- a/src/backend/users/management/commands/migrate_housing_seeker_data.py
+++ b/src/backend/users/management/commands/migrate_housing_seeker_data.py
@@ -391,7 +391,7 @@ class Command(BaseCommand):
             destination = Destination.objects.create(
                 profile=user_profile,
                 label=label,
-                location=Point(x=lat, y=lon),
+                location=Point(x=lon, y=lat),
                 purpose=purpose,
                 primary_destination=des[DestinationKeys.PRIMARY],
             )


aaronxsu (approved):
Luke demoed this in a pairing on Tuesday. This fix worked very well on staging as we tested. Good to go after CI is happy!

src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -36,8 +36,9 @@ class LocationKeys:
 
 
 class PositionKeys:
-    LON = "lon"
-    LAT = "lat"
+    # swap lon & lat fields to account for mislabled profile data
```
ddohler: This does fix the problem but I don't think that this is an accurate characterization of the issue here -- if you look at the latitude and longitude values in one of the profiles, they do match the real-world latitudes and longitudes of the Boston area. However, on [line 394](https://github.com/azavea/echo-locator/blob/develop/src/backend/users/management/commands/migrate_housing_seeker_data.py#L394) when we're creating the geometry, we're doing it with `Point(x=lat, y=lon)`, which isn't correct (latitude is North/South, or the Y direction, while longitude is East-West, or X). I think I probably missed this in PR review, so apologies to @aaronxsu ! But in any case, I think changing to `Point(x=lon, y=lat)` is the right way to go here.

aaronxsu: @ddohler I think I see where the confusion may come from. So actually on [these lines](https://github.com/azavea/echo-locator/blob/develop/src/backend/users/management/commands/migrate_housing_seeker_data.py#L384-L385), `lat` and `lon` pull from the corrected property keys from the user profile json object, since they were flipped in this `PositionKeys`. 

An alternative fix would be to not flip the `PositionKeys`, but rather flipping how we pass arguments when creating `Point`.

aaronxsu: I do think for better readability and for future us to reason the semantic of these variable names, maybe flipping how we create `Point` is better than flipping how we handle original keys in the object.

ddohler: > An alternative fix would be to not flip the PositionKeys, but rather flipping how we pass arguments when creating Point.

Yep, I think that's the right way to solve this! :+1:

lukemckinstry: The latest commit I forced push to this PR implements the change as discussed by reversing the inputs when the `Point` is created. Also I rebased this branch on top of the latest version of `develop`.  We successfully performed the migration on staging again using this new commit.

aaronxsu: Good job. I believe this should be good to go!

ddohler (approved):

ddohler:
@lukemckinstry It looks like this is ready to go -- can it be merged?

lukemckinstry:
@ddohler yes it can be merged

PR #536, opened 2022-05-26
Update Sidebar "From" destinations Select

### LINKED ISSUE #525: Sidebar "from" destination should update on change

rachelekm:
## Overview

This PR updates the `Select` component that creates the dropdown of destinations for the "From" field on the main page sidebar. The value that displays to the user now updates to the correct primary destination, formatted as "purpose: address", and the other destination options are now able to be selected. All options are also able to be translated. In addition, the origin pop-up that displays on the map should now show a translatable "purpose: address" label.

### Checklist

### Notes
This ended up being a trickier issue for a couple reasons:
- Once I set the option's value to a more unique identifier (`position`) the "purpose: address" formatting was lost for the value on page load, though not for the dropdown options since those values are created separately. However, if I added the formatting to the value, once an option was selected it would update to "purpose: purpose: address" since the options labels had already been formatted.
- While the destination purpose would display after selecting an option from the dropdown, the translation became static in the value and would not longer update until selecting another option. I _think_ this is because updating the `Select` value to a more unique identifier means that we're referencing the `destination` state's `label` property instead of referencing the option's `label`, the former of which has already resolved the translation into a string?

For these two reasons I ended up creating a new `purpose` property on the `destination` object so that, after an option is selected, purpose and address are re-separated from `label` so that they can be put back together again with translation and correct formatting when defining the `Select` value that displays. If `purpose` doesn't exist, it's because the `destination` state is the original object from `destinations` vs the formatted `location` options so the purpose is stored in `destination.value`.

## Testing Instructions

 * `scripts/server`
 * Login as user or anonymous
 * Create a couple destinations of the same purpose
 * Save and confirm the value that displays in the "From" field is the current primary address
 * Confirm the purpose translates
 * Confirm you can select all other destinations (and that destinations in dropdown translate as well)

Resolves #525

### DIFF
--- a/src/frontend/src/components/form.js
+++ b/src/frontend/src/components/form.js
@@ -146,6 +146,7 @@ class Form extends React.PureComponent<Props> {
         label: d.location.label,
         position: d.location.position,
         value: d.purpose,
+        purpose: d.purpose,
       };
     });
     const locationsWithLabels = locations.map((loc) => {
@@ -160,14 +161,7 @@ class Form extends React.PureComponent<Props> {
     const selectDestination = (option?: ReactSelectOption) => {
       // To make translations dynamic in Select value, reseparate label and purpose
       const loc = option && locations.find((loc) => loc.position === option.position);
-      const destinationObj = option
-        ? {
-            label: loc.label,
-            purpose: loc.purpose,
-            position: lonlat(option.position),
-            value: loc.value,
-          }
-        : null;
+      const destinationObj = option ? { ...loc, position: lonlat(option.position) } : null;
       this.setState({ destination: destinationObj });
       this.props.updateOrigin(destinationObj);
     };
@@ -200,10 +194,7 @@ class Form extends React.PureComponent<Props> {
             wrapperStyle={SELECT_WRAPPER_STYLE}
             value={{
               ...destination,
-              label:
-                t(`TripPurpose.${destination.purpose ? destination.purpose : destination.value}`) +
-                ": " +
-                destination.label,
+              label: t(`TripPurpose.${destination.purpose}`) + ": " + destination.label,
             }}
           />
         </div>
--- a/src/frontend/src/components/form.js
+++ b/src/frontend/src/components/form.js
@@ -116,6 +116,7 @@ class Form extends React.PureComponent<Props> {
       ? {
           label: destination.location.label,
           position,
+          purpose: destination.purpose,
           value: position,
         }
       : null;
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/form.js
```diff
@@ -186,11 +194,17 @@ class Form extends React.PureComponent<Props> {
...
-            onChange={this.selectDestination}
+            onChange={selectDestination}
             placeholder={t("Geocoding.StartPlaceholder")}
             style={SELECT_STYLE}
             wrapperStyle={SELECT_WRAPPER_STYLE}
-            value={destination.value}
+            value={{
+              ...destination,
+              label:
+                t(`TripPurpose.${destination.purpose ? destination.purpose : destination.value}`) +
```
ddohler: I think this ternary can be avoided by adding `purpose: d.purpose` on [line 148](https://github.com/azavea/echo-locator/pull/536/files#diff-bbddc51addecf1196a0906c70a4a81c4686a5584d94b3e5dfaf18fb949ce8b8cR148) (at least, it worked in my quick testing). In other words, it didn't make sense to me why `purpose` would ever be missing, since it's present in the API response, and I think it's just getting dropped as part of some of the earlier transformations here, so a simpler option is to make sure that we preserve it, rather than filling it in from `value` if it's missing.

I didn't totally trace it through, but I suspect that making this change here will get pushed down to `map.js` via props as well; I removed the ternary from there and didn't notice any problems.

ddohler (approved):
I think that this can be simplified somewhat but it's definitely functional as-is; thanks for working through this tricky bug!

src/frontend/src/components/form.js
rachelekm: Yes, thank you! This worked well on my end too and does makes more sense. I added the fix. 👍

PR #537, opened 2022-05-27
Add Rollbar to backend

### LINKED ISSUE #534: Add Rollbar to backend with Slack integration

BryanQuigley:
Only needed to be added to django, the terraform and keys were
already setup.

Testing plan:
Try changing default_from_email to something not verified
and then signing up should fail and error should be reported to slack.
Or some other kind of crash.

Resolves: #534

### Checklist

### Demo

See Slack.


ddohler (approved):
I saw the crash on Slack when you were testing so I didn't test that path again. Looks good!

Issue #494, opened 2022-05-10
Allow saving and editing neighborhood images


ddohler:
- There are relatively few of them, so they could be saved directly to the DB. We could also potentially save them to S3
- Users should be able to upload new images and (potentially) adjust cropping.
- We have examples of similar functionality on Cicero and on GoPhillyGo.
- Adjust the frontend so that it utilizes the new fields for neighorhood display. This may involve creating new endpoints or altering existing ones.

Once this is finished we can delete the [`neighborhood_bounds.json` and `neighborhoods.json` files](https://github.com/azavea/echo-locator/tree/develop/src/frontend/src/data).

PR #538, opened 2022-05-31
Add neighborhood models and import script

### LINKED ISSUE #492: Import neighborhood_bounds.json and neighborhoods.json into Django models
- Create models in Django matching the data fields available in neighborhood_bounds.json and neighborhoods.json
- Write a management command that populates the models using the existing GeoJSON data files.
- The models should be editable on the admin (by administrators but not counselors).
- Don't worry about neighborhood images for the time being

ddohler:
## Overview

Adds models to store neighborhood information and an import script to populate them from `neighborhoods.json` and `neighborhood_bounds.json`. Photos are not handled yet (#494). Also switches maps on the backend to use OSM rather than satellite basemaps.

### Checklist

### Demo

Screen Shot 2022-05-31 at 08 27 36
Screen Shot 2022-05-31 at 08 27 26
Screen Shot 2022-05-31 at 08 27 09

### Notes

- I chose not to do much error-handling in the import script because while the neighborhood data is ultimately human-generated, it passes through a number of scripts in order to be turned into the JSON files that the import script consumes, so those should catch most errors and leave the data in a pretty good state by the time it reaches this import script.
- The script is designed to update existing zipcodes if they exist because BHA has said that in the future, they would like to carry out data updates once per year. This also makes error correction easier because the script can be re-run multiple times and will update existing neighborhoods and/or create new ones, as needed.

## Testing Instructions

 * Run `./scripts/update` to run migrations, and then `./scripts/server`.
 * Navigate to `localhost:8085/admin/` and log in.
 * Edit a user profile and confirm that the basemap is OSM streets rather than a satellite view.
 * Run `AWS_PROFILE=echo-locator ./scripts/manage import_neighborhood_bounds s3://echolocator-production-data-us-east-1/neighborhoods.json s3://echolocator-production-data-us-east-1/neighborhood_bounds.json` (the files have already been uploaded to S3). Confirm that it runs without errors and reports that all Neighborhoods were newly created.
 * Run the script again and confirm that it reports that neighborhoods were updated.
 * Spot-check some neighborhoods using the Django Admin (and maybe geojson.io) and confirm that the values, polygon, and centroid match the values from the GeoJSON.

Resolves #492


ktohalloran (approved):
Looks great! I also double checked that the addition of a new table didn't somehow throw off the `pk`s of the permissions given to Counselors in `group_permissions.json` just to make sure, which also looked good. This did inspire one possibly tangential question, though&#8212;we currently have user's favorites stored as an array of zipcodes&#8212;is there any benefit to changing that to use the `pk`s of neighborhoods now that we have this model?

rachelekm (approved):
Looks great and works as expected! The new basemaps helped me notice that if I add a new `Destination` to a `UserProfile` via the admin the location point doesn't actually update from the default. It's happening on the `develop` branch as well/seems unrelated to this PR, so I opened issue #540 for it.

ddohler:
> This did inspire one possibly tangential question, though—we currently have user's favorites stored as an array of zipcodes—is there any benefit to changing that to use the pks of neighborhoods now that we have this model?

Good question! I thought about something similar when I was setting this up, in the sense that zipcodes are required to be unique, so we could use zipcodes as the PK directly and not have a separate `id` column for the PK. I eventually decided against it because these zipcodes will be updated relatively frequently (at least once per year) and one thing that often happens in those types of cases is that we get asked to enable the client to make edits that don't "go live" immediately. If that happened then we'd probably need to have multiple `Neighborhood` objects with the same zipcode, and it'd be a lot easier to relax the unique restriction on `zipcode` and carry on using the `id` column for the PK than to switch over to a new column for the PK if we were using `zipcode` for that purpose.

Relatedly I noticed that Django seems to have decided to use an integer-ified version of the zipcode for its PKs rather than a sequential ID starting at 1 as it might otherwise do: if you look at the PKs you'll see that they all match the zipcodes, but with leading zeroes stripped. That seemed strange, and while I'm not sure whether that's an artifact of the import process or something that Django decided to do internally, it also didn't seem to be obviously harmful, so I decided to leave it as-is. If we ever do get asked to enable "unpublished" changes to Neighborhoods then we can address the issue at that point.

But by the same token, I think we do want to continue to store favorites as zipcodes, at least until something pushes us in a different direction. Both it and `id` are required to be unique, so there shouldn't be a difference in lookup cost for the time being, and if we ever do end up in a world where we can have published and unpublished Neighborhoods that share the same zipcode, favorites will continue to work as-is if they use zipcodes but we'd need to migrate the references every time we published a new `Neighborhood` if we were using `id` for the favorites.

PR #539, opened 2022-05-31
Fix final bugs in signup / login flow

### LINKED ISSUE #529: Get signup / login fully working in staging

ktohalloran:
## Overview

This PR addresses the final issues that were preventing staging's login and signup flow from working, namely fixing the following issues:
* Magic link emails were always being sent with a "http" protocol, which resulted in broken links when clicked
* Any edits made to a user's profile were erased if they edited a destination, particularly during user sign up
* Destinations changed or added were not being restored to their previous state if users clicked "cancel"

### Checklist

### Notes

This PR also included investigating 500 errors being sent by the server on attempting to sign in; this ended up being an issue with migrations not having been run on staging and resulted in issue #534. 

This PR also removes `componentWillReceiveProps` from the `EditProfile` component, as it was causing a very difficult to debug disconnect between component and app state when a user would edit an address. Initially, `componentWillReceiveProps` was necessary because it was possible for the component to be handed a profile that was null; however, our new login process means profile should never be empty. This seemed to leave a potential problem if profile does end up null somehow, so I wanted to add some handling for that case. In order to replicate this, I cleared local storage in my application, assuming this was the only conceivable way a user could trigger this vulnerability; however, no matter what I did, I could never get it to result in an error, as the app would just use app state to re-populate local storage. So I haven't added any error handling for that case, as it seems incredibly difficult, if not impossible, to trigger.

## Testing Instructions

 * After checking that this branch is the active one on staging, try the sign up flow:
     * Verify you receive the sign up link
     * Verify that clicking the link takes you into the app
     * Verify that any fields you fill out prior to adding your first address save as they should and are not empty if you open edit profile again
* Now check that if you edit or add an address and then click cancel, those destinations are restored to their previous state and did not save your changes

Resolves #529

### DIFF
--- a/src/backend/echo/settings.py
+++ b/src/backend/echo/settings.py
@@ -137,6 +137,8 @@ DATABASES = {
     }
 }
 
+# Set protocol for sent links like password reset or magic links
+SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 
 # Password validation
 # https://docs.djangoproject.com/en/3.2/ref/settings/#auth-password-validators
--- a/src/frontend/src/components/edit-profile.js
+++ b/src/frontend/src/components/edit-profile.js
@@ -58,6 +58,9 @@ class EditProfile extends PureComponent<Props> {
     this.validDestinations = this.validDestinations.bind(this);
 
     const profile = props.userProfile;
+    // this makes a deep copy of the user's profile with which to restore all
+    // changed fields on cancel
+    this.initialProfile = JSON.parse(JSON.stringify(profile));
     this.state = this.getDefaultState(profile);
   }
 
@@ -127,6 +130,7 @@ class EditProfile extends PureComponent<Props> {
   cancel(event) {
     // Navigate back to the last page visited, discarding any changes.
     if (this.props.location.state && this.props.location.state.fromApp) {
+      this.props.setProfile(this.initialProfile);
       this.props.history.goBack();
     } else {
       // User navigated to this page directly
--- a/src/frontend/src/components/edit-profile.js
+++ b/src/frontend/src/components/edit-profile.js
@@ -61,17 +61,6 @@ class EditProfile extends PureComponent<Props> {
     this.state = this.getDefaultState(profile);
   }
 
-  componentWillReceiveProps(nextProps) {
-    // Listen for when profile to appear on props, because it is not present
-    // on initial load. Only load once by checking state.
-    if (!nextProps.isLoading && nextProps.userProfile) {
-      if (!nextProps.userProfile.destinations || !nextProps.userProfile.destinations.length) {
-        nextProps.userProfile.destinations = [Object.assign({}, firstAddress)];
-      }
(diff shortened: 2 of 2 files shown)


BryanQuigley (changes requested):
Login works fine, destinations do the right things on save/cancel.

However most other fields save on cancel.
1. Add ABC to your name.
2. Click Cancel.
3. Go Back to profile, still see ABC.

Testing with develop this doesn't happen.

ktohalloran:
Thanks for reviewing and for catching that @BryanQuigley! I had been trying to only save a copy of destinations going into a profile edit but it seems there are just too many ways for state in the component, app, and local storage to diverge. So I've changed it to just copy and then restore the entire profile on cancel. I'm not convinced that's the most elegant way to handle it, but it seems to be the one that best guarantees state stays consistent across all parts of the app.

BryanQuigley (approved):
Latest commit makes cancel work perfectly in my testing.

ddohler (approved):
Works great! I think the commits could use a little cleanup here before merging:
- I'd squash b4ada8ead962bd9fca7b111b00272518557b9d4f and c975d4b15268d27ad469b5f329c3ce7633bf6590 together (or maybe just delete b4ada8ead962bd9fca7b111b00272518557b9d4f entirely?)
- I'd turn the formatting commit (474e189fcc369fee2553def7e4dd368a6adcbadf) into fixups and then squash them down with the logical changes. You could split it but it might be easier to remove the commit, re-run `./scripts/format` and then add separate fixup commits for each file, and then rebasing with `--autosquash`.

That should leave us with just the three logical commits:
1. Changes to magic link protocol
2. Fix profile field reset when editing addresses
3. Deep copy profile for restoration when canceling.

src/backend/users/views.py
```diff
@@ -26,8 +26,12 @@ def send_login_link(request):
     user = User.objects.get(username=email)
     login_token = utils.get_query_string(user)
     host = request.get_host()
-    protocol = "https://" if request.is_secure() else "http://"
-    login_link = protocol + host + reverse("obtain_token") + login_token
+    protocol = (
```
ddohler: Not surprisingly (because Django is both broad and deep), I just discovered that Django has a built-in feature that looks like it should make this quite a bit easier: https://docs.djangoproject.com/en/4.0/ref/settings/#secure-proxy-ssl-header

It looks like we can reconfigure our `settings.py` to specify `SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')`, and then `is_secure()` should start to work as expected again.

I noticed this because the password reset links for the Django admin are exhibiting the same issue with request protocol, and this is apparently the suggested fix, but it should work for our purposes too! Sorry for the late breaking change after approval, but I think this would be worth investigating while this is still in flight, otherwise we'll have to open a separate issue for the Admin password resets not working and possibly reverse these changes.

PR #545, opened 2022-06-07
Remove voucher number field

### LINKED ISSUE #542: Remove voucher number field
BHA has decided that it's confusing for both housing seekers and counselors. They can use other fields to figure out who a person is.

We should remove the voucher number field from both the frontend profile editing form and from the backend model.

We should _keep_ the "Do you have a voucher?" question.

aaronxsu:
## Overview

This PR:
- updates the `UserProfile` data model so that the `voucher_number` field is removed
- adds a migration as a result of the above data model change
- updates the Django backend views, serializers, and admin to remove this field
- updates the housing seeker migration script so that `voucher_number` field is removed and `has_voucher` is set to `True` for all housing seekers
- updates profile edit page on frontend so that the voucher number question is removed

### Checklist

### Notes

There is a `voucherNumber` field on frontend, which is now only used for telling if the mode is anonymous or not. Posting or updating with the field in the payload body is no-op, which is fine.

## Testing Instructions

 * `./scripts/update` and then `./scripts/server`
 * The following items are aimed to test 3 scenarios: existing user, anonymous user, and new user
 * Log in with an existing user account. Go to the profile edit page. Make sure that the voucher number question is gone if you select "Yes" to "Do you have a voucher" question. Make edits to the form field values. Open network traffic tab. Make sure form submission works. Make sure the payload and response of the `/user` endpoint look fine.
 * Log in under anonymous mode. On the profile edit page, fill out the form. Open network traffic tab. Submit the form. Make sure the payload and response of the `/user` endpoint look fine. Go to the profile edit page. Test similarly and make sure it should still work. 
 * Create a new account. In the server logs, copy the link and use it to open the app. On the profile edit page, make sure the form submit correctly. Going back to the profile edit page and editing/submitting the form should also work.
 * If you have ingested user profile from production to your local DB before, go to the Admin UI and remove those users.
 * At the root of the directory, test the following command to make sure that the housing seeker migration script still works:
```
AWS_PROFILE=<your echo profile name> ./scripts/manage migrate_housing_seeker_data -b <production profile bucket name> -p public
```

Resolves https://github.com/azavea/echo-locator/issues/542


src/backend/users/management/commands/migrate_housing_seeker_data.py
```diff
@@ -315,7 +307,7 @@ def save_user_profile(self, user, profile, key):
                 self.stdout.write("User profile to be inserted is newer, updating...\n")
                 existing_profile = UserProfile.objects.get(user=user)
                 existing_profile.full_name = f"{user.first_name} {user.last_name}"
-                existing_profile.has_voucher = has_voucher
+                existing_profile.has_voucher = True
```
ddohler, line 311: Yeah this makes sense -- it was impossible to sign up for the old system without a voucher.

ddohler (approved):
Works great!

src/backend/users/management/commands/migrate_housing_seeker_data.py
aaronxsu: Agreed!

Issue #544, opened 2022-06-02
Change all user-visible mentions of "voucher" to "Boston Housing Authority voucher"


ddohler:
Apparently there are multiple types of vouchers that people can hold and it can be confusing for people to know which voucher if you just say "voucher". This should apply anywhere that's visible to any type of user, either on the frontend or the Django admin. We don't need to make internal code changes to variable or field names -- just changing labels (i.e. the copy) is sufficient.

Issue #532, opened 2022-05-25
Translations cleanup


ddohler:
We've added new text as part of the Phase II work. We'll need to send those strings to BHA for translation. This task is to assist with gathering the strings together (e.g. as a spreadsheet) so that we can send them to BHA, and to insert the new translations once received.

ktohalloran:
@ddohler what should we do about the language on the login page and magic link email? Are these phrases something we should get translations for now for in case we get to set up later, or should this list be limited to the phrases on pages inside the app? In a similar vein, does this card include setting up translation where it's needed (for example, the "logout" phrase in the header is currently not connected to `react18next` and would need to be connected and translated)?
EDIT: I realized that if you set the language in the app and then log out, that language preference is carried out onto the login page, at least until refresh. So it may be worth getting those translations just in case that happens, though that does seem to be a very specific case.

ddohler:
@ktohalloran Good questions! The scope of this task is to make sure that any strings that appear on a screen that has a language switcher are translated accurately if the user switches the language on that screen. So that includes:
1. Strings that are already translated but that we've changed on those screens (like the "voucher" -> "Boston Housing Authority voucher" case).
2. Strings that we've added on those screens (like "Logout")

But keeping the budget in mind, I would avoid translations on any screens that don't display a language switcher. So that includes the login screen and the sign-in emails; we don't need to translate those until/unless we add a language switcher to the login screen.

PR #546, opened 2022-06-07
Edit voucher copy and default selection

### LINKED ISSUE #543: Default "Do you have a voucher?" question to Yes
This should apply in both the backend and frontend code.

rachelekm:
## Overview

This PR updates the default "Do you have a voucher?" field to “yes” on the frontend (`True` in admin) as well as updates any user-facing copy that includes “voucher” to specify “Boston Housing Authority voucher” instead. 

### Checklist

### Notes

- I decided to leave any translations that include "voucher" alone. I _could_ pop in the "BHA" label but I think I better keep my hands off until someone can formally translate.
- There's some copy that reference "Client ID" that I think was used for the search page, but is no longer used elsewhere in the app (`Accounts.InvalidVoucherNumber`, `Accounts.MissingVoucherNumber`, `Accounts.Voucher`). I switched it to "Boston Housing Authority voucher" to it's consistent but maybe worth just deleting these lines altogether?

## Testing Instructions

 * `scripts/update`
 * `scripts/server`
 * Navigate to [frontend site](http://localhost:9966/) and create a new user. Confirm "Do you have a voucher?" field defaults to “yes”.
 * Work way through the site to confirm any copy that included "voucher" now specifies "Boston Housing Authority voucher". From what I can find this is mostly limited to the edit profile page for new and verified users. 
 * Navigate to [admin](http://localhost:9966/) and create a new `UserProfile`. Confirm the "Has a Boston Housing Authority voucher" field defaults to "checked".
 * Confirm any other fields on the admin that include "voucher" are now "Boston Housing Authority voucher".

Resolves #543 #544


ktohalloran (approved):
This looks good! I think you're right that just trying to add "BHA" into the previously translated text is probably not very helpful; however, I think leaving the translations as is will make them difficult to find again for translation. So I think somehow marking those&#8212;whether by replacing them with their English counterparts for now or adding a comment by each that links to #532 or adding a list of them to the issue itself&#8212;just something that makes it obvious they're going to need new translations when we get to that step. 

As for the strings in the translation files that are no longer used--@ddohler would it be worth adding clean up of those files, removing strings that are not in use within the app anymore, to issue #532?

ddohler:
@ktohalloran Good call about translatios -- I think adding a list to #532 is the right way to go. I would probably skip cleaning up the unused translations given the budget; they're messy but harmless. We should make an issue to clean them up later though.

ktohalloran:
Sounds good, I can make that issue now while I'm thinking about it.

rachelekm:
Thanks, all! @ktohalloran thanks for making those issues and good point about recording the missed translations for later. For #532, Here's a table outlining which phrases related to this PR have updated copy but aren't translated and others that are translated but have outdated copy and need to be re-translated: 
|  | ES | ES | ZH | ZH |
| --- | --- | --- | --- | --- |
||Translated|Updated| Translated|Updated|
| SignIn.Greeting | X | | X | |
| Profile.CreateClientAccountExistsError | X | | | X |
| Profile.HasVoucher |  | X | | X |
| Profile.VoucherNumber |  | X | | X |
| Profile.VoucherRooms | X | | X | |

And then the ones I found related to this PR that will need to be dropped for #548 are:
Accounts.InvalidVoucherNumber
Accounts.MissingVoucherNumber
Accounts.Voucher
Profile.ChooseVoucher

PR #550, opened 2022-06-08
Neighborhoods API

### LINKED ISSUE #493: Migrate frontend to use neighborhood data from Django
Create API endpoints to serve neighborhood data, and switch the frontend so that it uses the new endpoints. These can be read-only, the only expected editing of neighborhoods is via the admin and/or scripts.

ddohler:
## Overview

Pulls neighborhood data from the Django API rather than including it in the application bundle.

### Checklist

### Demo

N/A -- Everything should look the same.

### Notes

- I have had occasional issues with the neighborhoods showing up with strange corrupted polygons, e.g.:
Screen Shot 2022-06-08 at 10 47 05
I suspect the issue is some kind of race condition, so I'm hoping it's limited to my particular environment, but please be on the lookout for it and let me know if you encounter it.

## Testing Instructions

 * Before checking out this branch, run `./scripts/server` on `develop` and log in to the app. Take note of the neighborhood recommendations you are provided, including their routing instructions, school quality, and max rents.
 * Check out this branch and run `./scripts/update` followed by `./scripts/server`
 * Log in and log back out and confirm that you see network requests to `/api/neighborhoods/` and `/api/neighborhood-bounds/`. You may see multiple requests -- this is something we'd ideally eliminate because it indicates there are superfluous re-renders occurring in the app, but we don't have time to address that at the moment. I'll open a tech-debt issue for it.
 * Confirm that the neighborhood recommendations and attributes are the same as the ones on `develop`, except that neighborhood images will not work (this is expected and will be addressed in #494 ).

Resolves #493


src/backend/api/serializers.py
```diff
@@ -0,0 +1,34 @@
+from rest_framework import serializers
+from rest_framework_gis import serializers as gis_serializers
+
+from .models import Neighborhood
+
+
+class NeighborhoodSerializer(gis_serializers.GeoFeatureModelSerializer):
```
ddohler, line 7: I initially investigated combining these two into a single serializer and endpoint so that we can only make a single API request, but it quickly became clear that that approach would have required additional surgery on the frontend, so I ended up creating endpoints that mimic the format of the existing data files.

src/backend/api/views.py
```diff
@@ -1,3 +1,24 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from .models import Neighborhood
+from .serializers import NeighborhoodBoundsSerializer, NeighborhoodSerializer
+
+
+class ListNeighborhoods(APIView):
+    permission_classes = (IsAuthenticated,)
```
ddohler, line 10: There's nothing technically private about this data (and indeed, the existing data is publicly available in this repo) but I didn't want someone being able to access the API without signing up for an account in case someone does something strange and potentially problematic like writing a bot to scrape the data.

rachelekm (commented):
Looks good! Even though it makes the api call multiple times because of the `Authenticator` re-render on login, I think its the best place to put the `getNeighborhoods` dispatch for now and works well enough. One point of clarification is that I have some discrepancies in the neighborhood properties that display between branches. Taking a closer look, the data in the admin matches the files on s3 and displays correctly through the api, it just differs from my local `neighborhoods.json`. I'm assuming the neighborhood properties on the s3 file were updated and the file we have is outdated? Some examples:
**Boston - north end**
develop: 1br max rent = $2,275; here: 1br max rent = $2,400

**Boston - Seaport**
develop: 1br max rent = $2,575; here: 1br max rent = $2,600

**Norwood** 
develop: 1br max rent = $1,725; here: 1br max rent = $1,700

If I'm just missing that the files have been updated, then I think this is good to go! I also tried playing around to try and replicate the polygons issue and haven't had any issues yet:
Screen Shot 2022-06-08 at 4 40 04 PM

ddohler:
> If I'm just missing that the files have been updated, then I think this is good to go!

@rachelekm Thanks for catching this! Turns out I was the one who was missing the updated files; I had uploaded a file from my `neighborhood_data` folder that wasn't as up-to-date as the one in `src/frontend/src/data/`, which is the official source (the files in `neighborhood_data` are .gitignored). I've uploaded the right files to S3 now so everything should line up correctly with `develop` now if you re-import.

rachelekm (approved):
Thanks @ddohler, this looks good to go now! I've re-imported with the updated data and now the neighborhood recommendations and attributes are the same as on `develop` and display as expected.

PR #553, opened 2022-06-09
Drop NAT Gateway

BryanQuigley:
We can just use a public subnet, but have RDS (or anything else) not accessible via:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_VPC.WorkingWithRDSInstanceinaVPC.html#USER_VPC.Hiding

We do have an IPv4 address assigned to the ECS cluster tasks (it's free as long as it's running  From what I can tell we have default IP limit of 5 per AWS account and if some somehow don't get released properly we can incur charges at 0.005 an hour ($3.65 a month).  Total possible costs if things go poorly would be ~$14.60 (for 4 IPs) per month vs $65 guaranteed for NAT gateway.  

### Checklist

## Testing Instructions

 * Ensure sign in to app and admin works
 * Rollbar works (but can test again)

### DIFF
--- a/deployment/terraform-django/app.tf
+++ b/deployment/terraform-django/app.tf
@@ -6,7 +6,7 @@ resource "aws_security_group" "alb" {
   vpc_id = aws_vpc.default.id
 
   tags = {
-    Name        = "sg${local.short}AppLoadBalancer"
+    Name = "sg${local.short}AppLoadBalancer"
   }
 }
 
@@ -15,7 +15,7 @@ resource "aws_security_group" "app" {
   vpc_id = aws_vpc.default.id
 
   tags = {
-    Name        = "sg${local.short}AppEcsService",
+    Name = "sg${local.short}AppEcsService",
   }
 }
 
@@ -25,10 +25,10 @@ resource "aws_security_group" "app" {
 resource "aws_lb" "app" {
   name            = "alb${local.short}App"
   security_groups = [aws_security_group.alb.id]
-  subnets         = aws_subnet.public.*.id
+  subnets         = aws_subnet.main.*.id
 
   tags = {
-    Name        = "alb${local.short}App"
+    Name = "alb${local.short}App"
   }
 }
 
@@ -51,7 +51,7 @@ resource "aws_lb_target_group" "app" {
   target_type = "ip"
 
   tags = {
-    Name        = "tg${local.short}App"
+    Name = "tg${local.short}App"
   }
 }
 
@@ -123,7 +123,7 @@ resource "aws_ecs_task_definition" "app" {
   })
 
   tags = {
-    Name        = "${local.short}App",
+    Name = "${local.short}App",
   }
 }
 
@@ -141,9 +141,11 @@ resource "aws_ecs_service" "app" {
   enable_execute_command = true
   force_new_deployment   = true
 
+
   network_configuration {
-    security_groups = [aws_security_group.app.id]
-    subnets         = aws_subnet.private.*.id
+    security_groups  = [aws_security_group.app.id]
+    subnets          = aws_subnet.main.*.id
+    assign_public_ip = true
   }
 
   load_balancer {
@@ -191,7 +193,7 @@ resource "aws_ecs_task_definition" "app_cli" {
   })
 
   tags = {
-    Name        = "${local.short}AppCLI",
+    Name = "${local.short}AppCLI",
   }
 }
 
--- a/deployment/terraform-django/database.tf
+++ b/deployment/terraform-django/database.tf
@@ -4,10 +4,10 @@
 resource "aws_db_subnet_group" "default" {
   name        = var.rds_database_identifier
   description = "Private subnets for the RDS instances"
-  subnet_ids  = aws_subnet.private.*.id
+  subnet_ids  = aws_subnet.main.*.id
 
   tags = {
-    Name        = "dbsngDatabaseServer"
+    Name = "dbsngDatabaseServer"
   }
 }
 
@@ -28,6 +28,7 @@ resource "aws_db_instance" "postgresql" {
   final_snapshot_identifier  = var.rds_final_snapshot_identifier
   skip_final_snapshot        = var.rds_skip_final_snapshot
   copy_tags_to_snapshot      = var.rds_copy_tags_to_snapshot
+  publicly_accessible        = false
   multi_az                   = var.rds_multi_az
   storage_encrypted          = var.rds_storage_encrypted
   vpc_security_group_ids     = [aws_security_group.postgresql.id]
@@ -42,7 +43,7 @@ resource "aws_security_group" "postgresql" {
   vpc_id = aws_vpc.default.id
 
   tags = {
-    Name        = "sgDatabaseServer",
+    Name = "sgDatabaseServer",
   }
 }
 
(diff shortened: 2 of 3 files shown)


deployment/terraform-django/app.tf
```diff
@@ -15,7 +15,7 @@ resource "aws_security_group" "app" {
   vpc_id = aws_vpc.default.id
 
   tags = {
-    Name        = "sg${local.short}AppEcsService",
+    Name = "sg${local.short}AppEcsService",
```
ddohler, line 18: I double-checked this and it looks like it only allows inbound traffic from the load balancer, which seems right to me if we're placing the application instances in a public subnet.

deployment/terraform-django/app.tf
```diff
@@ -141,9 +141,11 @@ resource "aws_ecs_service" "app" {
   enable_execute_command = true
   force_new_deployment   = true
 
+
   network_configuration {
-    security_groups = [aws_security_group.app.id]
-    subnets         = aws_subnet.private.*.id
+    security_groups  = [aws_security_group.app.id]
+    subnets          = aws_subnet.public.*.id
+    assign_public_ip = true
```
ddohler, line 148: What happens if we set this to `false`? Right now I'm not seeing a public IP allocated to the running task, but I don't know if the attempted CI deploy may have messed things up from the way you had set them up while testing.
Screen Shot 2022-06-10 at 12 37 14

But anyway I'm curious what breaks (if anything) if this is false.

deployment/terraform-django/database.tf
```diff
@@ -4,10 +4,10 @@
 resource "aws_db_subnet_group" "default" {
   name        = var.rds_database_identifier
   description = "Private subnets for the RDS instances"
-  subnet_ids  = aws_subnet.private.*.id
+  subnet_ids  = aws_subnet.public.*.id
```
ddohler: Can the private and public subnets talk to one another? I assume they should be able to but my mental model of networking inside a VPC is a bit fuzzy. But I'm wondering if we could keep the private subnet around but just take the NAT Gateway off of it. That way we could get the cost savings but maintain availability of a private subnet for things like RDS that don't need internet access.

deployment/terraform-django/app.tf
ddohler: A further thought: it looks like we might be able to achieve the same goal while keeping the application instances in the private subnet by expanding our use of VPC endpoints. https://docs.aws.amazon.com/vpc/latest/privatelink/create-interface-endpoint.html . My thinking is that while this application in particular doesn't store sensitive information, some applications we work on do and those clients may not be comfortable with the use of public subnets, so if we're looking for a general, NAT-less solution we may want to stick with private subnets (even though I think the current solution is fine for this particular project; the security groups keep them from being wide open to the world).

The main downside to a VPC endpoint that I can see is that they're not free. However, they're a lot cheaper than NAT Gateways because they cost $0.01/hr + $0.01/GB compared to $0.045/hr + $0.045/GB. The other downside is that they're more complicated -- we'll have to pick and choose based on the services used by each project if we don't want to get charged unnecessarily.

ddohler (commented):
This is looking good! I had some questions about specific details but I think this seems like a good way to cut hosting costs.

deployment/terraform-django/app.tf
BryanQuigley: It did have Public IP previously - two things that break would be rollbar and ECR access (although that is likely fixable with VPC Endpoints)

It was not obvious to me that you could actually make VPC endpoints for rollbar or non-AWS services.

ddohler: Oops, no, I just forgot about Rollbar 🤦 So this may be a situation where there's not a good one-size-fits-all solution, because yeah I don't see a way to connect to Rollbar from inside a VPC without a NAT Gateway.

deployment/terraform-django/database.tf
BryanQuigley: That's similar to what I originally tried:
LB/ECS on public subnet
ECS/RDS on private subnet

I couldn't get the terraform to work with having ECS on both.

I just don't see the benefit to us for the increased complexity. With AWS public subnets we get effectively the ability to say on an instance by instance based if we want it to have a public IP or not.  Otherwise they only get private IPs. (Effectively we get two subnets with a public subnet)

ddohler: Okay, got it, so if I'm understanding right, a private subnet is one which:
1. Has no default route that routes to an internet gateway. Therefore outgoing access to non-private IPs is impossible.
2. May optionally have a default route to a NAT gateway, which then provides outgoing access to non-private IPs, but via NAT rather than plain old IP.

A public subnet is one which:
1. Has a default route to an internet gateway. Therefore outgoing access to non-private IPs is possible, subject to security group rules.
2. May assign a public IP to resources, which additionally allows incoming traffic, subject to security group rules.

One thing I'm confused about with the public subnets, assuming I'm thinking about it the right way, is how they're providing public internet access without either NAT or public IP addresses. I'm curious what happens if you `curl whatismyip.com` from inside an instance with no public IP? If there's a NAT Gateway then I know to expect the IP of the NAT Gateway to be reported, but I'm not sure what to expect in the case where there's only an internet gateway but there isn't a public IP address. Is AWS in fact assigning public IPs but just rotating them so that they're not static? If you discover the current IP address of a resource with no public IP address, can you then send unsolicited traffic to it until the IP address changes? (subject to the security group, of course)

If it's "they do have public IPs but not _static_ public IPs" then I can understand the recommendation to put things in a private subnet, since things in a public subnet could in that case receive incoming traffic from bots that are scanning subnets looking for vulnerable hosts and the only remaining line of defense would be a properly-crafted security group.

It'd be helpful to know if that was true, but in this case we have the security groups set up such that the app servers reject all traffic except from their own load balancer, so we should be okay.

BryanQuigley: Will test it to confirm, but an Internet gateway only works with those that have public IPs. (https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Internet_Gateway.html)

Also renamed to the subnet to be just "main" cause public/private is not actually properly descriptive given the above.

BryanQuigley: Oh, failed to test this because, without public IP it can't access ECR to pull the image it needs.  With the public IP it works.

I did find an alternative that would let us get away from needing public IPs for the containers - Egress only internet gateways using IPv6- https://docs.aws.amazon.com/vpc/latest/userguide/egress-only-internet-gateway.html.  That's likely the best long term option - but I don't think that's worth pursuing for echo.

ddohler: Cool, yeah, everything is clear now. My concluding high-level thoughts are that:
1. This is a good option for reducing costs on low-traffic, low-resource sites like this. NAT Gateways shouldn't double our hosting costs.
2. Even for low-traffic sites we may get client pushback if "security" is a concern because of the requirement for public IPs in this solution.
3. This is maybe another indication that a different provider like Digital Ocean or Heroku might be more appropriate for these types of sites?
4. I really like the sound of the IPv6 egress-only gateways and hope we can find a way to make that work on future projects because it sounds like it combines the best of both solutions (that is, it achieves low cost without requiring us to allocate static public IP addresses).

ddohler (approved):
Thanks for putting this together and answering my many questions!

Issue #557, opened 2022-06-10
Update loginMessage for users unable to log in


rachelekm:
If a User attempts to log in and has an inactive account or no account at all they will still receive the same copy on signin: "Thank you! Please check your email for a link to sign into ECHO". Since they are not receiving any email, we should update the copy to alert them that there is no active account and prompt them to create one.

ktohalloran:
@ddohler I have a vague memory of avoiding messaging like this for privacy reasons so that people couldn't figure out whether another person was using the app; is this still the case?

ddohler:
These are good questions! I think this is an area where some additional UX work would be helpful, so I think we should hold off on addressing this unless / until we can get a design contract to rethink the site's UX.

ktohalloran:
As discussed elsewhere, one possible way of handling this is to is not require people to sign up separately before logging in, and handle "attempting to log in without an account" as if the user had clicked "sign up".

PR #558, opened 2022-06-10
Add is_active login conditional

### LINKED ISSUE #549: Don't send login email if user is not active
Users can be marked inactive in the admin by unchecking the "Is active" checkbox; inactive users should not receive a login email if they attempt to log in.

rachelekm:
## Overview

This PR adds a conditional that prevents inactive users from receiving a login email.

### Checklist

### Notes

I didn't add any messaging to display on the frontend in the event an inactive user attempts to login, as discussed in RRP. On submit, an inactive user will receive the same display message as if successful. There is currently an issue open to address this in the future #557 

## Testing Instructions

 * Create a new user either through the admin or using the frontend signup form
 * Log in as the new houseseeker user and confirm you receive an email
 * In the admin, navigate to the newly created user and uncheck the "Is active" checkbox and save
 * Attempt to log in and confirm you no longer receive an email
 * Attempt to log in as a user without an account and confirm you do not receive an email (this existed before, but confirm it still works with the new conditional)

Resolves #549

### DIFF
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -54,9 +54,11 @@ def send_login_link(request):
 class LoginPage(APIView):
     def post(self, request, **kwargs):
         try:
-            # Will throw exception if cannot find the User
-            # For privacy reasons, do not send 500 error back if not found
-            send_login_link(request)
+            email = request.data["username"]
+            if User.objects.get(username=email).is_active:
+                # Will throw exception if cannot find the User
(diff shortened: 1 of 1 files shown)


aaronxsu (approved):
Tested and worked as advertised. Good to go.

As mentioned in the notes section, there is another issue to deal with the case when the user profile could not be found or the user is not active.

PR #559, opened 2022-06-21
Add neighborhood images to database

### LINKED ISSUE #494: Allow saving and editing neighborhood images

ktohalloran:
## Overview

This PR adds neighborhood images to the database for easier saving and editing. It also adds image importing to the `import_neighborhood_data` script (previously named `import_neighborhood_bounds`) and removes the now unnecessary `neighborhoods.json` and `neighborhood_bounds.json` files.

### Checklist

### Demo

The images as they are displayed in the frontend.

Screen Shot 2022-06-21 at 9 07 56 AM

Screen Shot 2022-06-21 at 9 08 14 AM

Users can edit and replace the images from the admin.

Screen Shot 2022-06-21 at 10 22 30 AM

### Notes

This PR saves the images to S3 via a custom storage backend. This set up required a few additions, including:
* adding `aws_s3_photo_bucket` environment variable to Terraform and the `terraform.tfvars` file in S3
* adding `aws_s3_photo_bucket` environment variable to `.env` and telling Django to use `.env` for local development. This change to `.env` has been updated in LastPass
* In order to guarantee local development has the required permissions without breaking STRTA, `AWS_PROFILE=echo-locator` has been added to `./scripts/server`

This PR adds two new fields for each image, an image field that allows you to update the file, and an image preview field. As a result of this, there are now a total of 9 fields per image: the original Wikipedia link, the Wikipedia thumbnail link, the image field with the link to S3, the image preview field, the image license, license URL, description, artist, and artist username. The first three of these fields seem especially redundant and all fields could probably be organized differently, but I have left them as is until the client tells us if they would prefer something else. 

Due to budget, I did not add any validation to picture size, only help text to instruct users as to what size and ratio the images will be displayed at. It seems to do fine with images that are over-sized but not with pictures that are portrait rather than landscape. I also did not add cropping to the images; this was for two reasons: 1) budget and 2) the images are downloaded and displayed at such a small size that cropping would have been difficult, anyway.

## Testing Instructions

 * Run `./scripts/update` to apply the image field migration
 * Run `AWS_PROFILE=echo-locator ./scripts/manage import_neighborhood_data s3://echolocator-production-data-us-east-1/neighborhoods.json s3://echolocator-production-data-us-east-1/neighborhood_bounds.json s3://echo-locator-production-site-us-east-1/assets/neighborhoods/` to import the images from S3
 * Verify that the images display as they should in the admin
 * Verify that the images display as they should both on the neighborhood card in the dock and in the neighborhood detail view

Resolves #494 

EDIT: Changed length of image field help text to fix spacing issues.


src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -102,17 +120,37 @@ def handle(self, *args, **options):
         bounds_data = json.load(
             s3.Object(bounds_url_parts.netloc, bounds_url_parts.path[1:]).get()["Body"]
         )["features"]
+        image_bucket = s3.Bucket(images_url_parts.netloc)
+        # Iterate through and generate list of item keys from assets/neighborhood, excluding .gitkeep
+        neighborhood_images = [
+            img.key.split("/")[2]
```
ddohler: Python has built-in functionality for this; something like the following should work
```python
from pathlib import Path
...
Path(img.key).name
```

src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -102,17 +120,37 @@ def handle(self, *args, **options):
...
+            for img in image_bucket.objects.filter(Prefix=images_url_parts.path[1:])
+            if img.key.split(".")[1] != "gitkeep"
```
ddohler: These `.gitkeep` files shouldn't be getting pushed to S3 at all. Could you open an issue for filtering these out at the deployment stage and add a TODO here?

As a suggestion, could this be simplified to `if img.key != ".gitkeep"`?

src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -102,17 +120,37 @@ def handle(self, *args, **options):
...
+        ]
         # Zip data sources together by shared unique key (zip code)
         neighborhoods_by_zip = {nhd["properties"]["zipcode"]: nhd for nhd in neighborhood_data}
-        nhd_and_bounds_by_zip = {
-            bound["properties"]["id"]: (neighborhoods_by_zip[bound["properties"]["id"]], bound)
+        imgs_by_zip = {}
+        for zipcode in neighborhoods_by_zip:
+            img_and_key = {}
+            for img in neighborhood_images:
+                if img.split("_")[0] == zipcode:
```
ddohler: This can be simplified to `if img.startswith(zipcode):`

src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -102,17 +120,37 @@ def handle(self, *args, **options):
...
+                    # save key to dictionary without extension
+                    img_and_key[img.split(".")[0]] = img
```
ddohler: We can reduce the amount of post-processing we have to do later if we can store each image by its purpose alone, without including the zipcode. Right now the dictionary being generated here looks like the following:
`{'02790_open_space_or_landmark': '02790_open_space_or_landmark.jpg', '02790_street': '02790_street.jpg'}`

But since we know there are only four types of images and the zipcode is already available elsewhere, we don't need to store that information here, and it should make later usage of this dict more straightforward if we use a format more like this:
`{'open_space_or_landmark': '02790_open_space_or_landmark.jpg', 'street': '02790_street.jpg'}`

What I'd suggest is creating a small function that takes a filename and detects what type of image it is. Something like the following:

```python
def detect_image_purpose(filename):
    for purpose in IMAGE_FIELDS:
        if purpose in filename:
            return purpose
    raise ValueError(f"Couldn't identify image purpose for {filename}")
```

Then this line could become `img_and_key[detect_image_purpose(img)] = img`.

Once that's done, I think `img_and_key` becomes a bit of a misnomer so I'd rename it to something like `images_by_purpose` or `images_by_scene` or similar.

src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -102,17 +120,37 @@ def handle(self, *args, **options):
...
+            for img in image_bucket.objects.filter(Prefix=images_url_parts.path[1:])
+            if img.key.split(".")[1] != "gitkeep"
+        ]
         # Zip data sources together by shared unique key (zip code)
         neighborhoods_by_zip = {nhd["properties"]["zipcode"]: nhd for nhd in neighborhood_data}
-        nhd_and_bounds_by_zip = {
-            bound["properties"]["id"]: (neighborhoods_by_zip[bound["properties"]["id"]], bound)
+        imgs_by_zip = {}
+        for zipcode in neighborhoods_by_zip:
+            img_and_key = {}
```
ddohler: Once the other comments in this for-block are addressed, then I think the purpose of this for-block can be made clearer by switching to a structure more like this:
```python
zip_imgs = [filename for filename in neighborhood_images if filename.startswith(zipcode)]
zip_images_by_scene = {detect_image_scene(filename): filename for filename in zip_imgs}
```

Then this can continue with `imgs_by_zip[zipcode] = zip_images_by_scene` below.

src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -60,15 +62,25 @@ def sanitize_value(key, val):
 
     # The Django model keys exactly match the keys present in the GeoJSON properties.
     sanitized_properties = {key: sanitize_value(key, val) for key, val in nhd["properties"].items()}
+    # Remove zipcodes from image file names in order to match to image fields
+    img_no_zips = {"_".join(img.split("_")[1:]): imgs[img] for img in imgs}
```
ddohler: Once the images are coming in in the format `{'open_space_or_landmark': '02790_open_space_or_landmark.jpg', 'street': '02790_street.jpg'}` it won't be necessary to do any of this processing.

src/backend/api/management/commands/import_neighborhood_data.py
```diff
@@ -60,15 +62,25 @@ def sanitize_value(key, val):
...
+    for field in IMAGE_FIELDS:
+        for img in img_no_zips:
+            if img == field:
+                identify_imgs[field] = img_no_zips[img]
     _, created = Neighborhood.objects.update_or_create(
         zipcode=nhd["properties"]["zipcode"],
         defaults=dict(
             centroid=GEOSGeometry(json.dumps(nhd["geometry"]), srid=4326),
             boundary=GEOSGeometry(json.dumps(bound["geometry"]), srid=4326),
+            street_image=identify_imgs.get("street"),
```
ddohler: Same here -- once the incoming format of `imgs` is adjusted, this can just be `street_image=imgs.get("street")`

ddohler (approved):
Works great! I left some comments relating to code organization in the `import_neighborhood_bounds` script but they're all cosmetic.

ktohalloran:
Thanks for reviewing and for helping me wrap my head around simplifying via comprehensions @ddohler! I've made the changes you recommended and will merge now.

PR #561, opened 2022-06-22
Enable cloudfront -> LB -> Container

### LINKED ISSUE #450: Ensure IPv6 works (staging)
IPv6 is enabled, but doesn't actually work.  Investigate (could just be AAAA records).

Needs Cloudfront setup, and then dns pointing to cloudfront not the LB.

BryanQuigley:
## Overview

Enable cloudfront -> LB -> Container
    
    Previously it was just LB -> Container.
    
    As part of this change:
     * Disabled HTTPS for LB, just use http-only (alternative that I've seen is use the
         same cert and have it be at origin.*, but I'm not sure what this gains us
     * Cleaned up the S3 buckets that were for a totally diferent Cloudfront setup, in this
       case we copied cdn from DistrictBuilder.
     * Dropped a few more tags that aren't useful.

### Checklist

### Notes

Right now we don't vary the caching by anything in particular.  This might be useful to improve performance if needed.

## Testing Instructions

 * Confirm site still works as expected
 * I got a "Can't find profile directory." error in console when using a non-logged in user.  Not sure if that's related to this change or not.  Works fine with a logged in user.

Resolves #450

### DIFF
--- a/deployment/terraform-django/app.tf
+++ b/deployment/terraform-django/app.tf
@@ -4,19 +4,11 @@
 resource "aws_security_group" "alb" {
   name   = "sg${local.short}AppLoadBalancer"
   vpc_id = aws_vpc.default.id
-
-  tags = {
-    Name = "sg${local.short}AppLoadBalancer"
-  }
 }
 
 resource "aws_security_group" "app" {
   name   = "sg${local.short}AppEcsService"
   vpc_id = aws_vpc.default.id
-
-  tags = {
-    Name = "sg${local.short}AppEcsService",
-  }
 }
 
 #
@@ -26,10 +18,6 @@ resource "aws_lb" "app" {
   name            = "alb${local.short}App"
   security_groups = [aws_security_group.alb.id]
   subnets         = aws_subnet.main.*.id
-
-  tags = {
-    Name = "alb${local.short}App"
-  }
 }
 
 resource "aws_lb_target_group" "app" {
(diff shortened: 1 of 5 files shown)


ddohler (approved):
Looks good to me -- everything seems to be working fine on staging. I agree that we don't gain much from having the CDN hit the load balancer over HTTPS.

I wasn't able to reproduce the issue with staging not working in anonymous mode, though one difference may be that I was already logged in and had to log out to test. If you log out and then try in anonymous mode, does it start working?

This is making me realize that the path of least resistance for getting neighborhood images working may not have led us to an optimal infrastructure setup. We are loading the neighborhood images via direct S3 links to the existing S3 bucket that serves the v1 version of the site. That was the easiest way to integrate with Django, but it means the images are bypassing the CDN, and the bucket they're currently in was created via the _old_ terraform setup, rather than the new terraform setup for Django. So to resolve that we could potentially add a neighborhood images bucket to the new terraform setup and then proxy the S3 links behind a Django endpoint so that everything flows through Django (and therefore the CDN).

However, given the current project budget I'm inclined to put that on a back-burner for now, but take a look at #559 and see if you think there's anything you think we need to change more urgently.

BryanQuigley:
We could also use the CloudFront to redirect directly to the s3 images. I'm guessing the performance differences are minimal though.

The old/new terraform is critical whenever we do https://github.com/azavea/echo-locator/issues/448 - so noted it there.

PR #565, opened 2022-06-24
Update how the app lists neighborhood data

### LINKED ISSUE #564: Logging in anonymously sometimes shows a blank map
Steps to reproduce:
- If you are logged in with a user account, log out.
- Log in anonymously and set up your profile preferences, then save so that you're directed to the map.
- Close the browser window and reopen it

Expected behavior:
- You should see the same neighborhood suggestions that you saw previously.

Actual behavior:
- The map is blank and "No results" is displayed in the sidebar.
- Additionally, logging out and logging back in anonymously does not fix the issue, but logging in with a user account and then logging out and logging back in anonymously does fix the problem, until you close the browser window again (I'm not sure what happens if you log out before closing the browser window).

Notes:
My best guess is that this is related to something about how the app handles data that exists in local storage on page load, but I haven't seen any console errors, so I'm not entirely sure.

aaronxsu:
## Overview

This PR:
- removes the auth check from the neighborhood endpoints
- lists neighborhood data when auth component mounts
- adds `AWS_S3_PHOTO_BUCKET` env var to the local dotenv file sample

### Checklist

## Testing Instructions

- `./scripts/update` 
- `./scripts/server`
- Follow the steps in the linked issue that used to trigger the empty list and map state. Make sure the bug is fixed and make sure the neighborhood list result is as expected

Resolves https://github.com/azavea/echo-locator/issues/564


src/frontend/src/components/authenticator.js
```diff
@@ -21,6 +22,11 @@ export default function Authenticator(Comp) {
       this.props.loadProfile();
     }
 
+    componentDidMount() {
+      const props = this.props;
+      props.store.dispatch(getNeighborhoods());
```
aaronxsu, line 27: This used to be called in the `setAuthToken` action, which happens right in this component render method. So I assumed that listing neighborhood should happen when this component mounts, since the related endpoints are not behind auth anymore after the change on the backend.

ddohler: Good call! Now that we don't rely on having the authentication token, we can load the neighborhoods whenever it's most convenient, and this does seem like a better spot.

ddohler (approved):
Looks great and works well! Good cleanup along with the bug fix!

PR #566, opened 2022-06-28
No AWS creds needed if not test/develop

BryanQuigley:
Test to see if we can have no need for credentials when doing a cibuild mostly run (and bootstrap/linting).


ddohler (approved):
Looks good -- I can see that the workflow skipped the credentials step that is tripping up Dependabot.

Issue #548, opened 2022-06-08
Clean up translation files


ktohalloran:
The translation files have grown as the app has evolved, adding new phrases but not removing old, now irrelevant ones. This task is to clean up those files and remove phrases that are no longer necessary.

ddohler:
@ktohalloran can this be closed or are we waiting on final approval from BHA?

ktohalloran:
@ddohler This is more of a tech debt card to clean up translations from the files that are no longer needed, as those files are unwieldy and contain a lot of language from previous iterations of the app. The final edits to translations for this phase of the project were finished in card #568 so we don't need anything more from BHA on that front.

ddohler:
Got it, thanks, I put a technical debt label on this to clarify!

PR #568, opened 2022-07-05
Add in missing translations

### LINKED ISSUE #532: Translations cleanup

ktohalloran:
## Overview

This PR adds in final missing translations following recent copy changes, such as those in #461 and #544.

### Checklist

### Notes

This does not include translating login components because there's no way to change the language from the login page, though the language can change on the login page if a user sets a language and then logs out.

One of my concerns was missing something when compiling the list of phrases in need of new translations. As a result, I found the easiest way to be as thorough as possible was to make a list of all possible use cases and pages in which language appears. I've added this here in case it is helpful for future translation checks or for #548.

**Log In** 
- Sign in (existing user)
- Sign up (not existing user)
- Magic link email
- Sign in: submitted email success message
- Sign in: user does not exist message
- Sign up: submitted email success error message
- Sign up: user already exists error message

**Edit Profile**
- With voucher
- Without voucher
- Pop ups
    - “Why are we asking you this?” on hover
    - Hover over delete address button
    - Hover over question marks next to importance factors
- Try to delete primary destination error message
- Try to save without a destination error message

**Dock**
- Using transit with express 
- Using transit without express
- By car
- Pop ups
    - Public security
    - Rental units

**Header**

**Neighborhood Detail**

**Map Pop Ups**
- Pop ups
    - Apartment listings
    - Origin marker

**Error Messages**
_These are error messages that are more difficult to trigger purposefully and are probably easier to check in the translation files_
- Profile.SaveError
- SignIn.SignUpError
- NeighborhoodDetails.ListingsFetchError

## Testing Instructions

 * Run `./scripts/server`
 * Look through the app and some of the pages/situations listed above (other than the login page) and make sure there are no missing translations
 * Verify that the translations listed in [this document](https://docs.google.com/document/d/1dZUQ3yj-Ll_K6Lb9EI4H2NGM33YmMOdERHcBu6NK8zE/edit?usp=sharing) appear as they should

Resolves #532


ddohler (approved):
Looks great! I couldn't find any strings that aren't translated.

Issue #572, opened 2022-07-22
Users created via the admin don't have associated emails


ddohler:
I think the quickest and most universal way to resolve this will be to [listen for the pre-save signal for the User model](https://docs.djangoproject.com/en/3.2/topics/signals/#connecting-to-signals-sent-by-specific-senders) and then set `instance.email = instance.username` if `instance.email` is blank at that point.

https://docs.djangoproject.com/en/3.2/ref/signals/#pre-save

PR #573, opened 2022-07-22
Fix case sensitive logins and missing emails

### LINKED ISSUE #571: Fix case sensitivity in frontend & admin login
Currently, if you enter your email into either the admin or frontend login in a different case than it is saved in the database, you will not be able to log in. This issue is to make the frontend sign in/sign up and admin sign in case insensitive.

ktohalloran:
## Overview

This PR addresses problems with users not receiving login emails from either the frontend or the backend. The issues were traced back to the fact that we had set up the username field to be called email and hid the actual email field from the admin. This PR fixes the issue by:
- Adding case insensitivity to both the frontend login form and admin login so that users can enter any variation of cases and still be able to login
- Adding a migration to 1) convert all existing users' emails to lowercase and 2) copy their username to the email field
- Adding a pre-save receiver to 1) convert all existing users' emails to lowercase and 2) copy their username to the email field on the creation of new accounts.

### Checklist

## Testing Instructions

 * If you don't have any accounts in your database with capitalized email addresses, add one from `develop` or before pulling this branch and running migrations
 * Run migrations: `./scripts/manage migrate`
 * Check that the previously capitalized email is now lowercase and that all entries now have their username twice: once in the username field and again in email
 * Create a new user from the admin with a capitalized email and verify that it converts to lowercase and copies to the email field on save
 * Log in with a capitalized version of an admin email. It should allow you to login.
 * Navigate to the frontend and login with a capitalized version of your email. It should still "send" you an email in the terminal.
 * Sign up as a new user and enter a capitalized email. Verify that it also saves as lowercase and is copied to the email field.
 

Resolves #571 & #572

### DIFF
--- a/src/backend/users/admin.py
+++ b/src/backend/users/admin.py
@@ -13,6 +13,13 @@ from .models import Destination, UserProfile
 
 class LoginForm(AuthenticationForm):
     username = forms.CharField(label="Email")
+    error_messages = {"invalid_login": "Please enter a correct email and password."}
+
+    # convert entered email to lowercase
+    def clean(self):
+        self.cleaned_data["username"] = self.cleaned_data.get("username").lower()
+        cleaned_data = super(LoginForm, self).clean()
+        return cleaned_data
 
 
 # Validates that the username is a valid email on save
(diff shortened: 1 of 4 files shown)


ddohler (approved):
Works great!

PR #576, opened 2022-07-29
Fix polygon corruption on slow internet

### LINKED ISSUE #574: Polygons on map sometimes appear corrupted
Screen Shot 2022-07-22 at 1 22 14 PM

I saw this during the development of #550 but @rachelekm wasn't able to replicate it, and it hasn't happen since for me. I suspect it may be a race condition since it happened when we switched from static neighborhood boundaries to loading them dynamically from the API, but I don't know exactly what might be triggering it.

Places where this is known to have shown up are:
- Firefox, on MacOS, in local development
- Firefox, on Windows, on staging.

It may be necessary to use Browserstack to diagnose this. If so, please post appropriately in #browserstack.

ktohalloran:
## Overview

Polygons were generating in a way that looked corrupted under slow internet conditions due to the update of transit networks before neighborhood bounds state had been updated. This PR conditionally adds a pause and recursive check to the generation of the networks, specifically where networks are assigned paths and times for the given origin point, so that they are not generated before `neighborhoodBounds` state exists.

### Checklist

### Demo

The same map generated on 2G-speed internet before and then after this fix.

_Before_
Screen Shot 2022-07-28 at 2 56 19 PM

_After_
Screen Shot 2022-07-28 at 2 57 24 PM

## Testing Instructions

 * Start the server, navigate to **localhost:9966**, and log in
 * In the network tab, click "No throttling" and set the internet to something much slower (I made my own custom internet profile that downloads at 1 Mbit/s and uploads at 20 kbit/s)
 * Refresh the page and wait for it to load. Once it loads, the map should look normal
 * Log out and log in again using the same tab with the slow internet connection. Again, the loaded map should look normal.

Resolves #574

### DIFF
--- a/src/frontend/src/actions/network.js
+++ b/src/frontend/src/actions/network.js
@@ -1,6 +1,7 @@
 // @flow
 import lonlat from "@conveyal/lonlat";
 import fetch, { fetchMultiple } from "@conveyal/woonerf/fetch";
+import { isEmpty } from "lodash";
 
 import { retrieveConfig, storeConfig } from "../config";
 import { ACCESSIBILITY_IS_LOADING, ACCESSIBILITY_IS_EMPTY, TAUI_CONFIG_KEY } from "../constants";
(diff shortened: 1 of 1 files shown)


ddohler (approved):
Thanks for digging into this! Timeouts are most appropriate in situations where we have no control over the order in which things occur, i.e. because we're waiting for events to take place on a remote system. Ordinarily, in a case like this where we (theoretically) have full control over what happens in what order, we'd want to refactor so that the action to load networks doesn't get dispatched until we're certain that the neighborhoods are in place.

In this case, however, I agree that this is the best approach given the current project constraints. I do think this is introducing technical debt that we should resolve in the future, so I would add a comment with a link to a tech debt issue.

Otherwise, this looks good, and it solved the problem for me!

ktohalloran:
Thanks so much for your review and for explaining when timeouts are appropriate @ddohler! I've created this tech debt issue in #577.

PR #581, opened 2022-09-06
Super-linter refinements

BryanQuigley:
Have GH super-linter pull from env file and on GH Actions run lint/cibuild in parallel
    
This creates a .github/super-linter.env that both a local ./scripts/test run and Github Actions will use to configure super-linter.
    
Separate the super-linter build job from the rest of the build process.  Can reduce the total time to just 4 minutes (as opposed to 8 min). Also should be faster to fail, previously we had to wait for the whole cibuild to finish before linting.
    
The suggestion for the one env variable was from https://github.com/github/super-linter/pull/3270.

### Checklist

### Demo

Show all check details

## Testing Instructions

 * Confirm that build looks the same as previously but ENV variables only needed in one place.

Note: this can be charged to Azavea Infrastructure Management - this will help our other projects more than this one.


ddohler (approved):
Looks great, nice cleanup! I wish the technique for doing this was a bit more prominent in the GitHub docs.

rachelekm (approved):
Awesome! Works great and much more clean 👍

PR #582, opened 2022-09-06
Fix 500 error when user clicks expired link

### LINKED ISSUE #580: Fix 500 error on magic link expiration
Currently, if you request a login link then let it expire (the lifespan is currently set to 1 hour), when you click on it it results in a 500 Server error and a 
```
IntegrityError: null value in column "user_id" of relation "authtoken_token" violates not-null constraint
```
error via Rollbar. We ideally should fix this to provide more helpful messaging like "Your login link has expired, please request a new one".

ddohler:
## Overview

Redirects the user back to the login page if they click on an expired link, rather than throwing a 500 error.

Also updates the email text to explain that login links only last for
one hour.

### Checklist

### Demo

N/A

### Notes

Ideally, we'd explain to the user that their link had expired within the application when they try to use an expired link, but I think this is sufficient for a quick fix.

## Testing Instructions

 * On `develop`, update `settings.py` and change the `SESAME_MAX_AGE` parameter to something small, like `60`, so that links expire quickly for testing.
 * Start `./scripts/server` and enter your email on the login screen.
 * Wait long enough for the token to expire, then use the email link from the console. The email text should specify that the link will expire in one hour (though that will be false, in this case).
 * You should see a 500 error.
 * Check out this branch and try to log in again, with either the old expired email or a new one.
 * You shouldn't see a 500 error; instead, you should simply be redirected back to the login screen.

Resolves #580

### DIFF
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -31,7 +31,7 @@ def send_login_link(request):
 
     html_message = """
     <p>Hi there,</p>
-    <p>Thanks for using ECHO! Here is your <a href="{}">link to login</a>. </p>
+    <p>Thanks for using ECHO! Here is your <a href="{}">link to login</a>. This link is valid for 1 hour.</p>
     <p>BHA</p>
     """.format(
         login_link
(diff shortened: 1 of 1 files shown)


ktohalloran (approved):
Looks good! 👍

PR #584, opened 2022-09-07
Feature/bryan/remove old terraform

### LINKED ISSUE #448: Remove old amplify style deployment

BryanQuigley:
## Overview

This is 3 commits that:
 * Remove the amplify script and update the README removing taui/amplify
 * Remove all the old terraform/amplify code
 * Rename terraform-django to terraform.

The production setup was modified live without terraform to do the redirect from the old to the new site - so the old production terraform is not longer valid. We also don't want someone wasting time with the old code when they have to make a change in the future.  

### Checklist

## Testing Instructions

 * Confirm renames we're correct and that all code removed was actually dead.

Resolves #448


README.md
```diff
@@ -2,8 +2,6 @@
 
 Website to explore Choice Neighborhoods in the Boston area.
 
-[![Build Status](https://travis-ci.org/azavea/echo-locator.svg?branch=develop)](https://travis-ci.org/azavea/echo-locator)
-
 ## Requirements
 
 While Docker is the preferred development environment, steps for running directly on host
```
ktohalloran: It looks like the directions for running directly have been removed, so this line and the "to run directly" requirements below it can also be removed.

README.md
```diff
@@ -53,19 +39,6 @@ and run a development server:
 $ ./scripts/server
```
ktohalloran, line 33: Should we add `./scripts/setup` as a step here?

ktohalloran (approved):
Looks good! 👍 It's nice to see all of this old code removed! Just a few comments:

I feel like this is a good opportunity to verify if we're still using the `ALLOW_ANONYMOUS` env var and remove from `.env.sample` if not. @ddohler, `ALLOW_ANONYMOUS` was made unnecessary by our shift away from Amplify/Cognito, right?

I can't comment on this part of the `README` directly, but this also seems a good opportunity to update line 63 in the [Creating a staff user account section](https://github.com/azavea/echo-locator/blob/feature/bryan/remove_old_terraform/README.md#creating-a-staff-user-account) to specify that in the new setup, the superuser's email and username should match.

I think we can remove [lines 76-77](https://github.com/azavea/echo-locator/blob/develop/.gitignore#L76) in the root-level `.gitignore`.

README.md
BryanQuigley: Indeed. I had it above but not in a obvious format.  Changed

README.md
BryanQuigley: Removed those too

BryanQuigley:
Thanks Katie! - I had missed a good amount of the amplify bits in the readme.  All other changes made too.

Specifically want to call out ALLOW_ANONYMOUS which I removed from sample env file per above but still would like the +1 from Derek.

ddohler (approved):
Looks good! I couldn't find anything else that I would have deleted, and I didn't see anything that was deleted that looks like it needs to stick around.

Regarding `ALLOW_ANONYMOUS`, that can indeed be removed; it used to be a flag that the Amplify/Cognito components would use to determine whether to allow anonymous login, but we now [allow anonymous login unconditionally](https://github.com/azavea/echo-locator/blob/develop/src/frontend/src/components/custom-sign-in.js#L83-L94), so it shouldn't have any effect at this point.

I did notice one change, which is that there's a typo in the file name of `deployment/README.md` (it is currently `READEME.md`). Other than that though, this looks great and is a very satisfying cleanup!

BryanQuigley:
Thanks - corrected typo.

PR #604, opened 2025-03-04
Update docker container infra and STRTA

### LINKED ISSUE #598: Project spinup & immediate dev env updates
Catch-all card for spinning up project locally and addressing any immediate issues. We're using the outdated Azavea docker django quay.io image, so this should at least cover some basic container infra updates.

rachelekm:
## Overview

Implements the following updates to unblock setup:

- We no longer maintain the Azavea docker-django quay.io image. This commit replaces it with the community image and adds any necessary dependencies to support.
- The `node:12-slim` image we use for our `app` container is breaking due to upstream dependencies, debian stretch repositories, that have since been archived and moved. We'll update the node image as part of re-writing the frontend but until then this implements a temporary fix informed by https://github.com/azavea/pwd-parcel-viewer/pull/943#issuecomment-1753353001)

Not blocking setup, but this PR also updates `docker-compose` to `docker compose` syntax in line with compose v2. Additionally adds `scripts/_sourced` and `scripts/yarn` to clean up existing STRTA for improved readability as well as a CHANGELOG file.

### Checklist

- [x] fixup! commits have been squashed
- [x]  CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] README.md updated if necessary to reflect the changes
- [x] CI passes after rebase

## Testing Instructions

 * Run through setup instructions from `README` and confirm no errors with spin up
 * Run `scripts/server` confirm no errors running application
 * Run `scripts/format` and `scripts/test` and confirm no errors. NOTE: I recommended commenting out the commands for the GH Super-Linter within the `scripts/test` before running and instead confirming a successful CI build. The superlinter will pull down a massive docker image/take ~10 min to run locally. I've opened issue https://github.com/azavea/echo-locator/issues/605 to migrate to megalinter instead.

Resolves #598


scripts/bootstrap
```diff
@@ -1,19 +1,10 @@
 #!/bin/bash
 
-set -e
-
-if [[ -n "${ECHOLOCATOR_DEBUG}" ]]; then
-    set -x
-fi
-
 DIR="$(dirname "${0}")/../"
+USAGE="Pull down secrets from S3."
```
rajadain, line 4: Since this is run by `setup`, which a dev is instructed to run in the README, should the README also include a reference to `ECHOLOCATOR_SETTINGS_BUCKET`? Without that, the `.env.sample` file gets copied to `.env`, instead of the real one being pulled down from S3.

src/frontend/Dockerfile
```diff
@@ -1,5 +1,12 @@
 FROM node:12-slim
 
+# Update debian stretch repositories
+# Fixes upstream dependency of the node:12-slim base image
```
rajadain, line 4: Is there a card to upgrade this container, or will that be done as part of the front-end refresh? We should use something modern for the new work.

rajadain (approved):
This looks good, ran `setup` and `server` and created a superuser account: things are working as instructed in README.

In addition to the small comments, the only other thing that stood out was that there's two documentation folders in this project: `doc/arch` and `docs/arch`. This is likely a mistake. `doc/arch` is more standard. Would recommend consolidating the contents there.

src/frontend/Dockerfile
rachelekm: As part of the front-end refresh but it wasn't documented, just added issue https://github.com/azavea/echo-locator/issues/606 to ensure that work is captured!

scripts/bootstrap
rachelekm: I had to [go back in history to check myself ](https://github.com/azavea/echo-locator/pull/500#discussion_r872517685)but I think that was the original intention, we want new dev users to copy `.env.sample` to `.env` and manually populate the values from the entry in 1password with the `ECHOLOCATOR_SETTINGS_BUCKET` mostly for loading the dotenv file in CI. However that feels clunky, especially when it could just pull from s3 and then go right to building. I checked and there *is* a `echo-locator-devdjango-config-us-east-1` bucket with an `.env` file that has the correct values for local development. I'll update the README to include `ECHOLOCATOR_SETTINGS_BUCKET` set to that value on initial spin up.

rachelekm: Added README change in [d250ccd](https://github.com/azavea/echo-locator/pull/604/commits/d250ccd5b1f396d9cac5525ef6af23d6bf2e3d10)

rachelekm:
Thanks for the review, @rajadain! Addressed those comments and also consolidated the doc directories. Will merge once review permissions are setup.

rajadain:
I was granted write access, and now my review above counts as legit. Go ahead and merge!

PR #627, opened 2025-05-09
Parameterize and update URLs to analysis files

### LINKED ISSUE #624: Create resources for BHA to host static analysis files
The routing functionality of ECHO is reliant on data files which are currently hosted by Conveyal. In order to improve the long-term resilience and sustainability of the application, we'll centralize these files into the existing AWS account that BHA controls and uses to host the remainder of the ECHO application's infrastructure.

This issue is to create the following to support hosting these static files:
- New S3 bucket dedicated to storing static analysis files
    - Also create bucket sub-directories for each year of analysis to better organize/archive files (right now, 2020 and 2025)
    - There should be analysis files for every network type: "Off Peak No Express", "Peak No Express", "Peak", "Off Peak"
    - Developer user group should have read/write access
- Create new read-only CloudFront distribution for S3 bucket

### LINKED ISSUE #625: Update network analysis links
The routing functionality of ECHO is reliant on data files which are currently hosted by Conveyal. In order to improve the long-term resilience and sustainability of the application, we'll centralize these files into the existing AWS account that BHA controls and uses to host the remainder of the ECHO application's infrastructure.

Following completion of #624, update CloudFront URL links in `src/frontend/src/config.json` and `src/frontend/src/reducers/data.js` from the Conveyal distribution to the BHA-hosted distribution for each network type.

aaronxsu:
## Overview

This PR:
- parameterizes the CloudFront URL that serves the static analysis files and reads it from the environment, which will make it easy in the future to update the values without code changes
- leaves a to-do item in the data reducer that we will remove the `grids` in the dev phase
- updates the bootstrap script to pull `dotenv` instead of `.env` file from S3 config bucket

What's done but not in code of this PR:
- fixed the CORS errors from the CloudFront endpoints
- updated `dotenv` files in dev and staging S3 buckets to include the new env var `NETWORK_URL_ROOT`

### Checklist

- [ ] CI passes after rebase

## Testing Instructions

 * Add `NETWORK_URL_ROOT='https://d1kinuc21h1lp5.cloudfront.net/2020'` to your local `.env` file; Or remove the file entirely and run `AWS_PROFILE=<echo profile> ECHOLOCATOR_SETTINGS_BUCKET=echo-locator-devdjango-config-us-east-1 ./scripts/bootstrap`
 * Spin up the app
 * Make sure the analysis json and data files load from the new URL and works fine in the app
 * Push up `test` branch and make sure it also works fine on staging

Resolves https://github.com/azavea/echo-locator/issues/624
Resolves https://github.com/azavea/echo-locator/issues/625


src/frontend/src/config.js
```diff
@@ -4,3 +4,26 @@ export const clearLocalStorage = () => window.localStorage.clear();
 export const retrieveConfig = (key) => JSON.parse(window.localStorage.getItem(key));
 export const storeConfig = (key, json) =>
   window.localStorage.setItem(key, JSON.stringify(json, null, "  "));
+
+export const networks = [
+  {
+    name: "Peak",
+    url: `${process.env.NETWORK_URL_ROOT}/peak`,
```
aaronxsu, line 11: We don't have to use the name `NETWORK_URL_ROOT`- let me know if you prefer something else

scripts/bootstrap
```diff
@@ -18,7 +18,7 @@ function pull_env() {
             cp .env.sample .env
             echo "Copied environment variables file from .env.sample to .env for local development."
             echo "Please populate the fields in this file manually."
-        elif aws s3 cp "s3://${ECHOLOCATOR_SETTINGS_BUCKET}/.env" ".env"; then
+        elif aws s3 cp "s3://${ECHOLOCATOR_SETTINGS_BUCKET}/dotenv" ".env"; then
```
aaronxsu, line 21: `dotenv` will make it easy for us to pull down and edit, since on some OS, if not setup, the file will be hidden, meanwhile we know what the file is.

src/frontend/src/config.json
```diff
@@ -6,27 +6,5 @@
       "url": "https://d2z7d5345ccuw9.cloudfront.net/5c9bdffa37ed813dbe27f8f1/Workers_with_earnings_1250_per_month_or_less.grid",
       "showOnMap": true
     }
-  ],
```
aaronxsu, line 9: I meant to leave a comment in this JSON file, but: 1. I did not find a good way to comment a JSON file; 2. we might be removing the JSON file in dev phase.

src/frontend/src/config.js
rachelekm: This naming makes sense to me 👍

src/frontend/src/reducers/data.js
```diff
@@ -159,6 +161,7 @@ export default handleActions(
     },
   },
   {
+    // TODO: grids are irrelevant and can be scrapped once we move away from Taui
```
rachelekm, line 164: 👍

src/frontend/src/config.js
rachelekm: Actually, what do you think about separating the year path from the NETWORK_URL_ROOT? If we have a separate TRANSIT_YEAR variable constant on the frontend it might be easier to reference/update after re-generating analyses in the future

rachelekm (changes requested):
The .json files are downloading without issue (🎉) but looks like I'm still having access trouble trying to fetch the data files:
Screenshot 2025-05-12 at 2 40 10 PM

aaronxsu:
>but looks like I'm still having access trouble trying to fetch the data files

@rachelekm Hmm, this is interesting. Looks like they are not CORS errors but 403 errors instead. Did you see them locally or in the `test` branch on staging? Maybe we should pick up a huddle to take a look. The data files load fine for me locally.

src/frontend/src/config.js
aaronxsu: >Actually, what do you think about separating the year path from the NETWORK_URL_ROOT? If we have a separate TRANSIT_YEAR variable constant on the frontend it might be easier to reference/update after re-generating analyses in the future

Good question. Do we expect to support year selection in the app? If we do, then we should leave out the year part from the `NETWORK_URL_ROOT` env var. Otherwise, we might not have a use case for separating out the year. Assuming there is no year selection feature, after we have data from 2025, we will just update the `NETWORK_URL_ROOT` env var on S3 with the new year in the URL segment and re-run the CI job to rebuild the app. What do you think? Does it make sense?

rachelekm: > Do we expect to support year selection in the app?

Nope! It's assumed we'll always use the latest transit analysis files and those will likely not update *that* often (annually, if that). Updating the env var on S3 makes sense 👍

rachelekm:
Thanks, @aaronxsu! Confirming the issue ended up being that the `*_times.dat` and `*_paths.dat` files are [dynamically prefixed and fetched based on a users address](https://github.com/azavea/echo-locator/blob/7ef73658618efecac1c2cb5f71eec83203689018/src/frontend/src/actions/network.js#L192). We only copied a small sample into our hosted bucket thinking these were static calls (like the json files) and I was inputting an address that was fetching a file we haven't brought in. Noting that aaron updated the bucket policy to allow CloudFront distribution to list objects in the S3 bucket, so now it should return 404 rather than 403 if a requested object is not found.

I'll open an issue (https://github.com/azavea/echo-locator/issues/628) to track the remaining data files copied into our bucket and test once complete. In the meantime we're all setup for hosting analysis files, we just need to push the remaining data files up under corresponding prefixes in the `echo-locator-analysis-files` S3 bucket.

rachelekm (approved):
This works well, thanks for setting these resources up

aaronxsu:
Thanks for testing and creating the new issue- that was a good catch and I am glad that we found that out now. Going to quash the fixup commits and merge.

Issue #605, opened 2025-03-04
Migrate from superLinter to megalinter


rachelekm:
This project is still using superlinter, which pulls a large image file, runs slowly, and is difficult to maintain in a development environment as well as parse errors in its log result. We since have switched to megalinter in our project template and other projects, we should make a similar migration here for more efficient testing and linting. On implementation, should largely follow the process outlined in the project template, https://github.com/azavea/azavea-project-template/pull/50.

rachelekm:
Superseded by #639

rachelekm:
Re-opening as this is still relevant following latest frontend bootstrapping, [see comment here. ](https://github.com/azavea/echo-locator/pull/666#discussion_r2162075635)

PR #666, opened 2025-06-19
Bootstrap new frontend

### LINKED ISSUE #639: Bootstrap frontend
Scrap the existing frontend and prepare for rewrite following decisions from #599 

Use recently updated project template as a React + Vite base, with the following changes:
- Update README
- Remove Chakra UI and related dependencies from project template
- Remove Cypress and related dependencies (see note in PR comment on Cypress and [end-to-end testing issues here](https://github.com/azavea/azavea-project-template/pull/89)) from project template
- Add `react-aria`, `tailwindcss`, and [recommended react aria tailwind plugin](https://react-spectrum.adobe.com/react-aria/styling.html#plugin)
- Update STRTA and CI/CD pipelines as necessary (we got the go ahead to deploy to staging)

aaronxsu:
## Overview

This PR:
- removes the old frontend
- bootstraps a new frontend with essential deps
- updates Django settings for how static assets are collected

### Checklist

- [ ] CI passes after rebase

## Testing Instructions

 * Test new frontend: http://localhost:9966 is the home page, and http://localhost:9966/components is the components page
 * Push up a `test` branch to trigger a staging deployment. After it's deployed, go to https://stg.echosearch.org/ and then also visit https://stg.echosearch.org/components. Also make sure https://stg.echosearch.org/admin works fine as well.

Resolves https://github.com/azavea/echo-locator/issues/639


docker-compose.yml
```diff
@@ -31,7 +31,9 @@ services:
       dockerfile: ./src/backend/Dockerfile
     volumes:
       - ./src/backend:/usr/local/src/backend
-      - ./src/frontend/build:/usr/local/src/backend/static/
+      # Mount Vite dist to a new, distinct directory
+      # This directory must match the path in STATICFILES_DIRS in settings.py
+      - ./src/frontend/dist:/usr/local/src/backend/vite_dist_output
```
rachelekm: I think it would be easier for other developers to spin up/troubleshoot if we [match the project template pattern here ](https://github.com/azavea/azavea-project-template/pull/89/commits/b66ec860a571428ae24a54a45661c1e0859765b9). Instead of having a uniquely named `vite_dist_output` and `collected_static_files` directories we just directly target and mount `/dist/assets`. What do  you think?

scripts/test
```diff
@@ -5,9 +5,6 @@ USAGE="Run linters and tests."
 source ./scripts/_sourced
 check_for_help_flag "$@"
 
-echo "Running frontend app tests..."
-./scripts/yarn run test
-
```
rachelekm, line 10: I can't add a comment below, but noting here that the project template shifted from using GH superlinter to megalinter ([this was captured in an issue I prematurely closed](https://github.com/azavea/echo-locator/issues/605)), if we have time in the scope of this card we might want to switch to megalinter to better match the project template. If it feels out of scope I can re-open that issue and we can tackle as part of the setup epic.

scripts/update
```diff
@@ -11,7 +11,7 @@ function cleanup() {
 
 trap cleanup ERR
 
-# Build app container image first
+# Build app container image
 docker compose build app
```
rachelekm: I don't know why the app and db/django builds were separate, but I wonder if this can now just be `docker compose build`?

src/backend/Dockerfile
```diff
@@ -2,6 +2,10 @@ FROM python:3.13-slim-bookworm
 
 # Set DEBIAN_FRONTEND to noninteractive to avoid prompts during build
 ENV DEBIAN_FRONTEND=noninteractive
+# Ensures Python output is unbuffered
+ENV PYTHONUNBUFFERED 1
+ENV DJANGO_SETTINGS_MODULE echo.settings
```
rachelekm: What does PYTHONUNBUFFERED do? Also DJANGO_SETTINGS_MODULE seems already set in [ASGI and WSGI configs](https://github.com/azavea/echo-locator/blob/feature/asu/bootstrap-new-frontend/src/backend/echo/asgi.py#L14), do we need it here too?

src/backend/Dockerfile
```diff
@@ -51,7 +55,10 @@ RUN set -ex \
...
 RUN psql --version
 
-ENTRYPOINT ["/usr/local/bin/gunicorn"]
-
+# Copy Django project source code
 COPY ./src/backend /usr/local/src/backend
-COPY ./src/frontend/build /usr/local/src/backend/static
+
+# Expose the port Gunicorn server will listen on
+EXPOSE 8085
```
rachelekm: Do we need this too if we have [gunicorn.conf.py](https://github.com/azavea/echo-locator/blob/feature/asu/bootstrap-new-frontend/src/backend/gunicorn.conf.py)?

src/frontend/vite.config.ts
```diff
@@ -0,0 +1,15 @@
...
+export default defineConfig({
+    plugins: [react(), tailwindcss()],
+    server: {
+        watch: {
+            usePolling: true,
+        },
+        host: true,
+        strictPort: true,
+        port: 9966,
+    },
```
rachelekm, line 14: ~~We had to add a [django proxy config as part of project template](https://github.com/azavea/azavea-project-template/pull/89/commits/b55e0b786aaf6bc0d1282579ce9bcc3b17c72d0f), do you think we need to add something similar here too?~~ Nevermind, I tried out `https://stg.echosearch.org/api/neighborhoods/` with the new deployment which is successful so I don't think the project template proxy config is necessary 🎉

src/frontend/package.json
```diff
@@ -1,117 +1,60 @@
...
-    "parser": "babel-eslint"
-  }
+    "name": "echo-locator",
+    "private": true,
+    "packageManager": "yarn@1.22.19",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/azavea/echo-locator.git"
+    },
+    "author": "Azavea",
```
rachelekm:
```suggestion
    "author": "Element 84",
```

rachelekm (changes requested):
This is working great! I have a couple small suggestions/questions, the most significant being the static directory naming convention which would be worth testing again if we change, but other than that this is close to merging!

scripts/test
aaronxsu: Ah, got it. How about we re-open that issue? The frontend part is fresh and should be straightforward to switch to a new linter. I suspect backend might be a bit more involved- just a guess.

docker-compose.yml
aaronxsu: Good call. Addressed in https://github.com/azavea/echo-locator/pull/666/commits/0e0197a30e97f11bf504c9f3af52070e36f53d27

scripts/update
aaronxsu: I don't recall the reason exactly, but I do vague remember something about the frontend docker container needs to build first (frontend used to have a dockerfile for itself before this PR) and then build the static app, and then the backend (django, postgres) containers build, so that later the static asset collection could work etc.

We shouldn't have that issue anymore. Addressed in https://github.com/azavea/echo-locator/pull/666/commits/71f43fe95b69a09be52b5c9b5d650643ea7745e5

src/backend/Dockerfile
aaronxsu: I removed `DJANGO_SETTINGS_MODULE` in https://github.com/azavea/echo-locator/pull/666/commits/0d9d7621d5399680ff38e2802c9ab729df8a02f6

>What does PYTHONUNBUFFERED do?

This was a result of giving Gemini the dockerfile and it suggested this improvement. I think the reasoning justifies it, and decided to add this in. Please let me know what you think!

> The line ENV PYTHONUNBUFFERED 1 in a Dockerfile is a very common and important best practice when running Python applications inside Docker containers.
> 
> Here's what it means and why it's used:
> 
> What is Buffering?
> 
> By default, Python's standard output (stdout) and standard error (stderr) streams are buffered. This means that when your Python application prints something (e.g., using print(), or logging messages), that output isn't immediately sent to the underlying operating system or terminal. Instead, it's held in a temporary memory buffer. The buffer is only flushed (i.e., its contents are actually sent out) when:
> 
> The buffer becomes full.
> The program explicitly flushes it (e.g., sys.stdout.flush()).
> The program exits.
> A newline character is encountered, but only if output is to a TTY (like a terminal you're directly interacting with).
> Why is it a Problem in Docker?
> 
> In a typical Docker environment, your application's stdout and stderr are usually redirected to Docker's logging driver (e.g., json-file, syslog, journald). Docker collects these streams as logs.
> 
> If Python's output is buffered:
> 
> Delayed Logs: Your application might print something, but it won't show up in docker logs <container_id> immediately. There could be a significant delay (seconds or even minutes) until the buffer fills up or the application eventually exits.
> Missing Logs on Crash: If your application crashes unexpectedly before the buffer is flushed, you might lose valuable log messages that occurred just before the crash, making debugging very difficult.
> Debugging Frustration: When you're trying to debug an issue, you want to see logs in real-time as they happen, not after a delay.
> How ENV PYTHONUNBUFFERED 1 Solves It:
> 
> Setting the PYTHONUNBUFFERED environment variable to 1 (or any non-empty string) forces Python's stdout and stderr streams to be unbuffered.
> 
> This means:
> 
> Real-time Logs: Every print() statement or log message is immediately written out to the underlying stream.
> No Lost Logs: If the application crashes, all output up to that point should have already been sent to Docker's logging system.
> Easier Debugging: You see exactly what your application is doing as it happens in the docker logs output.

src/backend/Dockerfile
aaronxsu: Removed in https://github.com/azavea/echo-locator/pull/666/commits/0d9d7621d5399680ff38e2802c9ab729df8a02f6

src/frontend/package.json
aaronxsu: https://github.com/azavea/echo-locator/pull/666/commits/fd09ed2e972fe995cc3605f0b4286086adbe4d3d

aaronxsu:
@rachelekm thanks for the review and the feedback. I addressed them and replied to your comments. Please let me know what you think.

I also kicked off a staging deployment with the `test` branch after these changes, flushed the staging cloudfront cache, and it looks like staging still works well.

scripts/test
rachelekm: Good call on the backend I feel like I remember that being messy - I'll reopen and drop it into the setup epic lineup!

src/backend/Dockerfile
rachelekm: Oh neat, yeah let's keep `PYTHONUNBUFFERED` in!

rachelekm (approved):
Thank you for implementing those updates and kicking off a staging deployment for testing, this looks great and good to go!

aaronxsu:
Thanks for the re-review. I squashed the fixup commits and pushed up. Going to merge after CI.

PR #672, opened 2025-07-01
Add mobile translations subheader

### LINKED ISSUE #649: Create mobile translations subheader
Create mobile language options subheader [according to figma designs,](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=852-21648&m=dev) potentially utilizing rect-aria's `ButtonGroup` component.

aaronxsu:
## Overview

This PR builds on top of [`RadioGroup`](https://react-spectrum.adobe.com/react-aria/RadioGroup.html) from React Aria and creates the translations subheader component.

I ended up using `RadioGroup` rather than the button group, because there should always be one option selected, which makes a radio group semantically closer to the use case.

Tagging @philippschmitt for styling improvements.

### Checklist

- [ ] CI passes after rebase

### Demo

Screenshot 2025-07-01 at 6 28 59 PM

## Testing Instructions

 * Go to http://localhost:9966/components
 * Click on the buttons in the "Radio Button Group" section, and make sure the clicked option's button border is outlined

Resolves https://github.com/azavea/echo-locator/issues/649

### DIFF
--- a/src/frontend/src/components/SelectLanguageButtons.tsx
+++ b/src/frontend/src/components/SelectLanguageButtons.tsx
@@ -27,12 +27,13 @@ const SelectLanguageButtons = ({ language }: Props) => {
 
     return (
         <RadioButtonGroup
+            aria-label="Select a language"
             value={selectedLanguage}
             onChange={setSelectedLanguage}
         >
-            <RadioButton value={Languages.EN} label="English" />
-            <RadioButton value={Languages.ES} label="Español" />
-            <RadioButton value={Languages.ZH} label="中文" />
+            <RadioButton value={Languages.EN}>English</RadioButton>
+            <RadioButton value={Languages.ES}>Español</RadioButton>
+            <RadioButton value={Languages.ZH}>中文</RadioButton>
         </RadioButtonGroup>
     );
 };
--- a/src/frontend/src/components/SelectLanguageButtons.tsx
+++ b/src/frontend/src/components/SelectLanguageButtons.tsx
@@ -1,0 +1,40 @@
+import { useEffect, useState } from "react";
+
+import {
+    RadioButton,
+    RadioButtonGroup,
+} from "./base/RadioButtonGroup/RadioButtonGroup";
+
+export const Languages = {
+    EN: "en",
+    ES: "es",
+    ZH: "zh",
+};
+
+type LanguageKeys = (typeof Languages)[keyof typeof Languages];
+
+interface Props {
+    language: LanguageKeys;
+}
+
+const SelectLanguageButtons = ({ language }: Props) => {
+    const [selectedLanguage, setSelectedLanguage] =
+        useState<LanguageKeys>(language);
+
+    useEffect(() => {
+        setSelectedLanguage(language);
+    }, [language]);
+
+    return (
+        <RadioButtonGroup
+            value={selectedLanguage}
+            onChange={setSelectedLanguage}
+        >
+            <RadioButton value={Languages.EN} label="English" />
+            <RadioButton value={Languages.ES} label="Español" />
+            <RadioButton value={Languages.ZH} label="中文" />
+        </RadioButtonGroup>
+    );
+};
+
+export default SelectLanguageButtons;
--- a/src/frontend/src/components/base/RadioButtonGroup/RadioButtonGroup.styles.ts
+++ b/src/frontend/src/components/base/RadioButtonGroup/RadioButtonGroup.styles.ts
@@ -1,0 +1,20 @@
+import { tv } from "tailwind-variants";
+
+export const radioButtonGroupStyles = tv({
+    base: "flex gap-3",
+});
+
+export const radioButtonStyles = tv({
+    base: [
+        "flex items-center justify-center rounded-[var(--spacing-4)] border px-4 py-2 h-[34px] gap-3 self-stretch flex-1",
+        "text-gray-800 text-sm font-bold transition-colors border",
+        "cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-muted-blue-400 focus-visible:ring-offset-2",
+        "pressed:scale-[0.98]",
+    ],
+    variants: {
+        isSelected: {
+            true: "border-gray-500",
+            false: "border-gray-300 hover:border-gray-500",
+        },
+    },
+});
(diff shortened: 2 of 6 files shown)


src/frontend/src/components/SelectLanguageButtons.tsx
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useState } from "react";
+
+import {
+    RadioButton,
+    RadioButtonGroup,
+} from "./base/RadioButtonGroup/RadioButtonGroup";
+
+export const Languages = {
```
aaronxsu, line 8: I stopped using enums in this project because: https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/#the---erasablesyntaxonly-option and instead followed the solution recommended here: https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh

src/frontend/src/components/base/RadioButtonGroup/RadioButtonGroup.styles.ts
```diff
@@ -0,0 +1,20 @@
...
+export const radioButtonStyles = tv({
+    base: [
+        "flex items-center justify-center rounded-[var(--spacing-4)] border px-4 py-2 h-[34px] gap-3 self-stretch flex-1",
+        "text-gray-800 text-sm font-bold transition-colors border",
+        "cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-muted-blue-400 focus-visible:ring-offset-2",
+        "pressed:scale-[0.98]",
+    ],
+    variants: {
+        isSelected: {
+            true: "border-gray-500",
```
aaronxsu: I just changed the color of the border when the radio button is selected, which is the same style as when it's hovered. This might not be the best approach, but I did not get much luck finding the corresponding selected state's style in Figma. Please feel free to update these styles.

src/frontend/src/components/base/RadioButtonGroup/RadioButtonGroup.styles.ts
```diff
@@ -0,0 +1,20 @@
...
+            false: "border-gray-300 hover:border-gray-500",
+        },
+    },
```
aaronxsu, line 23: I did not add Small, Medium, Large sizes (unlike for the button) in the style definitions, since there seems to be only one style in the UI design, which is only used in the mobile view. We could customize size further as we encounter them later, I'd suggest.

src/frontend/src/components/SelectLanguageButtons.tsx
```diff
@@ -0,0 +1,40 @@
...
+    useEffect(() => {
+        setSelectedLanguage(language);
+    }, [language]);
+
+    return (
+        <RadioButtonGroup
+            value={selectedLanguage}
+            onChange={setSelectedLanguage}
+        >
+            <RadioButton value={Languages.EN} label="English" />
```
rajadain: Since all we're doing in our implementation of `RadioButton` is initialize styles, we should otherwise try to match the API of the underlying React Aria component. [They pass the label as a child](https://react-spectrum.adobe.com/react-aria/RadioGroup.html#value), not as a prop. Consider updating `RadioButton` to match that:

```suggestion
            <RadioButton value={Languages.EN}>English</RadioButton>
```

src/frontend/src/pages/Components.tsx
```diff
@@ -254,6 +257,17 @@ const Components = () => {
                         </div>
                     </div>
                 </section>
+
+                <section className="p-8 bg-white border border-gray-200 rounded-lg shadow-sm">
+                    <h2 className="text-3xl font-medium text-gray-800 mb-8 pb-4 border-b">
+                        Radio Button Group
+                    </h2>
+                    <div className="space-y-8">
+                        <div className="flex flex-col gap-4 w-[272px]">
```
rajadain, line 266: I hope we encode these arbitrary widths into theme variables and use those, instead of inline constants like this.

rajadain (approved):
+1 tested, looks good. Left some code comments. Also agree that the selected / active state is not distinctive enough yet, that should be tweaked further to be more prominent.

src/frontend/src/pages/Components.tsx
aaronxsu: Yes, I believe we will do that for the app UI pages according to the designs. The components page is temporary.

src/frontend/src/components/SelectLanguageButtons.tsx
aaronxsu: Agreed. Addressed in https://github.com/azavea/echo-locator/pull/672/commits/380fbe03ce2f249c0a5da92355258e6f6a902c3d

src/frontend/src/components/base/RadioButtonGroup/RadioButtonGroup.styles.ts
aaronxsu: @philippschmitt please feel free to push up styling touches to help improve this component.

philippschmitt:
Good point to use the semantic choice, `RadioGroup`! 

I don't think the button needs to give strong feedback because the entire page's language will change based on the selection. Still, I tweaked the style a bit so that the selected option looks more like a disabled button.

I also tried to add the blue outlines for keyboard focus, like all the other buttons have, but I didn't succeed. Do you know why that is? I think some focus indicator (outline/background/border change) would be important for a11y.

aaronxsu:
@philippschmitt thanks for the review.

>I also tried to add the blue outlines for keyboard focus, like all the other buttons have, but I didn't succeed. Do you know why that is? I think some focus indicator (outline/background/border change) would be important for a11y.

This is a great catch. I have a fix locally and will push it up.

>I don't think the button needs to give strong feedback because the entire page's language will change based on the selection. Still, I tweaked the style a bit so that the selected option looks more like a disabled button.

Did you get a chance to push it up yet? It might be good to push up your changes first and then I will push up my keyboard focus fix, so we won't override changes.

philippschmitt:
My bad, I forgot to push. It's up now.

aaronxsu:
@philippschmitt No worries. Thanks for pushing that up. It looks good to me. 

I pushed up my fixes for the tabbing focus. Could you please pull it down and see if it works on your end? As usual, please feel free to tweak the styles.

philippschmitt:
Works well now, thanks! 👏

aaronxsu:
Thank you both for the reviews and the suggestions. I squashed the fixup commits and added a changelog entry. Merging now.

Issue #640, opened 2025-06-17
Add static translations support


rachelekm:
Add static frontend translations support dependencies, `react-i18next`, and implement initial configuration for internationalization, for now adding empty translations files for Chinese and Spanish. Evaluate if any date/number/special characters in our figma designs not handled by `react-i18next`  that may be supported by adding `react-aria/i18n`.

rachelekm:
We already added the language switching dropdown menu components as part of #674, refactor to implement language switching logic to these components in line with react-i18next docs as part of this issue.

PR #674, opened 2025-07-05
Responsive menu component

### LINKED ISSUE #651: Create mobile & desktop menu component
Create mobile & desktop menu component [according to figma designs.](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=1075-10983&m=dev)

aaronxsu:
## Overview

This PR:
- adds a reusable hook for media query
- adds base components for dropdown and navigation tabs
- starts to add new routes to frontend router
- adds a responsive menu to all pages

### Checklist

- [ ] CI passes after rebase

### Demo

Large screen

Screenshot 2025-07-05 at 1 40 12 PM

Mobile

Screenshot 2025-07-05 at 1 40 03 PM

### Notes

1. For the language selection dropdown, I did not get much luck finding the designs for when the dropdown options are expanded, but I took a stab at a first pass of it. @philippschmitt please feel free to adjust how it looks

Screenshot 2025-07-05 at 1 43 41 PM

2. On the mobile view's hamburger menu, in Figma, I saw a note it's pending design. I wonder if there is any news on that. I'd suggest we come back to it in another issue when we have the design updated for it.

## Testing Instructions

 * Go to http://localhost:9966, and it should redirect to http://localhost:9966/discover
 * Make sure the top menu bar shows up
 * Change window sizes and make sure the menu changes based on the design of desktop v.s mobile views
 * Click on the Compare tab in the menu. Make sure the app route changes to http://localhost:9966/compare

Resolves https://github.com/azavea/echo-locator/issues/651


src/frontend/src/components/base/Button/Button.styles.ts
```diff
@@ -24,6 +24,7 @@ const buttonStyles = tv({
             outline:
                 "border border-gray-300 bg-transparent hover:border-gray-500",
             orange: "bg-orange-200 text-orange-800 hover:bg-orange-300",
+            ghost: "hover:bg-teal-200",
```
aaronxsu, line 27: Please feel free to update the styling for the "ghost" button, which is used as the trigger for the language selection dropdown in the menu bar and for the logout button

src/frontend/src/pages/Root.tsx
```diff
@@ -0,0 +1,16 @@
+import Menu from "src/components/Menu";
+import { Outlet } from "react-router";
+
+const Home = () => {
+    return (
+        <div className="min-h-screen flex flex-col">
+            {/* TODO: get compareCount from app store */}
+            <Menu compareCount={0} />
+            <div className="flex flex-1 flex-col items-center self-stretch gap-6 rounded-t-[16px] bg-white px-10 py-16">
+                <Outlet />
```
aaronxsu, line 11: https://reactrouter.com/api/components/Outlet

Assuming the menu bar shows for all pages, this outlet is like a "slot" for rendering all children components that's nested under the root path, which is `/`- see the router definition in `App.tsx`

src/frontend/src/components/base/Dropdown/Dropdown.styles.ts
```diff
@@ -0,0 +1,14 @@
+import { tv } from "tailwind-variants";
```
aaronxsu, line 1: The dropdown styles might need some design care

philippschmitt:
I tweaked a few style and refactored the main menu to use a CSS pseudo element instead of `border-top`. I think it makes it easier to style because you don't need adjust margins in multiple places/states to align the text vertically.

Thanks for adding the dropdown. I didn't make a dedicated design because I figured we'd start with whatever the component library provides. I made a few tweaks.

> On the mobile view's hamburger menu, in Figma, I saw a note it's pending design. I wonder if there is any news on that. I'd suggest we come back to it in another issue when we have the design updated for it.

Good point, that got lost. How about we just reuse the modal component that's in use all over the app? Like so:
image

aaronxsu:
@rachelekm, could you please take a look at the UX design for the hamburger menu on mobile view and let us know what you think? Thanks!

aaronxsu:
I pulled down the styling tweak commits- they look great to me.

src/frontend/src/App.tsx
```diff
@@ -1,19 +1,27 @@
...
+            children: [
+                // TODO: update redirect logic when we build frontend auth
+                {
+                    index: true,
+                    element: <Navigate to="/discover" replace={true} />,
+                },
+                { path: "discover", Component: Discover },
+                { path: "compare", Component: Compare },
+                { path: "components", Component: Components },
+            ],
```
rachelekm: Nice, thanks for setting this up early! We discussed the role of this project's routing in person and will implement gradually (with the initial setup cemented when we add frontend auth setup), I'll summarize here for our reference later:
- All child routes should be private with the exception of the login and logout pages
- We will prioritize updating the URL with syntactic paths for the following:
     -  root path (`/`), display login if unauthorized or the discover page by default
     -  Discover tab (`/discover`)
     - Compare tabe (`/compare`)
     - Neighborhood detail view (`/neighborhood/{zipcode}`)
     - User profile wizard view (`/user/{wizard-step}`)
     - Trip wizard view (`/trips/{wizard-step}`)
     - Unmatched paths should redirect to discover page
- We might want to think about incorporating the language code in the URL (ex: `/es/compare`)
- Only the components within the child routes should re-render with use of `Outlet` component to prevent full page reloads and the `Menu` from re-rendering unecessarily

src/frontend/src/components/Menu.tsx
```diff
@@ -0,0 +1,117 @@
...
+                    <EchoLogo className="h-[19px] w-[20px]" />
+                )}
+            </div>
+
+            {/* The Discover/Compare section */}
+            {activeLink && (
+                <div className="flex self-stretch justify-center gap-2">
+                    <NavTab
+                        href="/discover"
+                        isActive={activeLink === "discover"}
```
rachelekm: Is having the `activeLink` state necessary in this component? Can we just have something like `isActive={pathname.includes("discover")}` instead for simplicity?

src/frontend/src/components/Menu.tsx
```diff
@@ -0,0 +1,117 @@
...
+import ArrowDownIcon from "assets/icons/arrow-down.svg?react";
+
+interface Props {
+    compareCount: number;
+}
+
+const Menu = ({ compareCount }: Props) => {
+    const [activeLink, setActiveLink] = useState<string | null>(null);
+    // TODO: hook up with app-wide language state
+    const [selectedLanguage, setSelectedLanguage] = useState("en");
```
rachelekm, line 24: Just noting that I added an additional comment in the frontend translations issue, https://github.com/azavea/echo-locator/issues/640, to add language switching logic here

src/frontend/src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,21 @@
...
+        const listener = () => setMatches(media.matches);
+        media.addEventListener("change", listener);
+
+        return () => media.removeEventListener("change", listener);
+    }, [matches, query]);
+
+    return matches;
+};
+
+export default useMediaQuery;
```
rachelekm, line 27: This is nice to have!

rachelekm (approved):
This looks great! I think the hamburger menu on mobile view that opens into the modal makes the most sense. If implementing that mobile language modal does not fit within the scope of this issue let's make a followup to create a common mobile-first modal component.

The only other issue I found was that switching between compare and discover tabs fully reloads the page, causing a little flash for the nav bar and the tab page icon/title. Aaron and I discussed our plans for page routing (in comment above) so this will likely be resolved as we implement auth/private routes.

src/frontend/src/components/Menu.tsx
aaronxsu: Addressed in https://github.com/azavea/echo-locator/pull/674/commits/d4d247efc8d599b1593d202b71508d040dd8806a

aaronxsu:
@rachelekm I pushed up the following, please pull down and see if they fix the flickering issue on your end as well:

- https://github.com/azavea/echo-locator/pull/674/commits/713ff9d9b140bfb0f97eb6d178b11d2f45f1f1db: changes to declarative router definitions to make the route nesting more readable imo
- https://github.com/azavea/echo-locator/pull/674/commits/e3f6dffc5bb62233069f68f77587528c8fc65c48: refactors the `useMediaQuery` hook to fix the flickering issue. A new React 18+ hook `useSyncExternalStore` is used in the refactor, which seems to be [a use case by their docs](https://react.dev/reference/react/useSyncExternalStore#subscribing-to-a-browser-api)

aaronxsu:
>If implementing that mobile language modal does not fit within the scope of this issue let's make a followup to create a common mobile-first modal component.

I'd suggest we create a separate issue to cover it, since it will create a modal base component, plus this PR's branch is the base branch of some subsequent work (I'd prefer to have it merged asap to avoid rebases)

src/frontend/src/components/Menu.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    <EchoTextLogo className="h-[19px]" />
+                ) : (
+                    <EchoLogo className="h-[19px] w-[20px]" />
+                )}
+            </div>
+
+            {/* The Discover/Compare section */}
+            {(pathname.includes("discover") ||
+                pathname.includes("compare")) && (
+                <div className="flex self-stretch justify-center gap-2">
```
rachelekm, line 48:
```suggestion
            {/* The Discover/Compare section */}
            {(
                <div className="flex self-stretch justify-center gap-2">
```

Can we remove this conditional? I think we would want to include these tabs regardless of the page, even if neither of them display as "actively" selected

src/frontend/src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,27 @@
+import { useSyncExternalStore } from "react";
```
rachelekm, line 1: Nice refactor! Taking a look again this is kind of complex for conditionally displaying a handful of components based on screen size. A simpler alternative _could_ be adding tailwind breakpoints to conditionally show/hide components like below:
```
<div className="block md:hidden">Mobile hamburger menu</div>
<div className="hidden md:block">Desktop dropdown</div>
```

But in that approach Tailwind still loads all components in the html regardless if hidden or not. I asked gemini for a little pros-cons list:
| Criteria                      | Tailwind CSS (`hidden md:block`) | `useMediaQuery` Hook         |
|------------------------------|----------------------------------|-------------------------------|
| **JS Needed**                | ❌ No                             | ✅ Yes                        |
| **DOM Cost (Both Render?)**  | ❌ Yes (both render)              | ✅ No (only one)              |
| **Performance on Slow Devices** | ⚠️ Medium (due to DOM bloat)     | ✅ Better                     |
| **Accessibility Control**    | ✅ Good with classes              | ✅ Good                        |
| **Code Complexity**          | ✅ Simpler                        | ⚠️ More involved              |


We're deviating from tailwind classes but I think it's worth it for the DOM optimizations and to simplify component classnames. If we find it's causing some strange rendering or is difficult to troubleshoot in the future we can re-evaluate

src/frontend/src/App.tsx
```diff
@@ -1,23 +1,27 @@
...
+                <Route
+                    index
+                    element={<Navigate to="/discover" replace={true} />}
+                />
+                <Route path="discover" element={<Discover />} />
+                <Route path="compare" element={<Compare />} />
+                <Route path="components" element={<Components />} />
+            </Route>
+        </Routes>
+    </BrowserRouter>
```
rachelekm, line 24: This is a nice change, I agree more readable 👍

rachelekm (approved):
Thanks, @aaronxsu! I have one small nitpick to address but this looks great and good to go! I also opened #676 to capture the modal work.

The `useSyncExternalStore` definitely solved the nav bar flickering I was seeing! On toggling between tabs I'm still seeing some flickering of the page title/icon but that's because it's fully reloading the page which we'll address as part of implementing auth:

https://github.com/user-attachments/assets/1ae82aab-e4cc-4a96-8e05-50e8dcfb4a52

src/frontend/src/hooks/useMediaQuery.ts
aaronxsu: Hmmm, I don't think I followed the suggestion here- would you suggest we switch over to the tailwind approach?

rachelekm: Sorry no suggestion and think this is the correct approach! This was new to me and wanted to share what I learned about the trade-offs from both sides, but ultimately using `useSyncExternalStore` seems like the way to go 👍

src/frontend/src/components/Menu.tsx
aaronxsu: We could remove the conditional, but I would not recommend it. 

For example, the following sign-in page does not have those two tabs, which will likely be under `/login` or just `/`. Since we want to show pages and components based on app paths, the conditions in the component right now serve as a starting point. As we build more pages under more paths, this conditional section and other menu sections will change based on path or app state. So I recommend we leave it as is, and we iterate from here.

Screenshot 2025-07-08 at 5 51 32 PM

aaronxsu:
>On toggling between tabs I'm still seeing some flickering of the page title/icon but that's because it's fully reloading the page which we'll address as part of implementing auth:

@rachelekm I looked into it. The culprit was from the use of [Link](https://react-spectrum.adobe.com/react-aria/Link.html) component from React Aria in the previous implementation.

In https://github.com/azavea/echo-locator/pull/674/commits/f4b20bd677f05cc315359e03d55c24f9300fca0e, I switched to using [NavLink](https://reactrouter.com/api/components/NavLink) from React Router with some refactors. It should now look and work the same without the page reload. I am not sure what it means for accessibility though, since this is not a React Aria component anymore, but it's now a native React Router component.

src/frontend/src/hooks/useMediaQuery.ts
aaronxsu: Ahh, got it, no worries!

src/frontend/src/components/Menu.tsx
rachelekm: That makes sense, ok let's leave it in 👍

rachelekm:
> 
> @rachelekm I looked into it. The culprit was from the use of [Link](https://react-spectrum.adobe.com/react-aria/Link.html) component from React Aria in the previous implementation.
> 

That did the trick!! Great find, I think switching to the native NavLink React Router component is the way to go and since it renders an `<a>` element at least basic accessibility is covered.

rachelekm (approved):
Thanks for working through this, it looks great!

aaronxsu:
Thank you both for the reviews, changes, and suggestions. I added a changelog entry and squashed the fixup commits. Merging after CI.

PR #675, opened 2025-07-05
Add bar charts: Neighborhood Criteria and Commute Time

### LINKED ISSUE #652: Create commute time bar chart
Create commute time bar chart [according to figma designs](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=450-1603&m=dev), should also include [label version](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=450-1784&m=dev).

Units of commute time will be in minutes, the minimum is 0 and the maximum is 120 minutes. If bus/train mode is selected, a neighborhood will have a min and max commute range (for context, created from up to 4 transit results, "peak", "off-peak", "peak express", "off-peak express"). However, if car mode is selected a neighborhood only has one commute time and should be represented as a point.

Potentially could use [react-aria's `Meter` component](https://react-spectrum.adobe.com/react-spectrum/Meter.html#meter) to implement

### LINKED ISSUE #653: Create neighborhood criteria bar chart
Create neighborhood criteria bar chart [according to figma designs](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=450-1348&m=dev), should also include [label version](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=450-1365&m=dev).

Base bar should be 0-100, each neighborhood criteria category will be represented a percentage in state (ex. "crime_percentile": 97.0, "education_percentile": 77.0).

Potentially could use [react-aria's `Meter` component](https://react-spectrum.adobe.com/react-spectrum/Meter.html#meter) to implement

aaronxsu:
## Overview

This PR is branched out from https://github.com/azavea/echo-locator/pull/674:
- adds a customized Range component for Commute Time
- adds a customized Meter component for Neighborhood Criteria

@philippschmitt there might be styling rough edges that need your care.

### Checklist

- [ ] CI passes after rebase

### Demo

Screenshot 2025-07-05 at 7 31 14 PM

## Testing Instructions

 * Go to http://localhost:9966/components and make sure the bar charts section has example uses of Range and Meter components for Neighborhood Criteria and Commute Time

Resolves #652 
Resolves #653

### DIFF
--- a/src/frontend/src/components/base/Meter/Meter.styles.ts
+++ b/src/frontend/src/components/base/Meter/Meter.styles.ts
@@ -1,0 +1,22 @@
+import { tv } from "tailwind-variants";
+
+export const meterStyles = tv({
+    slots: {
+        root: "flex flex-col gap-3 w-full ",
+        labelContainer: "flex flex-col",
+        mainLabel: "text-sm text-gray-600",
+        valueLabel: "text-sm font-bold text-black",
+        track: "absolute top-1 h-1 w-full rounded-[var(--spacing-1)] bg-gray-300",
+        fill: "absolute h-[6px] rounded-[var(--spacing-1)]",
+        thumb: "absolute rounded-[var(--spacing-1)] top-[3px] h-4 w-2 -translate-y-1/2 left-[calc(50%-2px)] bg-gray-500 border border-white",
+    },
+    variants: {
+        status: {
+            low: { fill: "bg-low" },
+            belowAvg: { fill: "bg-below-avg" },
+            average: { fill: "bg-avg" },
+            aboveAvg: { fill: "bg-above-avg" },
+            high: { fill: "bg-high" },
+        },
+    },
+});
--- a/src/frontend/src/components/base/Meter/Meter.styles.ts
+++ b/src/frontend/src/components/base/Meter/Meter.styles.ts
@@ -2,10 +2,10 @@ import { tv } from "tailwind-variants";
 
 export const meterStyles = tv({
     slots: {
-        root: "flex flex-col gap-3 w-full ",
+        root: "flex flex-col gap-2 w-full ",
         labelContainer: "flex flex-col",
         mainLabel: "text-sm text-gray-600",
-        valueLabel: "text-sm font-bold text-black",
+        valueLabel: "text-sm font-bold text-black -mt-1 mb-[1px]",
         track: "absolute top-1 h-1 w-full rounded-[var(--spacing-1)] bg-gray-300",
         fill: "absolute h-[6px] rounded-[var(--spacing-1)]",
         thumb: "absolute rounded-[var(--spacing-1)] top-[3px] h-4 w-2 -translate-y-1/2 left-[calc(50%-2px)] bg-gray-500 border border-white",
--- a/src/frontend/src/components/base/Meter/Meter.styles.ts
+++ b/src/frontend/src/components/base/Meter/Meter.styles.ts
@@ -12,11 +12,11 @@ export const meterStyles = tv({
     },
     variants: {
         status: {
-            low: { fill: "bg-low" },
-            belowAvg: { fill: "bg-below-avg" },
-            average: { fill: "bg-avg" },
-            aboveAvg: { fill: "bg-above-avg" },
-            high: { fill: "bg-high" },
+            Low: { fill: "bg-low" },
+            "Below Avg": { fill: "bg-below-avg" },
+            Average: { fill: "bg-avg" },
+            "Above Avg": { fill: "bg-above-avg" },
+            High: { fill: "bg-high" },
         },
     },
 });
--- a/src/frontend/src/components/base/Range/Range.styles.ts
+++ b/src/frontend/src/components/base/Range/Range.styles.ts
@@ -7,6 +7,7 @@ export const rangeStyles = tv({
         mainLabel: "text-sm text-gray-600",
         rangeLabel: "text-sm font-bold text-black",
         track: "absolute top-1 h-1 w-full rounded-[var(--spacing-1)] bg-gray-300",
-        fill: "flex flex-col absolute h-[6px] rounded-[var(--spacing-1)] bg-gray-500",
+        fill: "flex flex-row justify-end absolute h-[6px] rounded-[var(--spacing-1)] bg-gray-500",
+        dots: "m-1 w-3",
     },
 });
--- a/src/frontend/src/components/base/Range/Range.styles.ts
+++ b/src/frontend/src/components/base/Range/Range.styles.ts
@@ -11,7 +11,7 @@ export const rangeStyles = tv({
     },
     variants: {
         variant: {
-            point: { fill: "rounded-full" },
+            point: {},
         },
     },
 });
--- a/src/frontend/src/components/base/Range/Range.styles.ts
+++ b/src/frontend/src/components/base/Range/Range.styles.ts
@@ -7,11 +7,6 @@ export const rangeStyles = tv({
         mainLabel: "text-sm text-gray-600",
         rangeLabel: "text-sm font-bold text-black",
         track: "absolute top-1 h-1 w-full rounded-[var(--spacing-1)] bg-gray-300",
-        fill: "absolute h-[6px] rounded-[var(--spacing-1)] bg-gray-500",
-    },
-    variants: {
-        variant: {
-            point: {},
-        },
+        fill: "flex flex-col absolute h-[6px] rounded-[var(--spacing-1)] bg-gray-500",
     },
 });
--- a/src/frontend/src/components/base/Range/Range.styles.ts
+++ b/src/frontend/src/components/base/Range/Range.styles.ts
@@ -1,0 +1,17 @@
+import { tv } from "tailwind-variants";
+
+export const rangeStyles = tv({
+    slots: {
+        root: "flex flex-col gap-3 w-full",
+        labelContainer: "flex flex-col",
+        mainLabel: "text-sm text-gray-600",
+        rangeLabel: "text-sm font-bold text-black",
+        track: "absolute top-1 h-1 w-full rounded-[var(--spacing-1)] bg-gray-300",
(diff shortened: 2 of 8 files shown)


src/frontend/src/components/base/Range/Range.styles.ts
```diff
@@ -0,0 +1,17 @@
+import { tv } from "tailwind-variants";
+
+export const rangeStyles = tv({
+    slots: {
```
aaronxsu, line 4: The slots here have similar style definitions as those in Meter. However, I separated them considering styling changes and tweaks down the road.

philippschmitt:
These are lovely 👏 
I added a small style tweak and `min-width` to the bar for edge cases where the meter is at 1% or the range is in the 1-5 minute range.

src/frontend/src/components/base/Meter/Meter.styles.ts
```diff
@@ -0,0 +1,22 @@
...
+export const meterStyles = tv({
+    slots: {
+        root: "flex flex-col gap-2 w-full ",
+        labelContainer: "flex flex-col",
+        mainLabel: "text-sm text-gray-600",
+        valueLabel: "text-sm font-bold text-black -mt-1 mb-[1px]",
+        track: "absolute top-1 h-1 w-full rounded-[var(--spacing-1)] bg-gray-300",
+        fill: "absolute h-[6px] rounded-[var(--spacing-1)]",
+        thumb: "absolute rounded-[var(--spacing-1)] top-[3px] h-4 w-2 -translate-y-1/2 left-[calc(50%-2px)] bg-gray-500 border border-white",
+    },
```
rachelekm, line 12: Slots!!! Still a little over my head but very cool to see it working here

src/frontend/src/pages/Components.tsx
```diff
@@ -237,6 +240,71 @@ const Components = () => {
...
+                            label="Commute"
+                            rangeText="58-60 min"
+                            start={58}
+                            end={60}
+                        />
+                        <Range
+                            label="Commute"
+                            rangeText="120 min"
+                            start={120}
+                            end={120}
```
rachelekm: We manually cap a neighborhood's commute time at 120 min and mark it as inaccessible as part of the ranking logic, but the actual values in state could include an end time of >120 min and it might be useful to still show that to the user in the label and represented visually in the Range. If possible, if an end value is >120 can we [include the "..." svg as a background image to the range bar like in the designs?](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=450-1678&m=dev)

src/frontend/src/pages/Components.tsx
```diff
@@ -237,6 +240,71 @@ const Components = () => {
...
+                        />
+                        <Range
+                            label="Commute"
+                            rangeText="0 min"
+                            start={0}
+                            end={0}
+                        />
+                        <Range
+                            label="Commute"
+                            rangeText="60 min"
```
rachelekm: Can we create the `rangeText` dynamically as part of the `Range` component so we can simplify redundant props?

src/frontend/src/pages/Components.tsx
```diff
@@ -237,6 +240,71 @@ const Components = () => {
...
+                        <Meter
+                            label="Schools"
+                            category="Below Avg"
+                            value={25}
+                        />
+
+                        <Meter label="Schools" category="Average" value={50} />
+                        <Meter
+                            label="Schools"
+                            category="Above Avg"
```
rachelekm: Can we have the category label string dynamically set in the `Meter` component and instead pass in something like a `displayCategoryLabel` boolean? Since we're already passing in a value and have existing logic to style by category group we should contain all that logic together for consistency.

rachelekm (changes requested):
These look GREAT! Really polished and easy to use too, great job. I added a couple comments that I think will help us use these more easily and fit one known edge case, let me know what you think

src/frontend/src/pages/Components.tsx
aaronxsu: This is good to know. I was wondering what those dots are for in this part of the design when working on this issue.

The more I think about it, the more I feel like we might want to break down the cases we are working with:

>the actual values in state could include an end time of >120 min and it might be useful to still show that to the user in the label
1. In **car mode**, we basically pass the same commute time value to `start` and `end`
    (a) For example, 15 min, we show a point, and the label is `15 min`
    (b) What if the commute time is equal to or more than 120 minutes? We show `At least 120 min` label (or something shorter like `> 120 min`, or maybe even the actual commute time), then what about the point in the bar chart since the chart has a cap of 120?
2. In **transit mode**, we pass different values to `start` and `end`
    (a) For example, 20 - 30 min, we show the range, and the label is `20 - 30 min`
    (b) If the lower end is below 120 minutes, but the higher end of the commute time is more than 120 minute, say the range is 110 to 130 min, the bar chart should start from 110 to 120 and show the three dots, then what about the label - `110 - more than 120 min` (this feels a bit strange)?
    (c) If both the lower and higher ends of the commute time range are above 120 minutes, what should we do for both the label and the range bar in the chart?

The more I think about the visual representation (aka, the bar charts) vs the labels (`10 min` or `100 - 120 min`), especially when looking at those map navigation apps (think, google maps), the more I feel like the text labels convey more information in a more direct way than the charts, since people may want to know what commute time they are dealing with. So this gets me thinking:
- We should provide the accurate range no matter if there is an arbitrary max or not. I think the max/cap we have here is mainly for the bar chart. 
- Then what about the bar chart in 1(b) and 2(c). I'd consider the value is "out of range" of the chart. When value is out of range, we usually don't see a chart in apps. So I'd recommend we hide the bar chart in such cases.

Please let me know what you think. I am open to suggestions. I can also hop on a huddle.

philippschmitt: Let me know what you think!

**Car mode**

> 1. In car mode, we basically pass the same commute time value to start and end

~~My understanding was we always show the `Peak` and `Off peak` durations, so there'd be a range for both car and transit. The [`Choose Traffic Conditions`](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=1391-4598&t=zpbv7JfjvDMyLwhz-11) toggle in the settings only affects the sorting, not the visual presentation.~~

Car doesn't make a peak/off-peak distinction, so we do need single points. All the below remains true.

> (b) What if the commute time is equal to or more than 120 minutes? We show At least 120 min label (or something shorter like > 120 min, or maybe even the actual commute time), then what about the point in the bar chart since the chart has a cap of 120?

As far as I understand, the algorithm considers neighborhoods with duration >= 120 min "out of range". For the edge case that we hit 120 exactly, I'd just say _it's over 2 hrs_.

image


**Transit mode**
> (b) If the lower end is below 120 minutes, but the higher end of the commute time is more than 120 minute, say the range is 110 to 130 min, the bar chart should start from 110 to 120 and show the three dots, then what about the label - 110 - more than 120 min (this feels a bit strange)?

If the low value is < 120 and the high value is > 120min, draw a bar, at least wide enough to fit the dots:
image

> (c) If both the lower and higher ends of the commute time range are above 120 minutes, what should we do for both the label and the range bar in the chart?

image

philippschmitt: @aaronxsu Updated my comment based on our meeting.

src/frontend/src/pages/Components.tsx
aaronxsu: https://github.com/azavea/echo-locator/pull/675/commits/94d24f019b0c24de57fbaf998063d796a04e6deb

src/frontend/src/pages/Components.tsx
aaronxsu: https://github.com/azavea/echo-locator/pull/675/commits/e4c8ed99ac1a437db307fb6a176faed095f88b95

src/frontend/src/pages/Components.tsx
aaronxsu: @philippschmitt thanks for the designs and the updates. I took a stab at adding the dots to the Range component, but just couldn't get them right. I constantly got tripped by styling.

If you have capacity, could you please take a look (especially the cases highlighted in the screenshot) and see if they could be styled correctly? I added a `TODO` to this commit where the dots SVG should be: https://github.com/azavea/echo-locator/pull/675/commits/94d24f019b0c24de57fbaf998063d796a04e6deb

Screenshot 2025-07-10 at 4 01 16 PM

philippschmitt: Just pushed 2 commits. I _think_ they do the trick but please test it on your machine as well. I'm out tomorrow but can revisit this Monday if further tweaks are needed.

aaronxsu: I pulled it down and tested locally. Worked very well on my end. Thanks!

aaronxsu:
@rachelekm thanks for the suggestions. Both Philipp and I pushed up some changes to address them. It's ready for another look when you get a chance!

src/frontend/src/pages/Components.tsx
aaronxsu: By the way, I just noticed (since I was too focused on the dots in the track): when the text range, e.g. 110-120 min, spans two rows, it might not look great when the range component is right next to things like School Quality, which has a category label (e.g. "Above Avg") always in one row. Though I am not sure how it will actually look in reality as the widths I gave on the components page are kind of arbitrary.

My suggestion for this would be to come back to this issue when we actually see it in the app page development.

rachelekm: These look so good! 👏

rachelekm (approved):
Thank you for working though all these edge cases and design tweaks, these look really good!!

aaronxsu:
Thanks for the reviews and the design tweaks. I squashed relevant commits and updated changelog. Merging after CI.

PR #677, opened 2025-07-09
Add remaining Input related components

### LINKED ISSUE #650: Create remaining reusable styled components
Implement following components according to figma design:
- [text inputs](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=814-21298&m=dev)
- [base label tags](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=865-2130&m=dev)
- [favorite button](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=286-14347&m=dev)

aaronxsu:
## Overview

This PR adds the following components and uses them on the components page
- Text input
- Number input
- Checkbox input
- Slider input

Tagging @philippschmitt here for design and styling care.

### Checklist

- [ ] CI passes after rebase

### Demo

Screenshot 2025-07-09 at 3 54 34 PM

Screenshot 2025-07-09 at 3 54 40 PM

## Testing Instructions

 * Go to http://localhost:9966/components and make sure the Inputs component section includes the above mentioned components.
 * Use them and make sure they perform as expected
 * Also try navigating to them using the tab key from the keyboard and make sure the navigations works

Resolves #650


philippschmitt:
Added a lil tweak ✨ but looks great overall and is fully navigatable by keyboard 👏

src/frontend/src/components/ImportanceSliders.tsx
```diff
@@ -0,0 +1,63 @@
+import { useState } from "react";
```
aaronxsu, line 1: I'm open to suggestions for the name of this component.

aaronxsu:
I rebased this PR on top of develop and added a changelog entry

src/frontend/src/components/ImportanceSliders.tsx
rachelekm: Looking back in git history these were referred to as "importance" (`crimeImportance`, `schoolsImportance`, etc) so I think this naming is intuitive and keeps the thread 👍

src/frontend/src/components/ImportanceSliders.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+        <div className="flex flex-col gap-8">
+            <Slider
+                step={1}
+                minValue={1}
+                maxValue={4}
+                label="Commute time"
+                value={factor.commute}
+                onChange={(value: number | number[]) => {
+                    if (Array.isArray(value)) return;
```
rachelekm:
```suggestion
```
Do we need this line in the `onChange`? This was confusing to see at first. I see how the type is defined as possibly an array in the SliderProps, but we're not implementing range sliders so in theory it should always just return type `number`. What do you think about taking it out for our own simplicity?

src/frontend/src/components/base/Checkbox/Checkbox.styles.ts
```diff
@@ -0,0 +1,26 @@
+import { tv } from "tailwind-variants";
+
+export const checkboxStyles = tv({
+    slots: {
+        root: "flex items-start gap-3",
+        box: "flex h-6 w-6 flex-shrink-0 items-center justify-center rounded-[var(--spacing-3)] border-2 transition",
+        labelContainer: "flex flex-col gap-1",
+        label: "text-rg font-bold text-gray-900",
+        description: "text-sm text-gray-600",
+    },
```
rachelekm, line 10: Nice! The checkbox example is where slots is finally clicking for me! One thing I'm thinking about is we now have two styling patterns: conditionally styled variants and static variants. If a style is static what do you think about keeping that in `base` and then only conditionally styled elements are defined in `slots`? That might keep our different base component style files more cohesive and be easier to get oriented for others jumping in

rachelekm (approved):
These look great! Approving since these are functionally good to go, but I left one nitpick comment and one thought on how we're setting up our style files. Let me know what you think!

src/frontend/src/components/ImportanceSliders.tsx
aaronxsu: Great question. I added this line to make TypeScript happy, otherwise, it would complain about types, even after we remove the types from `(value: number | number[])`. I would not recommend adding `// @ts-ignore`. I added a comment about it: https://github.com/azavea/echo-locator/pull/677/commits/e7dd1fce75087c89d767234013e6e44d44041029

src/frontend/src/components/base/Checkbox/Checkbox.styles.ts
aaronxsu: Hmmm, I don't think that's what the `base` is for. This might be easier to comprehend if we discuss in a call.

src/frontend/src/components/ImportanceSliders.tsx
rachelekm: I missed that taking this out caused a type error locally, thanks!

src/frontend/src/components/base/Checkbox/Checkbox.styles.ts
rachelekm: Documenting that Aaron and I discussed in a call which helped clarify my understanding of slots! Slots isn't necessarily _just_ to support conditionally styling a specific element, that functionality is an additional benefit to slots which fundamentally organize a component's styling using keys to target specific elements to style. It doesn't benefit us to separate out element styles that are static vs dynamic, instead if a component needs targeted styling (ie is more than one element) we should use the above slots format to style. If we don't need to style different parts of a component we should continue with `base` (like `TextField` styles).

rachelekm:
Thanks for clarifying the role of slots with me, updated my comments and this is good to go!

aaronxsu:
Thank you both for the reviews!

@philippschmitt I accidentally dropped your styling tweak when rebasing this PR on top of `develop` branch and force-pushed... my bad! Would you mind pushing that commit up again?

philippschmitt:
@aaronxsu Style tweaks are back up ✔️

aaronxsu:
Thank you! Merging this PR now.

PR #678, opened 2025-07-11
Add Neighborhood Card

### LINKED ISSUE #654: Create neighborhood card
Create neighborhood card [according to the figma designs. ](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=535-1264&m=dev) Card component should be able to conditionally render top thumbnail photo. The "close" button and footer "detail" and "tour nav" buttons will be used as part of the mobile map view, so either implement as a wrapper card component if time allows or leave off until Discover epic

aaronxsu:
## Overview

This PR composes the existing components together and creates the Neighborhood Card component.

Tagging @philippschmitt for styling care.

### Checklist

- [ ] CI passes after rebase

### Demo

Variations of the neighborhood card depending on props

Screenshot 2025-07-11 at 6 28 49 PM

### Notes

1. The neighborhood image used in this PR is from S3. I think they were supposed to be thumbnails used in the old frontend and thus they are very small, which explains the poor quality images in the screenshot. I am not sure yet what DB table column we will use for the neighborhood images. If we plan to still use the images on S3, then we will need to grab higher resolution images.
2. I did not have much luck grabbing the icon for `ECC Benefits` from the design. It needs to be swapped by the correct one @philippschmitt 

## Testing Instructions

 * Spin up the app
 * Visit the components page
 * Make sure the variations of the neighborhood card look good and make sense.

Resolves #654


philippschmitt:
I added the icon as you requested and made a few tweaks. 

One thing I noticed is that the cards aren't clickable. The ones with a `Detail` button could be fine, but the ones you labelled _Info only_ appear in lists where a user should be able to go to the neighborhood's detail on click. 
In case it's relevant: For accessible cards with multiple interactive elements (like the close/carousel buttons) I like wrapping the heading (so neighborhood name) in an `<a>` and using the _[pseudo-content trick](https://www.damianwajer.com/blog/accessible-card-component/)_

Regarding image resolution: That's unfortunate... In the existing app each image links to the higher resolution version on Wikipedia so I hope we can use those somehow?

rachelekm (changes requested):
These look good! I'll reiterate Philipp's comment above that there are no neighborhood cards that are "info only, no action", they are either navigable by explicit footer buttons or the entire card can be clicked to open details. I'll also note that the "basic info only" card's action is not simply a close button but would remove that neighborhood from the favorites selection/compare page (@philippschmitt correct me if I'm interpretting the designs wrong here) so we may want to have the option of a callback for the close button as well.

Re: grainy neighborhood images, let's continue to use S3. Looks like the images on prod were populated by running a script to fetch image and image metadata and then applying `optimize-image` library [as part of updating neighborhood data, ](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/update_data.sh)so we'll likely need to make a refactor in there for the new image sizes. I've opened #681 to capture that work.

rachelekm (approved):
Discussed this PR with Aaron IRL and addressed the following points:

- The different neighborhood card formats are all completed in this PR as well as support for callbacks for all actions buttons (including the close button). We should hold off on implementing any more action logic that extends beyond the card itself (like full-card click callbacks) until we add these cards into pages. We'll likely need to add refactors along the way as we build out the frontend logic and state so as long as styling and basic setup is covered we can more easily build pages with the least amount of re-work.

- For demo purposes, we'll use a neighborhood's wikimedia link directly instead of the existing S3 image. Once we refactor and run the image update script we can go back to using the S3 neighborhoods images directly.

Once we swap out the neighborhood image URLs this is good to go!

aaronxsu:
Thanks for the reviews and the suggestions.

I swapped the image used in the components demo page. I found this image by following the link in the DB table `api_neighborhood`, which takes you to a webpage, I then grabbed the image link from the webpage and used it in this PR. It's a good idea to run the script again and grab the high resolution images and upload them to S3 as these web image links can change in the long run.

```
street | https://commons.wikimedia.org/wiki/File:Washington_and_Harvard_Streets,_Brookline_Village_MA.jpg
```

Screenshot 2025-07-15 at 2 02 54 PM

aaronxsu:
For making a card clickable, I like that idea. I do think we should defer that to when we use the component with the clicking feature on the app page. It would be more efficient to adjust and update the component depending on the actual use case than to building it out for all scenarios now.

Regarding how to do it in the context of React Aria, I asked Gemini. It suggested two solutions, depending on if we want it be a button or a link. They make sense to me. We should revisit this when we get to the implementation.

Here is what I asked:

>I am using React Aria in my project. I have a card by having a styled div. Now I want this card be clickable while also using the accessibility features from React Aria. How to do it?

Here's the answer:

Of course. To make a styled `div` act as a clickable and accessible card, you should use the `useButton` hook from React Aria.

This hook applies all the necessary accessibility attributes and event handlers to your `div`, making it behave like a button for screen readers and keyboard users.

## How to Implement a Clickable Card
The key is to treat your card component as a button. You'll pass its props and a `ref` to the `useButton` hook, and then spread the returned `buttonProps` onto your `div`.

Here is a complete example of a `ClickableCard` component.

1. Create the `ClickableCard` Component

This component wraps a `div` and makes it accessible using `useButton`.

```js
import { useRef } from 'react';
import { useButton } from 'react-aria';

export function ClickableCard(props) {
  // A ref is needed to connect the hook to the DOM element.
  let ref = useRef(null);

  // The useButton hook provides accessibility and press behavior.
  // It takes the component's props and the ref.
  let { buttonProps } = useButton(props, ref);

  // The children are the content of the card.
  let { children } = props;

  return (
    // Spread the buttonProps onto your styled div.
    // This adds tabIndex, event listeners, and ARIA attributes.
    <div
      {...buttonProps}
      ref={ref}
      style={{
        border: '1px solid #ccc',
        borderRadius: '8px',
        padding: '16px',
        cursor: 'pointer',
        boxShadow: '0 2px 4px rgba(0,0,0,0.1)',
        // Style for the focus ring provided by React Aria
        outline: 'none',
        '--is-pressed-card-shadow': '0 1px 2px rgba(0,0,0,0.1)',
        '--is-focused-card-outline': '2px solid blue',
      }}
    >
      {children}
    </div>
  );
}
```
2. Use the Component in Your App

Now you can import and use `ClickableCard` just like you would a regular button, passing an `onPress` handler.

```js
import { ClickableCard } from './ClickableCard';

function App() {
  return (
    <ClickableCard onPress={() => alert('Card was pressed!')}>
      <h3>My Awesome Card</h3>
      <p>This entire card is clickable and fully accessible.</p>
    </ClickableCard>
  );
}

export default App;
```

## For Cards That Are Links
If your card should navigate to a new page instead of performing an action, you should semantically use an `<a>` tag and the `useLink` hook for better accessibility. The pattern is very similar.

```js
import { useRef } from 'react';
import { useLink } from 'react-aria';

export function LinkCard(props) {
  let ref = useRef(null);
  // Use the useLink hook for navigational elements
  let { linkProps } = useLink(props, ref);

  return (
    // Use an <a> tag as the base element
    <a
      {...linkProps}
      ref={ref}
      href={props.href} // Make sure to pass the href
      style={{
        /* card styles */
      }}
    >
      {props.children}
    </a>
  );
}

// Usage:
// <LinkCard href="/details">
//   <h3>Go to Details</h3>
// </LinkCard>
```

PR #685, opened 2025-08-05
Add state management migrate ADR

### LINKED ISSUE #610: Investigate routing logic to bring over from existing frontend
Following evaluation in #600, we'll be migrating away from Taui for the frontend implementation of project. There is existing logic in the current Taui-based version for fetching and deriving route and neighborhood ranking data, we should investigate the minimum logic between redux selectors and utility functions we need to bring over to replicate this functionality and how it should be implemented in new frontend.

rachelekm:
## Overview

Adds ADR investigating the minimum logic needed from the old codebase for neighborhood trip routing and ranking functionality and outlines a plan for how to implement in new frontend.

### Checklist

- [ ] CI passes after rebase

### Testing Instructions
- Confirm this ADR is clear as well as no formatting or grammatical issues

Resolves #610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add input components: Text, Number, Checkbox, Slider [#677](https://github.com/azavea/echo-locator/pull/677)
 - Add static translations support [#679](https://github.com/azavea/echo-locator/pull/679)
 - Add modal dialog component [#682](https://github.com/azavea/echo-locator/pull/682)
+- Add state management migrate ADR [#685](https://github.com/azavea/echo-locator/pull/685)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


aaronxsu (approved):
Looks good to me. @rachelekm and I quickly discussed and clarified on the selector functions to bring over. Two small tweaks to the write-up before merging:
1. I'd suggest we call out that the listed selectors under the "Updates to State Flow & Logic" section are the ones supporting data and map renderings, and they are based on a bunch of other selectors
2. If it's not too complex to update, I'd suggest we change the color of the selector function boxes in the first diagram to help tell the distinction between taui vs custom-echo selector functions

rachelekm:
Thanks for the review and discussion, @aaronxsu! I implemented those two points so that its more clear we'll be bringing in additional selectors to support the core ones that map to props as well as updated the diagram so nodes are colors by taui source code, our old additions, or taui source code adapted for ECHO.

I'll merge following a successful CI

PR #686, opened 2025-08-07
Add initial neighborhoods redux state

### LINKED ISSUE #655: Migrate Taui routing and networks logic
Following investigation from #610, migrate the needed Taui logic and bring the following calculated values into state: 1)Transit route details to all neighborhoods from an origin point for given selected analysis mode, 2) Commute time ranges to all neighborhoods from an origin point for all analysis modes, 3) Accessibility scores for each neighborhood for given selected analysis mode. We may also want to load and cache all network analysis json's on first destination to prevent unnecessary re-fetching.

Look into implementing with redux and [proxy-memoize ](https://redux.js.org/usage/deriving-data-selectors#proxy-memoize)following evaluation from #603.

rachelekm:
## Overview

Adds the initial redux setup to the app following decision from #685 including the basic neighborhoods state, fetching neighborhoods and neighborhoodsBounds data on discover page load.

One note is that neighborhoods api endpoints are authenticated and we need to pass session tokens as part of the GET request. I added a temporary way to use a session token as an env var (we'll need to set one for staging as well via tfvars) until we setup login on the frontend. Curious what you think of this strategy, I think the alternative is adjusting the endpoint permissions or just not fetching data and using some sample hard-coded values instead.

### Checklist

- [x] CI passes after rebase

## Testing Instructions

 * `scripts/update`
 * `scripts/server`
 * Login to the django admin and create a new Token obj for the admin user
 * Copy the newly created token value and add it to your .env as `VITE_DEVAUTHTOKEN={your-token-value}`
 * Navigate to the discover page of the app
 * Confirm redux logger is working and neighborhoods data (neighborhoods details and neighborhoods bounds) is successfully fetched and in state

Partially resolves #655


src/frontend/vite.config.ts
```diff
@@ -3,6 +3,8 @@ import react from "@vitejs/plugin-react";
 import tailwindcss from "@tailwindcss/vite";
 import svgr from "vite-plugin-svgr";
 
+const djangoProxyTarget = { target: "http://django:8085" };
```
aaronxsu, line 6: Would this be the same for all dev, staging, and production environments?

src/frontend/src/features/neighborhoods/types.ts
```diff
@@ -0,0 +1,55 @@
+import type { FeatureCollection, MultiPolygon, Point } from "geojson";
+
+interface NeighborhoodProperties {
+    crime_percentile: number;
```
aaronxsu: I'd double check the data types of these fields against the columns in the DB or the data models on the Django side, because it looks like at least the `crime_percentile` field is nullable- there might be more

src/frontend/src/pages/Discover.tsx
```diff
@@ -1,4 +1,25 @@
...
+        useAppSelector(({ neighborhoods }: RootState) => neighborhoods);
+
+    useEffect(() => {
+        // Fetch initial neighborhoods data
+        const isNeighborhoodDataEmpty =
+            !neighborhoods?.features || !neighborhoodBounds?.features;
+        // TODO: Refactor on adding login workflow
+        const testAuthToken = import.meta.env.DEV
+            ? import.meta.env.VITE_DEVAUTHTOKEN
+            : import.meta.env.VITE_STGAUTHTOKEN;
```
aaronxsu: I wonder if a simpler (and also temporary) solution would be just making it `VITE_AUTHTOKEN`, and we supply different values in the dotenv files for the corresponding environments- thinking about what to use on staging after we merge this PR. What do you think?

src/frontend/src/pages/Discover.tsx
```diff
@@ -1,4 +1,25 @@
+import { useEffect } from "react";
+import { useAppDispatch, useAppSelector, type RootState } from "../store/store";
+import { getNeighborhoodsAndBounds } from "src/features/neighborhoods/neighborhoodsThunk";
```
aaronxsu: I'd consider adding store and features to [here](https://github.com/azavea/echo-locator/blob/develop/src/frontend/vite.config.ts#L18) and [here](https://github.com/azavea/echo-locator/blob/develop/src/frontend/tsconfig.app.json#L28) to have simpler imports and avoid relative imports. If we go with this solution, I'd update the relative imports in other modules/files.

Additionally, I am not familiar with the naming convention of using `features` under `src`. Could you expand on it a bit?

src/frontend/src/features/neighborhoods/neighborhoodsThunk.ts
```diff
@@ -0,0 +1,21 @@
+import { createAsyncThunk } from "@reduxjs/toolkit";
+import {
+    fetchNeighborhoods,
+    fetchNeighborhoodBounds,
+} from "../../api/neighborhoods";
+
+export const getNeighborhoodsAndBounds = createAsyncThunk(
+    "neighborhoods/getNeighborhoodsAndBounds",
+    async (authToken: string | null) => {
```
aaronxsu: I wonder if passing in `authToken` is a temporary solution or a pattern we'd like to keep. I am assuming the former, but would like to double check

aaronxsu (commented):
Tested based on instructions and they work as advertised. I left some questions and suggestions- would like to hear what you think.

src/frontend/vite.config.ts
rachelekm: These [server options are only applied in dev ](https://vite.dev/config/server-options.html#server-options) and then our frontend static files are bundled with django and served from the same task definition so I don't think we need to setup any proxying for deployed environments.

src/frontend/src/features/neighborhoods/types.ts
rachelekm: Great catch, I copied these from the old codebase but can definitely be refined. Just added that fix in b3f1952187e51e96c4bf046d4995a7437044c998.

src/frontend/src/pages/Discover.tsx
rachelekm: I like that, much simpler! Added a commit adjusting to use one `VITE_AUTHTOKEN` between environments (6f6a80b9a4affdac7a7f889d0e392b5f56b13833).

src/frontend/src/pages/Discover.tsx
rachelekm: The `features` is new to me also as a convention, too. I got here because `reducers` didn't quite feel right now that it's slices + thunks + selectors and apparently redux recommends applications should[ structure files using a "feature folder" approach.](https://redux.js.org/style-guide/#structure-files-as-feature-folders-with-single-file-logic) We're still leaving react components in their own separate folder so we're not _really_ doing that recommendation, but at the time "features" seemed more accurate than "reducers". I just took a look around other projects with RTK and we still use the "reducers" convention even when its not longer just reducers so I think that naming convention might be the better choice after all, what do you think? I preemptively updated the naming to reducers as part of adjusting the path imports (good call!) in commits 7b427b6b2fce5b8f949ef936fd50df5f755d8dbc and ef2885d7978d630fd28a5e556f69c9a84ab488c4.

src/frontend/src/features/neighborhoods/neighborhoodsThunk.ts
rachelekm: The former! This is also how the old codebase was setup but once we setup login/auth on the frontend I think we should access the token from the store directly when needed vs passing it down.

rachelekm:
Thank you for the review, @aaronxsu! I made those adjustments and this is ready for another look.

src/frontend/src/pages/Discover.tsx
aaronxsu: Great. Now we will just need to update the dotnet file for staging.

src/frontend/src/pages/Discover.tsx
aaronxsu: Ah, thanks for pointing me to the RTK best practices guide! Now I understand where the naming/structuring convention is from. It seems like the `features` directory's sub-dirs group RTK related files as well as component or page files, which also makes sense to me. If we are moving towards this direction, we will maybe move the page files into the features directory? Though for the sake of time + budget and ease for others to spin up, I'd say we keep on with the `reducers` convention.

aaronxsu (approved):
Looks great to me. Thanks for addressing the questions. This is good to go.

src/frontend/src/pages/Discover.tsx
rachelekm: Yeah if we take this all the way I think base components and their styles could continue to be in their own dir but pages and any of their components would be in their respective "features" sub-dir. This is a pretty different setup than our other projects so similar to you above I don't think it's worth the time and budget. I agree let's keep our current setup and go with the reducers convention!

rachelekm:
Thanks, @aaronxsu! I'll rebase and merge following successful CI. Confirming I also just updated the staging `dotenv` file with a token I generated on staging for our shared echo admin user.

Issue #661, opened 2025-06-18
Add list-view neighborhood search filter


rachelekm:
Implement neighborhood search filter. See updated designs for map search overlay (on mobile, the list view displays a search input as well). Implement neighborhood search using [react-aria's Autocomplete](https://react-spectrum.adobe.com/react-aria/Autocomplete.html) with unique neighborhood names as options.

On text input, follow existing filtering implementation logic:
- update text search filter state
- recalculate list ranking on search submit

On search submit, map bounds should update to fit filtered list of neighborhoods

Issue #660, opened 2025-06-18
Add neighborhood list view


rachelekm:
Implement the neighborhood list view [according to figma designs for mobile and desktop](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=1069-7350&m=dev). There are 3 groupings for neighborhoods: "top 10" (derived from recommended), "recommended", and "too far" (ie. not recommended). Additionally this list view should reference the "list view" ranking order in state that groups together neighborhoods with the same name (example: "Cambridge").

The old codebase "paginated" this list and the neighborhoods to display was determined by page number, but this was not a backend pagination. We've kept the "load more" button in designs and are bringing the `selectPage` state forward but we might need more thought on if we should implement a load more button and at what number we should paginate.

Note there is static text that will need to be hooked up to our translations library.
The neighborhood search filter will be in a separate card.

philippschmitt:
SVG for the helper illustration:
![Image](https://github.com/user-attachments/assets/40933c07-a9a6-4970-a576-7e19ea9ae550)
[Image](https://github.com/user-attachments/assets/40933c07-a9a6-4970-a576-7e19ea9ae550)

PR #687, opened 2025-08-11
Create discover page component structures

### LINKED ISSUE #659: Add discover page mobile-first component structure
Create initial responsive component structure with a header menu component, translations subheader + "logout" on menu options open, the discover page subheader, and map and list view that matches figma designs for discover page. This issue to create initial layout of components and responsive design so these components can be empty or have void interactions.

aaronxsu:
## Overview

This PR:
- refactors the user profile header components
- creates mobile and desktop structures for the discover page using static component placeholders
- uses lazy-loading to optimize chunk sizes

Tagging @philippschmitt for styling and design cares

### Checklist

- [ ] CI passes after rebase

### Demo

Desktop list + map view
Screenshot 2025-08-11 at 10 04 40 AM

Mobile list view
Screenshot 2025-08-11 at 10 05 04 AM

Mobile map view

Screenshot 2025-08-11 at 10 04 53 AM

### Notes

1. Components used on the page are static to show where they are on the discover page as the scope of work here is to create page structures for both mobile and desktop
2. In mobile view, there is now a `mode` query parameter in the app URL to differentiate list or map view
3. I marked some todos in the code comments as they will be covered in other cards
4. In the designs, the search component is new, as it's not something we already created on the components page. I used the simple text input component we already have as a placeholder. The search component should be create separately as it has icons and maybe will also have typeahead (?)

On the discover page as placehoder:
Screenshot 2025-08-11 at 10 09 41 AM

In the designs:

Screenshot 2025-08-11 at 10 10 16 AM

## Testing Instructions

 * Spin up the app and go to discover page
 * Make sure the desktop view looks as expected
 * Switch to mobile view and make sure changing between map and list views work as expected with proper query param change in the app URL

Resolves https://github.com/azavea/echo-locator/issues/659

### DIFF
--- a/src/frontend/src/App.tsx
+++ b/src/frontend/src/App.tsx
@@ -1,46 +1,57 @@
+import { lazy, Suspense } from "react";
 import { BrowserRouter, Routes, Route } from "react-router";
 import { Navigate } from "react-router";
 
-import Root from "pages/Root";
-import Components from "pages/Components";
-import Discover from "pages/Discover";
-import Compare from "pages/Compare";
-
 import "./App.css";
-import LanguageRedirect from "components/LanguageRedirect";
+
+const Root = lazy(() => import("pages/Root"));
+const Components = lazy(() => import("pages/Components"));
+const Discover = lazy(() => import("pages/Discover/Discover"));
+const Compare = lazy(() => import("pages/Compare"));
+const LanguageRedirect = lazy(() => import("components/LanguageRedirect"));
+
+const LoadingFallback = () => <div>Loading page...</div>;
 
 const App = () => (
     <BrowserRouter>
-        <Routes>
-            {/* Redirect from the root path / to the default language page */}
-            <Route index element={<Navigate to="/en/discover" replace />} />
+        <Suspense fallback={<LoadingFallback />}>
+            <Routes>
+                {/* Redirect from the root path / to the default language page */}
+                <Route index element={<Navigate to="/en/discover" replace />} />
 
-            {/* Handle paths without a language prefix */}
-            {/* TODO: we need to update this block as we have more pages */}
-            <Route
-                path="discover"
-                element={<LanguageRedirect to="discover" />}
-            />
-            <Route path="compare" element={<LanguageRedirect to="compare" />} />
-            <Route
-                path="components"
-                element={<LanguageRedirect to="components" />}
-            />
+                {/* Handle paths without a language prefix */}
+                {/* TODO: we need to update this block as we have more pages */}
+                <Route
+                    path="discover"
+                    element={<LanguageRedirect to="discover" />}
+                />
+                <Route
+                    path="compare"
+                    element={<LanguageRedirect to="compare" />}
+                />
+                <Route
+                    path="components"
+                    element={<LanguageRedirect to="components" />}
+                />
 
-            {/* Language-specific pages */}
-            <Route path="/:lang" element={<Root />}>
-                {/* Handle /en, /es, etc. by redirecting to discover */}
-                <Route index element={<Navigate to="discover" replace />} />
+                {/* Language-specific pages */}
+                <Route path="/:lang" element={<Root />}>
+                    {/* Handle /en, /es, etc. by redirecting to discover */}
+                    <Route index element={<Navigate to="discover" replace />} />
 
-                {/* Rest of app pages */}
-                <Route path="discover" element={<Discover />} />
-                <Route path="compare" element={<Compare />} />
-                <Route path="components" element={<Components />} />
-            </Route>
+                    {/* Rest of app pages */}
+                    <Route path="discover" element={<Discover />} />
+                    <Route path="compare" element={<Compare />} />
+                    <Route path="components" element={<Components />} />
+                </Route>
 
-            {/* Catch-all */}
-            <Route path="*" element={<Navigate to="/en/discover" replace />} />
-        </Routes>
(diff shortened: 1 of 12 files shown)


src/frontend/src/App.tsx
```diff
@@ -1,46 +1,57 @@
...
 
-import Root from "pages/Root";
-import Components from "pages/Components";
-import Discover from "pages/Discover";
-import Compare from "pages/Compare";
-
 import "./App.css";
-import LanguageRedirect from "components/LanguageRedirect";
+
+const Root = lazy(() => import("pages/Root"));
```
aaronxsu, line 7: Added these since I saw the following when running `./scripts/update`:

```bash
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
```

src/frontend/src/pages/Discover/Desktop.tsx
```diff
@@ -0,0 +1,30 @@
+import { UserProfileSubheader } from "components/UserProfileSubheader";
+import { Place } from "src/enums";
+import discoverDesktopStyles from "./Discover.styles";
+import Neighborhoods from "./Neighborhoods";
+
+const DiscoverDesktop = () => {
```
aaronxsu, line 6: The desktop and mobile components' implementations may look very similar at the first glance but not 100% the same in how the html are structured. They may look more different after we implement more dynamic features or they may become more similar- it's hard to tell at this stage. It's also easier to combine and create abstract components from concrete ones rather than the other way around. So they are now separate for simplicity.

src/frontend/src/pages/Discover/Mobile.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { UserProfileSubheaderMobile } from "components/UserProfileSubheader";
+import { Place } from "src/enums";
+import discoverDesktopStyles from "./Discover.styles";
+import Neighborhoods from "./Neighborhoods";
+
+const DiscoverMobile = () => {
+    const [searchParams, setSearchParams] = useSearchParams();
+    const mode = searchParams.get("mode");
+
+    useEffect(() => {
```
aaronxsu, line 13: This is to make sure map view is the default in mobile view.

philippschmitt:
Looks nice! I tweaked the text a little bit (and found that Google Fonts wasn't importing the right weights, so I updated that too).

src/frontend/src/pages/Discover/Desktop.tsx
rachelekm: Appreciate the thought in this setup, I agree it's simple to reason about here and combine if ever needed

src/frontend/src/pages/Discover/Mobile.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useSearchParams } from "react-router";
+
+import { UserProfileSubheaderMobile } from "components/UserProfileSubheader";
+import { Place } from "src/enums";
+import discoverDesktopStyles from "./Discover.styles";
+import Neighborhoods from "./Neighborhoods";
+
+const DiscoverMobile = () => {
+    const [searchParams, setSearchParams] = useSearchParams();
+    const mode = searchParams.get("mode");
```
rachelekm: It took me a minute to wrap my head around this component but I think because of the variable naming. Since we're already using "mode" to refer to transportation mode elsewhere can we rename this to a more unique identifier like "display"? Similarly what do you think about renaming `searchParams` and `setSearchParams` to `routerSearchParams` and `setRouterSearchParams`?

src/frontend/src/pages/Discover/Neighborhoods.tsx
```diff
@@ -0,0 +1,81 @@
...
+                placeholder="Search neighborhoods"
+                value={mediumTextInput}
+                onChange={setMediumTextInput}
+            />
+
+            {/* Sub-title */}
+            <div className={subTitleContainer()}>
+                <p className={subTitle()}>Discover Neighborhoods</p>
+                <p className={description()}>
+                    Recommendations are based on your profile and selected trip
```
rachelekm, line 49: Just a note, as we start adding in text content and labels we should be sure to add their translations setup along the way. We don't need to have any translated content in the respective language json files, simply add the english as values since that will be populated all at once at the end.

Don't worry about updating this here as we still have a list view card to implement these details, https://github.com/azavea/echo-locator/issues/660

rachelekm (approved):
Nice this works well and looks really great!! I left a comment on some variable naming but other than that I think this is good to go!

Noting that we do have an issue explicitly for the search component https://github.com/azavea/echo-locator/issues/661 so a low-effort default input is perfectly fine.

src/frontend/src/pages/Discover/Mobile.tsx
aaronxsu: Updated: https://github.com/azavea/echo-locator/pull/687/commits/056d8cdcc4ed6579436de7421d67e1821fbf8b05#diff-e9be325b6c27863a51c9fd206f754f2433f80b2430d300fbdb7d3757a0ca13dfR11

aaronxsu:
Thanks for the style changes and for the review suggestions. I addressed them and rebased on develop. Merging after CI.

Issue #628, opened 2025-05-12
Migrate all 2025 data files in echo-locator-analysis-files bucket


rachelekm:
Following merge of #627 we're all setup to serve transit analysis files from BHA AWS account. While all known files have been brought over, the `*_times.dat` and `*_paths.dat` files are dynamically prefixed and fetched based on a users address and are not yet in our bucket. This issue is to sync our `echo-locator-analysis-files` bucket with Conveyal's for each network type (by us or them) and confirm a variety of addresses successfully fetch different data files following sync.

- Before this card confirm no large costs associated with this work
- Look into using `s5cmd` for batch s3 work
- Create any missing 2025 networks buckets on S3
- Sync buckets
- Following syncing buckets, update `NETWORK_URL_ROOT` env variable to be for new 2025 dir

rachelekm:
See [slack discussion f](https://element84.slack.com/archives/C04U48AN9N3/p1753708702446369?thread_ts=1746026488.344319&cid=C04U48AN9N3)or the new IDs for the updated results

rachelekm:
Confirming new peak, off-peak, off-peak-no-express, peak-no-express analyses are copied over to our 2025 networks buckets on S3. However, we don't have car analyses and the incoming *_paths.dat files has an updated header format that no longer works with taui's `parsePathsData` function so these files do not work with the old version of our app to test. We're reaching out to Anson to confirm formatting changes and if he has any suggestions to updating taui logic.

rachelekm:
There are now 5 sub-directories in the `echo-locator-analysis-files` bucket for 2025: peak, off-peak, off-peak-no-express, peak-no-express, and car. All of which have the new 2025 analysis files (a `request.json`, `transitive.json`, and necessary `*_times.dat` and `*_paths.dat` files).

The paths and times data files did not initially work in our old codebase, stalling in the Taui source code `parse-paths-data` utility function. See[ this slack discussion](https://element84.slack.com/archives/C04U48AN9N3/p1753974020748079?thread_ts=1746026488.344319&cid=C04U48AN9N3) for more information. Since the last generated analysis, path files' header format has updated as well as the PathWriter class, specifically improvements to the Taui path scorer like the addition of access and egress modes in the path details. Both of these updates affect the offset to begin parsing data in the `parse-paths-data` function as well as iterating through paths details. Refactoring to support these updates unblocks testing the new analysis files and confirming 2025 routing works in our old codebase and is validated against other trip planning platforms:

[...]

PR #693, opened 2025-08-21
Migrate route and travel time logic

### LINKED ISSUE #655: Migrate Taui routing and networks logic

rachelekm:
## Overview

Migrates state and selector logic from the old codebase, refactoring with the latest Taui source code as needed, to support generating neighborhood trip routes and commute times for each network mode. Largely follows state management plan proposed in #685 up to generating rankings.

### Checklist

- [x] CI passes after rebase

## Testing Instructions

 * First go back in commit history to get old values to compare against
     *  `git checkout 6f17e33e875b9f8fc215936d45387805cb74bafb` (Commit before removing old UI)
     *  `scripts/update`
     * Update local .env so that the network url var does not have the `VITE_` preface
     * `scripts/server`
 * Login and add the destination address 700 Boylston St. and set the network mode to "Peak"
 * Open up the console to view the networks state and note shape and values of `state.data.networks[index of "peak"].paths`, `state.data.networks[index of "peak"].pathsPerTarget`, `state.data.networks[index of "peak"].targets`,  and `state.data.networks[index of "peak"].travelTimeSurface` (These values are the parsed results of the path/time fetches).
 * Add the following line[ somewhere in the render of `Application` component](https://github.com/azavea/echo-locator/blob/6f17e33e875b9f8fc215936d45387805cb74bafb/src/frontend/src/components/application.js#L166-L199), then take note of shape and values of derived travel times in console output:
     * `console.log(p.neighborhoodTravelTimes)`
 * Return back to this branch:
     *  Adjust .env back
     * `scripts/update`
     * `scripts/server`
 * Open the console and view the networks state, compare the `paths`, `targets`, `pathsPerTarget`, and `targets` shape and values from the ones noted from the old codebase. Confirm they are the same.
 * Add the below snippet to the `Discover` page component to view selector outputs:
 ```
const times = useSelector(selectNeighborhoodTravelTimes);
    const routes = useSelector(selectNeighborhoodRouteGeoJson);
    if (networks && networks[activeMode].timesAndPathsDataReady) {
        console.log("routes", routes);
        console.log("times", times);
    }
 ```
 * Confirm the times output matches the shape and values from the old codebase's `neighborhoodTravelTimes` output.
 * Copy the feature collection from the routes output and input it in a visualizer like https://geojson.io/
 * Open the production echo app and input "700 Boylston St" as the address and set the mode to "Peak"
 * Confirm the rendered routes between the two look the same

Resolves #655


src/frontend/src/reducers/neighborhoods/neighborhoodsSlice.ts
```diff
@@ -18,18 +16,6 @@ export const neighborhoodSlice = createSlice({
...
-            { payload: neighborhoods }: { payload: Neighborhoods }
-        ) => {
-            state.neighborhoods = neighborhoods;
-        },
-        setNeighborhoodBounds: (
-            state,
-            { payload: neighborhoodBounds }: { payload: NeighborhoodBounds }
-        ) => {
-            state.neighborhoodBounds = neighborhoodBounds;
-        },
```
rachelekm, line 32: Took these out because I realized they're set via the `exraReducers` below and there's no scenario where we'd be updating neighborhoods/bounds data from the frontend.

src/frontend/src/reducers/neighborhoods/selectors/allNeighborhoodTravelTimes.ts
```diff
@@ -0,0 +1,51 @@
+// Refactored from old codebase neighborhoodTravelTimes at commit 6f17e33
+
```
rachelekm, line 2: This is refactored from the old selector to support travel times for all networks modes up front so we can create the commute time ranges for each neighborhood

src/frontend/src/reducers/neighborhoods/selectors/drawNeighborhoodRoute.ts
```diff
@@ -0,0 +1,140 @@
+// Refactored from old codebase at latest commit before introducing listings: 8ca3b94
+// New code copied from latest Taui v3 to support routes with MapLibre
+
```
rachelekm, line 4: This selector has changed significantly from the old one and is now largely the updated logic from Taui, which shifted from leaflet to mapbox. New Taui logic generates geojson's that better fit our maplibre map. When we're ready to add routes we should follow a [similar styling and application from Taui.](https://github.com/conveyal/taui/blob/dev/src/map/geojson-routes.js)

src/frontend/src/reducers/networks/utils/parsePathsData.ts
```diff
@@ -0,0 +1,101 @@
+// Refactored from latest Taui v3 parsePathsData to support new paths data format
+
```
rachelekm, line 3: This updated code [ is from this comment](https://github.com/azavea/echo-locator/issues/628#issuecomment-3184615811), written when migrating to new analysis files in #628

src/frontend/src/store/store.ts
```diff
@@ -6,7 +6,10 @@ import rootReducer from "./rootReducer";
 
 export const store = configureStore({
     reducer: rootReducer,
-    middleware: getDefaultMiddleware => getDefaultMiddleware().concat(logger),
+    middleware: getDefaultMiddleware =>
+        getDefaultMiddleware({
+            serializableCheck: false,
+        }).concat(logger),
```
rachelekm, line 12: Added to support the paths and times data as array buffers in state

rachelekm:
This PR is following the same implementation as our old codebase and Taui: 1. get network analysis details for each mode, 2. on origin change, fetch and parse new paths and times data for each mode 3. Derive all needed route and travel time information in selectors once each state network object in updated with new time/path data.

In bringing things over I realized some cons to keeping this approach:
 - The parsed times and paths data are large binary array buffers that make development difficult. They are not serializable, cause issues in logging and break redux dev tools.
 - Since we're deriving all of our ranking / routing data from selectors we don't have access to ranked neighborhood lists, neighborhood travel times, or neighborhood route in state. This makes debugging more difficult and if we're calling selectors from different components we may be re-running calculations as[ createSelector only memoizes the most recent set of parameters.](https://redux.js.org/usage/deriving-data-selectors#createselector-behavior)
 - Whenever a user switches between addresses the paths and times data will be fetched and everything re-calculated, even if previously fetching/calculated for that address.

While migrating things over I mulled over keeping the networks `request.json` and `transitive.json` in state and then separating out the times and paths, deriving necessary route and travel time information asynchronously within the thunk vs as selectors. This would allow us to throw away those large array buffers and set traveltime and route directly to state. We could also potentially wrap the thunk dispatch in a useMemo. This approach might need some more thought, but I think still might be a worthwhile change. Curious on your thoughts, regardless I think that would be a followup issue so that we can focus on the bringing over the needed logic in this card.

src/frontend/src/reducers/networks/types.ts
```diff
@@ -0,0 +1,182 @@
+export type NetworkModeOptions =
```
aaronxsu: I could see them being enums. I used `enum` directly in another recent project but typescript/linter complained, and thus I switched to: https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh. Up to you.

src/frontend/src/reducers/networks/types.ts
```diff
@@ -0,0 +1,182 @@
...
+    width: number;
+    zoom: number;
+}
+
+export type originPoint = { lon: number; lat: number };
+
+/**
+ * Route data types from old codebase
+ */
+export type LonLat = { lat: number; lon: number };
```
aaronxsu, line 17: The type looks the same as `originPoint` above. Is it possible to use one of them and remove the other?

src/frontend/src/reducers/networks/types.ts
```diff
@@ -0,0 +1,182 @@
...
+    route_type?: number;
+};
+
+export type TransitiveData = {
+    patterns: TransitivePattern[];
+    routes: TransitiveRoute[];
+    stops: TransitiveStop[];
+};
+
+export type QualifiedLeg = [TransitiveStop, TransitivePattern, TransitiveStop]; // [boardStopId, Pattern, alightStopId]
```
aaronxsu: I am confused by the code comment here- are they meant to be here?

src/frontend/src/reducers/networks/types.ts
```diff
@@ -0,0 +1,182 @@
...
+    stopIndex: number;
+    stop_id: string;
+    stop_name: string;
+    stop_lat: number;
+    stop_lon: number;
+};
+
+export type TransitivePattern = {
+    pattern_id: string;
+    patterns?: TransitivePattern[];
```
aaronxsu, line 43: Ah, interesting that we have a recursive type!

src/frontend/src/reducers/networks/networksSlice.ts
```diff
@@ -0,0 +1,85 @@
...
+                    action.error.message ??
+                    "Failed to fetch time and paths data.";
+            });
+    },
+});
+
+export const selectAllNetworksDataReady = createSelector(
+    [(state: RootState) => state.networks.networks],
+    networks =>
+        networks &&
```
aaronxsu: I'd suggest we have `!!netwoks` here instead, so that the type of the return here is purely `boolean` rather than `null | boolean`.

src/frontend/src/reducers/networks/networksThunk.ts
```diff
@@ -0,0 +1,107 @@
+// Refactored from old codebase loadDataset and fetchAllTimesAndPathsForCoordinate at commit 6f17e33
+// New source code copied from latest Taui v3
```
aaronxsu, line 2: Good idea to call it out here. Since Taui is under MIT license, it should be permissive. I'd suggest researching if there is anything additional we should include as well.

src/frontend/src/reducers/networks/networksThunk.ts
```diff
@@ -0,0 +1,107 @@
...
+
+                    networkDataByMode[network] = {
+                        ...requestData,
+                        ready: true,
+                        transitive: transitiveResponse.data,
+                    };
+                })
+            );
+
+            return networkDataByMode as Networks;
```
aaronxsu, line 50: Ah I see why this `as` is necessary here, because of the `Promise<any>` in the API method. This is fine for now. But I'd suggest we add a TODO to come back to it later when budget allows.

src/frontend/src/reducers/networks/utils/coordinateToIndex.ts
```diff
@@ -0,0 +1,33 @@
...
+) {
+    const pixel = toPixel(coordinate, zoom);
+
+    return {
+        x: (pixel.x - west) | 0,
+        y: (pixel.y - north) | 0,
+    };
+}
+
+export function coordinateToIndex(network: any, coordinate: originPoint) {
```
aaronxsu: There is a `Network` type, I wonder why `any` is used here instead.

src/frontend/src/reducers/networks/utils/parsePathsData.ts
aaronxsu: Looks like some trial and error efforts were invested 👏🏼 Good job to make this work for the new data!

src/frontend/src/reducers/networks/utils/parsePathsData.ts
```diff
@@ -0,0 +1,101 @@
...
+    const distinctPaths = [];
+    for (let i = 0; i < nDistinctPaths; i++) {
+        const nLegs = dataView.getInt32(byteOffset, true);
+        byteOffset += Int32Array.BYTES_PER_ELEMENT;
+
+        // Access and egress modes, each is a single byte
+        // const accessMode = dataView.getInt8(byteOffset);
+        byteOffset += 1;
+        // const egressMode = dataView.getInt8(byteOffset);
```
aaronxsu: If `accessMode` and `egressMode` are no longer useful, let's remove them

src/frontend/src/reducers/networks/utils/createTransitiveRoutes.ts
```diff
@@ -0,0 +1,128 @@
...
+        slice(network.targets, baseIndex, baseIndex + network.pathsPerTarget)
+    ).filter(tpi => tpi !== -1); // some destinations will not have any paths
+
+    // Cannot reach the destination
+    if (targetPathIndexes.length === 0) {
+        return [];
+    }
+
+    // Find stop
+    const findStop = (stopId: string) => find(td.stops, ["stop_id", stopId])!;
```
aaronxsu: I see the uses of non-null assertion operator `!` a few times in this file. As long as we are certain that the element it's finding (a stop here, and a pattern/route in the below method) exists at runtime, it should be fine. Otherwise, we might see crashes.

src/frontend/src/reducers/neighborhoods/neighborhoodsSlice.ts
aaronxsu: Agreed. We should only fetch them once when the app initializes.

aaronxsu (commented):
What a great job migrating all these logic over and the refactoring 👏🏼 

I tested based on testing instructions and it works as advertised. I left some questions/suggestion in code and would like to hear what you think!

>This would allow us to throw away those large array buffers and set traveltime and route directly to state.

I'd prefer this since it would be cleaner. I'd consider it if budget allows or if we need it to have better app performance.

>regardless I think that would be a followup issue so that we can focus on the bringing over the needed logic in this card.

Makes sense to me.

src/frontend/src/reducers/networks/types.ts
rachelekm: Good catch, I created `originPoint` and then `LonLat` was copied over from the source code. I didn't notice these were the same, I just updated so that `LonLat` is the type used throughout the frontend and scrapped `originPoint` in 87a132286b8e28e4aa758004356e38329f5c6d72

src/frontend/src/reducers/networks/types.ts
rachelekm: These comments were copied from the source code purely as added context but I noticed were not actually even using `QualifiedLeg` or `QualifiedPath` anymore so I took just removed them in 87a132286b8e28e4aa758004356e38329f5c6d72

src/frontend/src/reducers/networks/networksThunk.ts
rachelekm: Added a TODO as part of 31dc63304019fb034e154381993743620cc99011

src/frontend/src/reducers/networks/utils/coordinateToIndex.ts
rachelekm: Whoops a leftover when copying this fn from Taui, this is updated the `Network` in 31dc63304019fb034e154381993743620cc99011

src/frontend/src/reducers/networks/utils/parsePathsData.ts
rachelekm: Sounds good, I simplified this comment to just the needed context in 31dc63304019fb034e154381993743620cc99011

src/frontend/src/reducers/networks/utils/createTransitiveRoutes.ts
rachelekm: I went back in a set up some guardrails so that I could take out all these non-null operators in 31dc63304019fb034e154381993743620cc99011

src/frontend/src/reducers/networks/types.ts
rachelekm: Cool this was a good suggestion! Added bonus is that this helps bring the dashed syntax from the file paths and the formatted labels to be displayed together using one consistent key syntax. That work to switch this over to enums is in 6866f4c286852f0d34b0f23050c1315feac04fdd

src/frontend/src/reducers/networks/networksThunk.ts
rachelekm: There doesn't seem to be official guidance other than ensuring that the license is brought forward into the new repo. General internet consensus is that a linking citation in the source files is good practice as well. I added more formal citations in these comments and then brought the Taui license over into the frontend dir in f8ea6679cad2aa604a7b6184211edaff36345a41

rachelekm:
Thank you for the thorough review, @aaronxsu! I addressed each one in the individual comments, this is all set for another look! Noting that I rebased on develop so you will need tot force pull this branch and then run `scripts/update` once again.

aaronxsu (approved):
Thanks for addressing the comments. This is now looking very good 👏🏼 Good to go after a changelog entry 🚀

rachelekm:
Thank you!! CHANGELOG entry up in d37fbeb731b3d394b2c77365526454e1837c42d3, will merge shortly

Issue #656, opened 2025-06-18
Implement neighborhood rank list


rachelekm:
Following migration of Taui routing logic and the algorithm, hook everything together to create the resulting neighborhood rank list. In addition to a neighborhood rank list in state we should also have a parallel "list view" neighborhood rank list that  re-orders so that neighborhoods with the same name ("cambridge") are grouped together.

Look into implementing with redux and [proxy-memoize ](https://redux.js.org/usage/deriving-data-selectors#proxy-memoize)following evaluation from https://github.com/azavea/echo-locator/issues/603.

PR #694, opened 2025-08-21
Add neighborhood ranking logic

### LINKED ISSUE #657: Migrate neighborhood scoring logic
Following investigation from https://github.com/azavea/echo-locator/issues/610, migrate "algorithm" logic into new state structure so that neighborhood scores are available in state, [likely all the logic from the neighborhoods-sorted-with-routes selector.](https://github.com/azavea/echo-locator/blob/master/taui/src/selectors/neighborhoods-sorted-with-routes.js#L50-L170)

Look into implementing with redux and [proxy-memoize ](https://redux.js.org/usage/deriving-data-selectors#proxy-memoize)following evaluation from https://github.com/azavea/echo-locator/issues/603.

rachelekm:
## Overview

Builds off of #693 and migrates the neighborhood ranking logic from the old codebase.

Adjusts the selector output so that it returns two lists (sorted, recommended neighborhood zip codes and sorted, "too far" neighborhood zip codes) for our updated UI. This output is the "true" list from which we will derive the grouped, "paginated" sorted-by-rank list in #656.

Note: I didn't touch the specifics of weighting/scores beyond moving the ranking algorithm into its own function for better clarity. This is the "core" of the app that's been refined over time with discipline-specific knowledge and iterating on it would likely be high effort that is not part of the frontend rewrite.

### Checklist

- [x] CI passes after rebase

## Testing Instructions

 * Since these lists are not in state we need to print the results of the selector to the console.
       - Add the below line to the `Discover` component to call the selector: 
               `useAppSelector(neighborhoodsSortedWithRoutes);`
       - Add the following print statement to the selector right above the return to view results:
               `console.log("recommended:", rankedLists.recommended, "Too far:", rankedLists.tooFar)`
* Spin up this branch and open the console
* Alongside this window, visit the production app
     * Ensure all the user profile importance selections are set to the defaults (somewhat important commute time and not important school and safety)
     * Select the destination 700 Boylston St. and network mode "Peak"
     * View the list results in the sidebar
 * Confirm length of the recommended list is largely the same as the recommendations count on production
 * Verify the order of the neighborhood zip codes make sense between the production site and local `recommended` list.
           * If commute time is weighted highest when scoring there may be some ranking differences because route and travel times have changed in 5+ years. To test exact matches between the old codebase and new, set commute time to "not important" in production and to "1" in userProfile slice of local dev.
           * Confirm recommendations are direct match between two sites

Resolves #657

### DIFF
--- a/src/frontend/src/reducers/neighborhoods/selectors/neighborhoodsSortedWithRoutes.ts
+++ b/src/frontend/src/reducers/neighborhoods/selectors/neighborhoodsSortedWithRoutes.ts
@@ -1,0 +1,200 @@
+// @flow
+import lonlat from '@conveyal/lonlat'
+import concat from 'lodash/concat'
+import filter from 'lodash/filter'
+import findIndex from 'lodash/findIndex'
+import get from 'lodash/get'
+import includes from 'lodash/includes'
+import orderBy from 'lodash/orderBy'
+import pullAt from 'lodash/pullAt'
+
+import {
+  BOOST_DOWNTOWN_RESULT_PLACE,
+  DEFAULT_ACCESSIBILITY_IMPORTANCE,
+  DEFAULT_CRIME_IMPORTANCE,
+  DEFAULT_SCHOOLS_IMPORTANCE,
+  DOWNTOWN_AREAS,
+  MAX_IMPORTANCE,
+  RESULTS_WITH_DOWNTOWN
+} from '../constants'
+import {NeighborhoodProperties} from '../types'
+import scale from '../utils/scaling'
+import { createSelector } from '@reduxjs/toolkit'
+
+import selectNeighborhoodRoutes from "../../networks/selectors/networkNeighborhoodRoutes";
+import selectNeighborhoodTravelTimes from "./allNeighborhoodTravelTimes";
+import type { RootState } from 'store/store'
+
+
+// ~130km estimate of maximum straight-line distance drivable in area within 120 minutes
+const MAX_DISTANCE = 130000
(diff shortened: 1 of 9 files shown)


src/frontend/src/reducers/neighborhoods/selectors/neighborhoodsSortedWithRoutes.ts
```diff
@@ -0,0 +1,148 @@
...
+        const recommendedNeighborhoodsList: Feature<
+            Point,
+            NeighborhoodProperties
+        >[] = [];
+        const tooFarNeighborhoodsList: Feature<
+            Point,
+            NeighborhoodProperties
+        >[] = [];
+
+        const useTransit = activeNetworkMode === "car";
```
aaronxsu: I am a bit confused by this line. Why is `useTransit` true if the active mode is car?

aaronxsu (approved):
Tested based on the instructions. Works well and looks good 👍🏼

src/frontend/src/reducers/neighborhoods/selectors/neighborhoodsSortedWithRoutes.ts
rachelekm: Whoops, sure enough this should be `!==` and not `===`. Thanks for catching!

rachelekm: Fix added in [442a880](https://github.com/azavea/echo-locator/pull/694/commits/442a88078b8064f617c70a07192669f7589454ed)

PR #698, opened 2025-08-27
Add formatted ranking lists to state

### LINKED ISSUE #656: Implement neighborhood rank list

rachelekm:
## Overview

Brings the derived ranked neighborhoods into state and formats lists for both map and list view. Redux state now has a "true", ungrouped top ten list to use for the map top ten tour as well as three grouped lists for the list view: top ten, recommended, and too far. The groupings within each list were created by traversing neighborhoods list and, for the first instance of a unique neighborhood name, finding all neighborhood matches close in ranking with a matching name and creating a group of those ids at the initial index.

I defined "close in ranking" as <= a difference of 50 indexes. This number is mostly from visually inspecting the grouped lists  at different values of index proximities. The max difference of 50 felt like it captured all matching name neighborhoods within a reasonable chunk of the total list (~240 neighborhoods) while also not creating a false sense of rank inflation

### Checklist

- [x] `README.md` updated if necessary to reflect the changes
- [x] CI passes after rebase

### Demo
Screenshot 2025-08-27 at 10 30 05 AM
Screenshot 2025-08-27 at 10 29 52 AM

## Testing Instructions

 * `scripts/server`
 * open up discover page in browser and open dev tools to see console
 * Confirm 4 separate lists of grouped neighborhoods lists are in state 
 * Open up production app and set values to defaults, set origin to 700 Boylston st, and mode to Peak
 * Compare ranked list in production to the ranked list in this branch's state and confirm these are aligned
 * Confirm groupings of zipcodes all have the same neighborhood name and there are no name matches close together in lists
      * I found this easiest by temporarily switching to push the neighborhood name to state list vs the zipcode, done with the following diff:
```diff
      diff --git a/src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts b/src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
index f7f6b78..1907a68 100644
--- a/src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
+++ b/src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
@@ -65,7 +65,7 @@ export const getRankedNeighborhoodLists =
                 }
 
                 const name = nameByZipcodeMap && nameByZipcodeMap[zip];
-                const group = [zip];
+                const group = [name];
                 alreadyGroupedIndexes.push(i);
 
                 list.forEach((compareId, j) => {
@@ -77,7 +77,7 @@ export const getRankedNeighborhoodLists =
                     const inProximityToFirstTownName = Math.abs(j - i) <= 50;
 
                     if (compareName === name && inProximityToFirstTownName) {
-                        group.push(compareId);
+                        group.push(compareName);
                         alreadyGroupedIndexes.push(j);
                     }
                 });

```

Resolves #656

### DIFF
--- a/src/frontend/src/pages/Discover/Discover.tsx
+++ b/src/frontend/src/pages/Discover/Discover.tsx
@@ -1,6 +1,9 @@
 import { useEffect } from "react";
 
-import { getNeighborhoodsAndBounds } from "reducers/neighborhoods/neighborhoodsThunk";
+import {
+    getNeighborhoodsAndBounds,
+    getRankedNeighborhoodLists,
+} from "reducers/neighborhoods/neighborhoodsThunk";
 import { useAppDispatch, useAppSelector, type RootState } from "store/store";
 import useMediaQuery from "hooks/useMediaQuery";
 import Desktop from "./Desktop";
@@ -9,7 +12,10 @@ import {
     getAllTimesAndPaths,
     getNetworks,
 } from "reducers/networks/networksThunk";
-import { setOrigin } from "reducers/networks/networksSlice";
+import {
+    selectAllNetworksDataReady,
+    setOrigin,
+} from "reducers/networks/networksSlice";
 
 const Discover = () => {
     const dispatch = useAppDispatch();
@@ -75,6 +81,13 @@ const Discover = () => {
             dispatch(getAllTimesAndPaths(origin));
         }
     }, [neighborhoods, networks, origin]);
+
+    const networksDataIsReady = useAppSelector(selectAllNetworksDataReady);
+    useEffect(() => {
+        if (networksDataIsReady) {
+            dispatch(getRankedNeighborhoodLists());
+        }
+    }, [networksDataIsReady]);
     // --------------------------------------
 
     return isDesktop ? <Desktop /> : <Mobile />;
--- a/src/frontend/src/pages/Discover/Discover.tsx
+++ b/src/frontend/src/pages/Discover/Discover.tsx
@@ -31,6 +31,7 @@ const Discover = () => {
         networks,
         origin,
     } = useAppSelector(({ networks }: RootState) => networks);
+    const networksDataIsReady = useAppSelector(selectAllNetworksDataReady);
     const isDesktop = useMediaQuery("(min-width: 768px)");
 
     // TODO: Refactor below following login and user profile
@@ -82,7 +83,6 @@ const Discover = () => {
         }
     }, [neighborhoods, networks, origin]);
 
-    const networksDataIsReady = useAppSelector(selectAllNetworksDataReady);
     useEffect(() => {
         if (networksDataIsReady) {
             dispatch(getRankedNeighborhoodLists());
--- a/src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
+++ b/src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
@@ -3,6 +3,14 @@ import {
     fetchNeighborhoods,
     fetchNeighborhoodBounds,
 } from "../../api/neighborhoods";
+import {
+    selectNeighborhoodNameByZipcode,
+    setRankedNeighborhoodLists,
+} from "./neighborhoodsSlice";
+import { type AppDispatch, type RootState } from "store/store";
+import neighborhoodsSortedWithRoutes from "./selectors/neighborhoodsSortedWithRoutes";
+import type { RankedNeighborhoodsLists } from "./types";
+import groupRankingsByLikeNeigborhoodName from "./utils/groupRankingListsByName";
 
 export const getNeighborhoodsAndBounds = createAsyncThunk(
     "neighborhoods/getNeighborhoodsAndBounds",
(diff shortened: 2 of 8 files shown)


src/frontend/src/pages/Discover/Discover.tsx
```diff
@@ -75,6 +81,13 @@ const Discover = () => {
             dispatch(getAllTimesAndPaths(origin));
         }
     }, [neighborhoods, networks, origin]);
+
+    const networksDataIsReady = useAppSelector(selectAllNetworksDataReady);
```
aaronxsu: I'd place this line closer to those `useAppSelector` calls at the top- up to you.

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
```diff
@@ -19,3 +23,79 @@ export const getNeighborhoodsAndBounds = createAsyncThunk(
...
+        groupedNeighborhoodsLists.topTen = neighborhoodsList.recommended.slice(
+            0,
+            10
+        );
+
+        // Group neighborhoods with the same name if close together in ranking
+        // Group separately between "top ten" and remaining list
+        const groupByMatchingNameAndProximity = (list: string[]): any => {
+            const groupedList: (string | string[])[] = [];
+            const alreadyGroupedIndexes: number[] = [];
```
aaronxsu: I'd use a `Set` instead of an array here, since this is mainly used for lookups later, and `set.has()` is generally considered more efficient than `array.includes()`

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
```diff
@@ -19,3 +23,79 @@ export const getNeighborhoodsAndBounds = createAsyncThunk(
...
+        const neighborhoodsList = neighborhoodsSortedWithRoutes(state);
+
+        groupedNeighborhoodsLists.topTen = neighborhoodsList.recommended.slice(
+            0,
+            10
+        );
+
+        // Group neighborhoods with the same name if close together in ranking
+        // Group separately between "top ten" and remaining list
+        const groupByMatchingNameAndProximity = (list: string[]): any => {
```
aaronxsu: There might be an opportunity to move this function out of `getRankedNeighborhoodLists` and be on its own. I'd also suggest changing the return type from `any` to be more specific.

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
```diff
@@ -19,3 +23,79 @@ export const getNeighborhoodsAndBounds = createAsyncThunk(
...
+            const groupedList: (string | string[])[] = [];
+            const alreadyGroupedIndexes: number[] = [];
+
+            const nameByZipcodeMap =
```
aaronxsu: It looks like this map gets created every time the function runs, which might not be the most efficient way, since my understanding is that the mapping between zip codes and town names don't actually change in our use case. I'd suggest we create this once, and pass it in as an argument.

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
```diff
@@ -19,3 +23,79 @@ export const getNeighborhoodsAndBounds = createAsyncThunk(
...
+            list.forEach((zip, i) => {
+                if (alreadyGroupedIndexes.includes(i)) {
+                    return;
+                }
+
+                const name = nameByZipcodeMap && nameByZipcodeMap[zip];
+                const group = [zip];
+                alreadyGroupedIndexes.push(i);
+
+                list.forEach((compareId, j) => {
```
aaronxsu: I'd improve the efficiency in this nested loop. If I understand correctly, in this loop, we might not need to loop from the beginning, but rather could just start from outer loop's (outer_index + 1). Additionally, since the "furthest" we should look at is only 50 indices away from the starting point (which is outer loop's index), so we might as well just "stop" before it reaches (outer_index + 50).

Additionally, I'd put 50 as a constant at the top.

aaronxsu (changes requested):
Works as advertised 👍🏼 I left suggestions related to some potential performance improvement ideas in a method to consider. I'd like to hear what you think!

src/frontend/src/pages/Discover/Discover.tsx
rachelekm: Updated in 5e4dda5d870ca94ae61cac512688fbf9ddbf627a

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
rachelekm: Great call! Pulled this out into its own selector in 89bbc024d182fff298d427c6089c918478953ab7

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
rachelekm: Great suggestion I implemented those improvements in 9fcc65b5e09765a7c5d36a73d038b5266ff6676c

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
rachelekm: Moved this into its own function and updated the types in 9fcc65b5e09765a7c5d36a73d038b5266ff6676c

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
rachelekm: Sounds good, made that update in 9fcc65b5e09765a7c5d36a73d038b5266ff6676c

rachelekm:
Thanks, @aaronxsu! These were great suggestions that really improves the readability/logic of this selector. Implemented your suggestions in the fixups above and this is ready for another look

aaronxsu (approved):
Thanks for addressing the feedback. This is looking great! Good to go 🚀

PR #699, opened 2025-08-27
Refactor networks data state to support place comparisons

### LINKED ISSUE #697: Refine source code state flow
This is a followup from #693. We brought over the trip logic in roughly the same implementation as the source code, by fetching and keeping network/paths/times data in state and then deriving all needed route and travel time information in selectors. However, this[ implementation doesn't fit our project use case well and will be hard to maintain in the future.](https://github.com/azavea/echo-locator/pull/693#issuecomment-3211535905)

We should instead refactor our state management plan so that neighborhood travel times and routes are accessible in state. This could be done by migrating `networkNeighborhoodRoutes`, `allNeighborhoodTravelTimes` selectors into utility functions that are called asynchronously within the `getAllTimesAndPaths` thunk. The thunk payload would then return the derived neighborhood travel times and paths vs the source array buffers.

We would also need to adjust the incoming `neighborhoodsSortedAsRoutes` and `displayPageNeighborhoods` selectors into functions as well that could be asynchronously called within a new action, dispatched following complete travel time and path data. This would allow us to have our ranked list of neighborhoods in state.

rachelekm:
## Overview

In reviewing #693 I missed that we need all travel/route info for all user destinations to support comparisons in network details and compare tab. This PR restructure networks' time and paths data state shape so that routes and travel times are available for every destination as well as network mode at one time. This supports the [route and commute time comparisons in neighborhood detail.](https://github.com/azavea/echo-locator/issues/697#issuecomment-3228496318)

Along the way simplifies the networks data implementation so that neighborhood routes and travel times are derived as part of the thunk to fetch path and times data (vs individual selectors). This refactor is [following the discussion ](https://github.com/azavea/echo-locator/pull/693#issuecomment-3211535905) in #693 and allows for simplified state management and greater maintainability as we populate components with data.

### Checklist

- [x] `README.md` updated if necessary to reflect the changes
- [x] CI passes after rebase

### Demo
New networks state structure:
- `networks` state no longer has parsed buffer responses from path and times endpoints
- Separate `timesAndRoutes` state in `networks` slice for derived route and travel time data, indexed by place label and then network mode:
Screenshot 2025-08-27 at 3 18 32 PM

- No regressions with ranked list for new network state setup
Screenshot 2025-08-27 at 10 30 05 AM

## Testing Instructions

 * `scripts/server`
 * Open up discover page in browser and open dev tools to see console
 * Confirm `networks.networks` state no longer has array buffer values (`paths`, `targets`, and other keys like `stops` or `travelTimeSurface`)
 * Confirm new state in `networks` slice, `timesAndRoutesData`, populated with key that matches the active destination value in user profile
 * Confirm values of this destination key is route and travel time data organized for every network mode
 * Alongside this window open up production app and set values to defaults, set origin to 700 Boylston st, and mode to Peak
 * Compare ranked list in production to the ranked list in this branch's state and confirm there are no regressions

Resolves #697


src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodRoutes.ts
```diff
@@ -0,0 +1,74 @@
...
+import memoize from "lodash/memoize";
+
+import type { Neighborhoods } from "src/reducers/neighborhoods/types";
+import type {
+    Location,
+    LonLat,
+    NeighborhoodRoutePaths,
+    NeighborhoodRoutes,
+    NetworkAndTimeAndPathsData,
+} from "../types";
```
aaronxsu: I'd move these imports into the import line above, since all of them import from the same module.

src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodRoutes.ts
```diff
@@ -0,0 +1,74 @@
...
+        `${n.name}-${i}-${lonlat(s.position).toString()}-${lonlat(e.position).toString()}`
+);
+
+export default function createNetworkNeighborhoodRoutes(
+    network: NetworkAndTimeAndPathsData,
+    start: LonLat,
+    neighborhoods: Neighborhoods
+): NeighborhoodRoutes {
+    const routes: NeighborhoodRoutes = [];
+    neighborhoods.features.map((neighborhood, neighborhoodIndex) => {
```
aaronxsu: This might be from the old codebase or Taui, but the use of `map` here seems a bit odd to me, since it uses the side-effect here to create `routes`- I'd either replace it by a `reduce` to have it ready in one go or a `forEach` as a proper loop.

src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodTravelTimes.ts
```diff
@@ -0,0 +1,22 @@
...
+    network: NetworkAndTimeAndPathsData,
+    neighborhoods: Neighborhoods
+): number[] {
+    return neighborhoods.features.map(neighborhood => {
+        const surface = network.travelTimeSurface;
+        const idx = coordinateToIndex(
+            network,
+            lonlat(neighborhood.geometry.coordinates)
+        );
+        return surface?.data ? surface.data[idx] : 0;
```
aaronxsu: Seems like if `surface?.data` is false-y, then we could return 0 early and skip the calculation for index. This might have performance implications. I'd suggest we improve it. For example:

```diff
diff --git a/src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodTravelTimes.ts b/src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodTravelTimes.ts
index 1374ba3..ea6db02 100644
--- a/src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodTravelTimes.ts
+++ b/src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodTravelTimes.ts
@@ -13,10 +13,11 @@ export default function createNeighborhoodTravelTimes(
 ): number[] {
     return neighborhoods.features.map(neighborhood => {
         const surface = network.travelTimeSurface;
+        if (!surface?.data) return 0;
         const idx = coordinateToIndex(
             network,
             lonlat(neighborhood.geometry.coordinates)
         );
-        return surface?.data ? surface.data[idx] : 0;
+        return surface.data[idx];
     });
 }
```

src/frontend/src/reducers/networks/networksThunk.ts
```diff
@@ -54,65 +57,105 @@ export const getNetworks = createAsyncThunk(
...
-                (dataByNetwork, data) => {
-                    dataByNetwork[data.name] = data;
-                    return dataByNetwork;
-                },
-                {} as { [key in NetworkModeOptionKey]: TimesAndPathsData }
-            );
-        } catch (error: any) {
-            return rejectWithValue(error.message);
+                        const travelTimeSurface = parseTimesData(
+                            timesResponse.value
```
aaronxsu, line 111: Looks like there was a check for `pathsResponse` before parsing above- should we do something similar here?

src/frontend/src/reducers/networks/networksThunk.ts
```diff
@@ -54,65 +57,105 @@ export const getNetworks = createAsyncThunk(
...
+                        );
+
+                        const routableNetwork = {
+                            ...networkDetails,
+                            name: network as NetworkModeOptionKey,
+                            ...pathsData,
+                            travelTimeSurface: travelTimeSurface,
```
aaronxsu: This could be simply `travelTimeSurface,`

src/frontend/src/reducers/networks/networksThunk.ts
```diff
@@ -54,65 +57,105 @@ export const getNetworks = createAsyncThunk(
...
+                            name: network as NetworkModeOptionKey,
+                            routesByNeighborhood: networkRoutesByNeighborhood,
+                            travelTimesByNeighborhood:
+                                networkTravelTimesByNeighborhood,
+                            timesAndRoutesDataReady: true,
+                        };
+                    })
+                );
+                return {
+                    label: label,
```
aaronxsu, line 147: After we have the user sign-in and profile hooked up with the backend API, I'd suggest we use the destination record's ID instead of the label (currently name or address here), because IDs are usually unique identifiers. If we are going with this solution, I'd suggest we add a TODO here.

aaronxsu (approved):
Works as advertised. I left some minor suggestions for potential improvements.

src/frontend/src/reducers/networks/networksThunk.ts
rachelekm: That check is to handle the empty Promise in the event the network mode is car (since there is no paths data for this mode). But we do expect a times data response for every network mode or else it should land in the try-catch to error.

src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodRoutes.ts
rachelekm: Updated in eb029d844da8a7f8ca827631df8efaa77f7d46db

src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodRoutes.ts
rachelekm: Updated in eb029d844da8a7f8ca827631df8efaa77f7d46db

src/frontend/src/reducers/networks/utils/createNetworkNeighborhoodTravelTimes.ts
rachelekm: Updated in eb029d844da8a7f8ca827631df8efaa77f7d46db

src/frontend/src/reducers/networks/networksThunk.ts
rachelekm: Updated in eb029d844da8a7f8ca827631df8efaa77f7d46db

src/frontend/src/reducers/networks/networksThunk.ts
rachelekm: Added this comment in c670861f359d18f80e79031c982e2f2391baedc4

rachelekm:
Thanks for the review and these suggestions, @aaronxsu! I implemented all of the above and rebased with develop. Will go ahead and merge on successful CI.

PR #709, opened 2025-09-10
Add region field to Neighborhood

### LINKED ISSUE #706: Add "Region" field to Neighborhood & Update CSV template
As part of the new designs users will be able to filter their recommendations list by "region". Region is a choice of four options: "Metro West", "North Shore", "South Coast", "Greater Boston".

This is data that BHA maintains and will be expected to be included in the new neighborhoods.csv for the [#643](https://github.com/azavea/echo-locator/issues/643)

Issue is to create a new nullable Choices CharField on Neighborhood model, update the neighborhoods.csv template to include this new field as a column, and refactor the neighborhoods update script (neighborhood_date/update_data.sh) to populate this field from column values, as needed.
In addition to model change, issue artifact should be an updated neighborhoods.csv template to send to BHA.

rachelekm:
## Overview

Adds new nullable "region" field to Neighborhood model and updates neighborhood data import scripts as needed.
Not captured in PR code changes, but I also updated the `neighborhoods.json` and `neighborhood_bounds.json` files in the `echolocator-staging-data-us-east-1` bucket on S3 with updated files that now include a region field and test data.

### Checklist

- [x] CI passes after rebase

### Demo

On running import using staging data:
```
Done!

0 neighborhoods created, 236 updated

```
New populated region field in serialized response following import:
Screenshot 2025-09-10 at 3 21 44 PM

### Notes

This field will be used for frontend filtering by region. There is an existing `town_area` field on `Neighborhood`, but this is already used for grouping zipcodes as part of the ranking algorithm and should not be repurposed.

Ideally, this field is non-nullable with a set of pre-defined region choices. However we do not have a list of those choices and some neighborhoods are not within the city of Boston but a suburb so don't belong to a "region". Additionally, these are informal designations and could likely change with more client or user feedback, so for now we'll leave this a nullable character field and can re-evaluate to a choices char field down the road. 

## Testing Instructions

 *  `scripts/manage migrate`
 * `cd neighborhood_data && ./update_data.sh`
 * Confirm no errors running update data script and all neighborhood objects in new `neighborhoods.json` have `region` field
 * Confirm neighborhood object with zipcode "01432" has the region set as "Metro West"
 * Import test data and images from S3 staging using the below command:
```
AWS_PROFILE=echo-locator ./scripts/manage import_neighborhood_data s3://echolocator-staging-data-us-east-1/neighborhoods.json s3://echolocator-staging-data-us-east-1/neighborhood_bounds.json s3://echo-locator-staging-site-us-east-1/assets/neighborhoods/
```
* Confirm no errors on get and import
* `scripts/server`
* Open dev tools console and verify new field region in each neighborhood
* Find neighborhood with zipcode "01432" (should be at index 0) and confirm region set as "Metro West"

Resolves #706


src/backend/api/models.py
```diff
@@ -97,6 +97,11 @@ class Neighborhood(models.Model):
         default=0,
         help_text="Total count of voucher holder families that have moved to neighborhood.",
     )
+    region = models.CharField(
+        max_length=100,
+        blank=True,
```
aaronxsu, line 102: After applying the migration, I see the following in the DB:

```
 Table "public.api_neighborhood"
               Column               |            Type             | Collation | Nullable |             Default              
------------------------------------+-----------------------------+-----------+----------+----------------------------------
 region                             | character varying(100)      |           | not null | 
```

From the PR comment, it seems like the `region` field is meant to be nullable. Is this intentional? Should we also add `null=True` on the Django model side, so the DB will also allow null?

aaronxsu (commented):
I posted a question in the comment and would like to hear what you think!

src/backend/api/models.py
rachelekm: Whoops, thank you yes it should be nullable! Just fixed in 4cf408de38d5b00c5bc64cbf36e414a37c5a78e9

aaronxsu (approved):
Looks good and works well. I'd also recommend updating the frontend type/interface for neighborhood as well. This is good to go.

rachelekm:
Thanks, @aaronxsu! Added that interface update in 18dd46eec8f461d7da67f843bdd97319901bd412 👍 
Rebased with develop and will merge on successful CI.

PR #718, opened 2025-09-23
Refactor user authentication APIs

### LINKED ISSUE #641: Refactor email authentication
~~The specific refactor is based on decision artifact from #680~~

~~The app was previously publicly accessible and included user authentication via password-less login. As part of the frontend rewrite we want to continue with password-less login but now the entire site is private to admins and echo program participants.~~

~~This issue is to refactor the login endpoint so that the `send_login_link` method dispatches with a valid token for authenticated users. Issue also includes cleaning up now-unnecessary user registration endpoint and any other historical authentication logic.~~

Copy of comment below: Following 07-22 client meeting we are now making this a public site and enforcing "security through obscurity" vs explicitly verifying users are ECHO program participants.

On the frontend there is no longer an option to create an account, it will be done automatically on the backend if an existing account does not exist. This PR is now to alter backend so that user registration and login are the same endpoints and emails with token link are triggered accordingly and still function.

aaronxsu:
## Overview

This PR refactors the backend user authentication API endpoints:
- POST to `api/login/` to log in or register as a new user, send an email with Auth token
- POST `api/logout/` to log out and delete API token
- GET to `api/auth/login/?echo_auth=` to obtain an API token by an Auth token (btw, I am open to changing the URL to something like `api/auth/?echo_auth=` or `api/token/?echo_auth=`; right now, I kept what's originally there though)

### Checklist

- [ ] CI passes after rebase

### Notes

1. The previous implementation is cookie-based. This PR changes to token-based:
    - Frontend checks local storage for an **API token**
    - If it does not have an API token, redirect to login page, go through the login workflow, user will receive an email. What happens after user clicks the link in the email is described in item 2 below
    - If it has an API token, read that token into Redux store
    - All subsequent API calls will use the API token from Redux store and pass it as header
    - On log out, frontend will remove the API token from local storage; backend will delete the API token
2. To build in the next card: The URL in the authentication email has the (one-time) **Auth token** embedded, and the path is actually a page to be supported in the frontend router. This will be a frontend page with path `/login/callback?echo_auth=<auth_token>`, which has no real component on the page, but only for redirect:
    - It calls the `/api/auth/login/?echo_auth=<auth_token>` endpoint, which return a JSON response `{"token": <API Token>}`
    - Store the API token in the Redux store and Local Storage. Redirect to the discover page
    - If the API endpoint call's status code is not 200, redirect to the login page

## Testing Instructions

* Replace with your email and run the following to log in. Look for the URL in the email logged to the API server console logs
```bash
curl --location 'http://localhost:9966/api/login/' \
--header 'Content-Type: application/json' \
--data-raw '{"username": "<replace with your email>"}'
```
* The URL in the email will be something like `http://localhost:9966/login/callback?echo_auth=<Auth Token>`. Grab the **Auth Token**. This is a one-time token
* Run the following to get an API token
```
curl --location 'http://localhost:9966/api/auth/login/?echo_auth=<Auth Token goes here>'
```
* The above response should be like `{"token": <API Token>}`. The **API Token** is long-living and is what we need for API calls. In the upcoming frontend development, this will be stored in Local Storage and in the Redux store
* Use the API token to list neighborhoods:
```
curl --location 'http://localhost:9966/api/neighborhoods' \
--header 'Authorization: Token <API Token goes here>'
```
* Log out:
```
curl -X POST --location 'http://localhost:9966/api/logout/' \
--header 'Content-Type: application/json' \
--header 'Authorization: Token <API Token goes here>'
```

Resolves https://github.com/azavea/echo-locator/issues/641

### DIFF
--- a/src/backend/api/views.py
+++ b/src/backend/api/views.py
@@ -1,3 +1,4 @@
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
@@ -6,6 +7,8 @@ from .serializers import NeighborhoodBoundsSerializer, NeighborhoodSerializer
 
 
 class ListNeighborhoods(APIView):
+    permission_classes = [IsAuthenticated]
+
     def get(self, request):
         neighborhoods = Neighborhood.objects.all()
         serializer = NeighborhoodSerializer(neighborhoods, many=True)
@@ -13,6 +16,8 @@ class ListNeighborhoods(APIView):
 
 
 class ListNeighborhoodBounds(APIView):
+    permission_classes = [IsAuthenticated]
+
     def get(self, request):
         neighborhoods = Neighborhood.objects.all()
         serializer = NeighborhoodBoundsSerializer(neighborhoods, many=True)
--- a/src/backend/users/views.py
+++ b/src/backend/users/views.py
@@ -1,13 +1,9 @@
 from django.conf import settings
-from django.contrib.auth.models import User
+from django.contrib.auth.models import Group, User
 from django.contrib.gis.geos import Point
 from django.core.mail import send_mail
 from django.db import transaction
-from django.db.utils import IntegrityError
-from django.http import HttpResponseRedirect
-from django.urls import reverse
 from rest_framework.authtoken.models import Token
-from rest_framework.exceptions import ValidationError
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
@@ -27,7 +23,7 @@ def send_login_link(request):
     login_token = utils.get_query_string(user)
     host = request.get_host()
     protocol = "https://" if request.is_secure() else "http://"
-    login_link = protocol + host + reverse("obtain_token") + login_token
+    login_link = f"{protocol}{host}/login/callback{login_token}"
 
     html_message = """
     <p>Hi there,</p>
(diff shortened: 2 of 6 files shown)


src/backend/api/views.py
```diff
@@ -1,18 +1,23 @@
 from rest_framework.response import Response
 from rest_framework.views import APIView
+from rest_framework.permissions import IsAuthenticated
 
 from .models import Neighborhood
 from .serializers import NeighborhoodBoundsSerializer, NeighborhoodSerializer
 
 
 class ListNeighborhoods(APIView):
+    permission_classes = [IsAuthenticated]
```
aaronxsu, line 10: Noticed that these endpoints were actually not behind auth. Now they are!

aaronxsu:
Based on the following in the DRF docs, tokens don'r expire unless it's removed from DB, which works for this project: https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication

>For an implementation which allows more than one token per user, has some tighter security implementation details, and supports token expiry, please see the [Django REST Knox](https://github.com/James1345/django-rest-knox) third party package.

This suggests the out-of-the-box token does not expire.

src/backend/users/views.py
```diff
@@ -51,41 +47,64 @@ def send_login_link(request):
...
+
+        if created:
+            # If a new user was created, perform any first-time setup
+            # This logic is moved from the old serializer's .create() method
+            user.email = email
+            user.save()
+            UserProfile.objects.create(user=user)
+            # Add them to a group
+            house_seeker_group = Group.objects.get(name="HouseSeeker")
+            user.groups.add(house_seeker_group)
```
rachelekm, line 75: In theory we no longer need this permission group, but we're still using this in the django admin and we're not cleaning up the backend as part of this phase of work so we'll leave this in for now.

rachelekm (approved):
This is working great! As discussed this morning updating from cookie-based to token-based makes sense and the upcoming frontend cards are updated with this new convention 👍

src/backend/users/views.py
aaronxsu: Ah got it. Sounds good to me.

aaronxsu:
Thanks for reviewing! I updated the change log and squashed the fixup commits. Merging after CI.

PR #723, opened 2025-10-01
Add frontend sign in page and authentication workflows

### LINKED ISSUE #645: Add sign-in page
[Implement sign-in page following figma design for mobile and desktop](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=925-6100&m=dev). Login endpoints, token generation, email and AWS SES already setup. 
In addition to UI:
- Create basic react router setup to place all pages (except sign-in) behind protected routes for authenticated users

Image

### LINKED ISSUE #700: Add frontend authentication state setup
Our application is currently setup with Django passwordless authentication via django-sesame and Token.objects. On login the backend authenticates the user and sends the user a login link via email. On clicking the link, `ObtainToken` endpoint view generates a secure auth_token, sets as a cookie and redirects to frontend. Old frontend used this cookie to set `authState` in combination with `js-cookie` for user's authenticated state. [See PR implementing this workflow for reference.](https://github.com/azavea/echo-locator/pull/482)

We should implement a similar frontend workflow:
- Create `authSlice` RTK state to manage auth
- Configure axios to automatically send API token with every request
- Fetch user profile and set state (already exists) on successful login
- Create a logout thunk that called the logout endpoint and clear redux state
- Ensure on logout redirected to signin page

aaronxsu:
## Overview

This PR:
- adds redux state management for authentication workflows
- adds a sign in page
- adds functional pages for auth callback and logout
- protects existing app pages by auth
- adds axios interceptors for auth purposes
- adds translations to the copies on the sign in page

Tagging @philippschmitt for some design and styling cares.

Here is an overview of the auth workflow:
1. User goes to the frontend app's root URL. If not authenticated, they are redirected to the log in page
2. User types in their email to sign in under `/login` page
3. They click on the URL in the email, which goes to the functional `callback?echo_auth=` page. App grabs the auth token from URL and requests an API token from the backend. 
4. App then stores this API token in redux and persists it to browser local storage
5. App then redirects to the discover page. All subsequent API calls will use this API token.
6. Next time user visits the app, app grabs the API token from local storage and stores it in redux. Then it proceeds to the usual biz logic
7. On logging out, app goes to the functional `/logout` page, which calls the log out endpoint to remove the API token from DB and clears it from both redux and local storage
8. In the future, when user visits the app, since there is no auth token in the local storage, user will be redirected to the login page to go from step 1 again

### Checklist

- [ ] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/d4404cad-5deb-45a4-a012-57f1dc3906f4

## Testing Instructions

 * Remove `VITE_AUTHTOKEN` from `.env` file locally, since we won't need it anymore. (Also, we should remove it from the dotenv file in staging S3 bucket- I did not do it yet)
 * Go to http://localhost:9966/. It used to redirect to the discover page, but now should take you to the sign in page. Also try to visit the discover page directly. It should take you to the sign in page as well. Since at this point, there is no API token in the redux store (which reads the token from local storage)
 * Sign in by your email. The login URL should be in the server logs, since this is development
 * Grab the URL from email in server logs. It should look like `/callback?echo_auth=`. Open your network tab in dev tools. Now press enter to go to this page. Observe that it calls the `/auth/login` API endpoint. When succeed, it should store the API token under `auth.token` in redux store and persists it to local storage under key `apiToken`. It should redirect you to the discover page
 * Observe the API requests to list neighborhoods and neighborhood bounds. They should use this new API token
 * Now close the tab. In another tab, go to http://localhost:9966/ again. This time it should not ask you to sign in but should take you to the discover page.
 * Click on the logout button in the menu bar. It should take you to the `/logout` page. Under the hood, observe: it sends out a `/logout` API request, it also clears the API token from redux and from local storage. It should then take you to the sign in page.
 * Visit http://localhost:9966/ again in a new tab, it should also take you to the sign in page, because there is no token in local storage (since it's cleared in the last step), and thus no token in redux store

Closes https://github.com/azavea/echo-locator/issues/645
Closes https://github.com/azavea/echo-locator/issues/700


src/frontend/public/locales/es/translation.json
```diff
@@ -1,5 +1,13 @@
 {
     "discover": "Descubrir",
     "compare": "Comparar",
-    "logout": "Cerrar sesión"
+    "logout": "Cerrar sesión",
```
aaronxsu, line 4: These are from google translate

src/frontend/public/locales/zh/translation.json
```diff
@@ -1,5 +1,13 @@
 {
     "discover": "发现",
     "compare": "对比",
-    "logout": "退出"
+    "logout": "退出",
```
aaronxsu, line 4: These are from my translations

src/frontend/src/pages/Auth/Callback.tsx
```diff
@@ -0,0 +1,51 @@
...
+            try {
+                // Call the backend API to exchange auth token with API token
+                const response = await getApiToken(authToken);
+                const { token: apiToken } = response;
+                // Add the API token to localStorage for persistence
+                localStorage.setItem(LOCAL_STORAGE_TOKEN_KEY, apiToken);
+                // Add the API token to the Redux state
+                dispatch(setToken(apiToken));
+                // Go to the /discover page
+                navigate(`/${language}/discover`, { replace: true });
```
aaronxsu, line 37: Right now, it goes to the discover page. I think user needs to set up a profile so that profile-related stuffs can show up in the discover page. I suggest we handle that on the discover page, because discover page is where user profile settings are needed, plus it covers both cases when user is logged in or not
- Grab user profiles from API
- If no profile at all or if the preferences are not complete, we take them to the user profiles page to go through the wizard
- Otherwise, biz as usual.

What do you think?

rachelekm: Yeah that sounds like the right plan 👍 as part of that sign-in flow we'll need to fetch all the other data too so in detail that looks like:
- On successful sign-in: fetch neighborhoods, neighborhood bounds, networks data and user profile

-  If a user profile exists, fetch paths and times data for all destinations and set active (primary) destination in state
       -   Once paths and times data is ready, dispatch calculate ranking logic
-  If a user profile does not exist (or destinations list is empty), take them to the user profiles page
        -   On save, fetch paths and times data and calculate ranking

src/frontend/src/pages/Auth/Login.tsx
```diff
@@ -0,0 +1,146 @@
...
+                onOpenChange={setIsOpen}
+            >
+                <Modal>
+                    <Dialog aria-label="Sign in with email">
+                        <div className={modalContainer()}>
+                            {!isEmailSent ? (
+                                <>
+                                    <p className={modalTitle()}>
+                                        {t("signInWithEmail")}
+                                    </p>
```
rachelekm, line 102: There's a [subtitle in the figma designs](https://www.figma.com/design/LAy4vzpOLf7YnXbB7pGfN8/BHA-ECHO?node-id=754-16791&m=dev) to include here with an inline link to[ bha echo site](https://www.bostonhousing.org/en/Section-8-Leased-Housing/Voucher-Programs/Expanding-Choice-in-Housing-Opportunities-(ECHO).aspx)

rachelekm (approved):
Looks really good and works as expected! I responded to comments directly but your decisions for sign-in behavior match the expected state flow 👍 

I left one comment for some missing text to address but once that's in this is good to go!

src/frontend/src/pages/Auth/Login.tsx
aaronxsu: Good call. I thought we would drop this subtitle due to previous auth related discussions. Added it in https://github.com/azavea/echo-locator/pull/723/commits/1cb56972498f3b54df9692a493c7c9fb4f669c1d

Screenshot 2025-10-06 at 2 14 58 PM
Screenshot 2025-10-06 at 2 15 20 PM
Screenshot 2025-10-06 at 2 26 51 PM

aaronxsu:
Rebased on developed, addressed the comment, added a changelog entry, and tested on staging via a `test` branch. Merging now.

Issue #733, opened 2025-10-09
Add edit user profile buttons


rachelekm:
Add edit user profile button to the filters modal. Button content should use data from state. On edit, open user profile wizard (this will not contain the edit trips steps) and update user profile on finish:

<img width="349" height="210" alt="Image" src="https://github.com/user-attachments/assets/32310c40-53dc-438e-b73a-4880f987bdf3" />

Update existing edit buttons on discover desktop view to show profile/trip data from state:

user profile: bedroom count and active travel mode as label
destination: active trip data from state (activeMode as icon, activeDestination purpose, and activeDestination address)
<img width="409" height="220" alt="Image" src="https://github.com/user-attachments/assets/3b9a0aab-ae64-4fc8-b568-0738b9a14f98" />

Issue #728, opened 2025-10-06
Add trips edit modal


rachelekm:
Create trip destination edit modal, to open whenever "edit trips" button in app is pressed ("your trips" component and discover header).

On trip destinations modal:
- "choose trip" sets the `userProfile.activeDestination` value
- selecting "edits trips" opens first (only) step in destinations wizard
- Selecting traffic conditions updates `networks.activeMode`
- Selecting "how get around" updates `userProfile.hasVehicle`,  `userProfile.useCommuterRail`, and `networks.activeMode`

If changes to "how get around" , update user profile on close. The trips wizard will update user profile on destinations change and traffic conditions are not saved to user profile.

<img width="311" height="635" alt="Image" src="https://github.com/user-attachments/assets/797d3519-25d8-412a-99f1-c01050815964" />

Issue #643, opened 2025-06-17
Update Neighborhood data


rachelekm:
Update neighborhood data on staging once new neighborhood CSV data available. We will run similar steps as part of production release 3.0.0 in #696.

Note we should also update remote dev data at `echo-locator-devdjango-config-us-east-1`.

rachelekm:
Steps to update new data on staging (steps for production/any extra notes to be captured in #696):

1. In local dev environment, clear out `neighborhood_data/images`, `neighborhood_data/zctas`, and all past generated `.csv` and .`json` files in root of `neighborhood_data` 
2. Place new `neighborhoods.csv` data file in local `neighborhood_data` dir
3. Run `cd neighborhood_data && ./update_data.sh`
4. Once successful, upload newly generated `neighborhoods.json` and `neighborhood_bounds.json` to `echolocator-staging-data-us-east-1` bucket
5. Run `./scripts/imagepublish staging` to add the new, scaled-up image assets to the assets/neighborhoods dir of the `echo-locator-staging-site-us-east-1` bucket 
6. Follow [instructions in deployment README](https://github.com/azavea/echo-locator/tree/develop/deployment#migrations) to start an interactive session in the django staging task
7. Run migrations and confirm schema is up to date
8. Run the below management script to update neighborhood data on staging:
```
python manage.py import_neighborhood_data s3://echolocator-staging-data-us-east-1/neighborhoods.json s3://echolocator-staging-data-us-east-1/neighborhood_bounds.json s3://echo-locator-staging-site-us-east-1/assets/neighborhoods/
```
9. Confirm new data and images on staging site without error
10. Exit task session

Following data update:
1. We should remove the `src/backend/api/tmp_data` dir following update as that csv and script will no longer be needed.
2. We chould confirm all distinct neighborhood regions and update NEIGHBORHOOD_REGIONS in frontend as needed (see comment https://github.com/azavea/echo-locator/pull/738/commits/3460532857133490b34154d939ed73543c75cb47#r2430001243)

PR #738, opened 2025-10-13
Add neighborhood list display filters

### LINKED ISSUE #727: Add regional and ecc filters
- Add base user profile edit modal (exclude edit user profile button)

Implement filtering logic by region type as well as exclusively ECC neighborhood:
- Add new filters state to `neighborhoods` redux slice
         - Region filter is list of unique region types. Default includes all region types. Reference list of regions, [here](https://docs.google.com/spreadsheets/d/1VqVj0sCNODwgFlD4n8RbxbJzsN7ksaV7A7anCeaKu7U/edit?gid=1355722612#gid=1355722612).
         - eccOnly filter is boolean value and false by default
- Add filtering logic to the `neighborhoodsSortedWithRoutes` selector, specifically following the "boost to downtown" logic but before grouping  for display. Filter by:
      - If a neighborhood region field is included in filter state "region" list
      - If eccOnly filter is true, filter out any neighborhoods where ecc field is false

Implement UI for filter selections following figma desings (region image incoming):

Image

rachelekm:
## Overview

Implements the ability to filter the recommended neighborhoods display by regions and exclusive to ECC. On selecting a filter, the recommended list will be filtered and the map styling and bounds adjust to new filtered list. Adds the filters modal in line with figma designs as well as the following cleanup:
- Adds translations to modals and edit buttons
- Updates edit buttons in desktop/mobile subheaders to use dynamic profile content
- Prevents showing invalid transit routes in neighborhood details (there are no segments for "too far" routes, it should default to outside directions like car routes)
- Adds temporary backend script to update neighborhoods with regions (CSV provided by ECHO to use until we receive the final neighborhoods update data)

### Checklist

- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/d570f3f4-ff81-45f1-8edc-bc1989de5b4f

## Testing Instructions

We were given a list of regions to neighborhoods to reference until the final updated neighborhoods data is received. For best testing, update your local db with this regions list using the below commands:
```
$ scripts/mange shell

>>> from api.tmp_data import update_neighborhoods_with_regions
>>> update_neighborhoods_with_regions.update()
```

Following the data update we can remove this new `tmp_data` dir, there is an additional step in #643 to capture this.

 * Run the above python script to load region values for neighborhoods locally (or update a handful of neighborhood regions manually to start)
 * `scripts/server`
 * In desktop view, confirm the new subheader edit buttons are now using dynamic values in place of static info and match designs
 * Click each button and confirm respective modal opens (the trips modal will be empty until #728)
 * Click a neighborhood in the list and scroll down to "Your Trips". Click button to edit trips and confirm trips modal opens.
 * Go back to the discover homepage and switch the mobile view
 * Click each edit button in the subheader and confirm that the respective modal opens
 * In the filters/profile modal, confirm content matches figma designs (note the edit trips button does not open anything yet, but will open trips wizard)
 * Confirm dynamic profile/trip information is accurate and translatable
 * All regions should be selected by default for filtering
 * The ECC filter should not be selected by default
 * Close the modal without any selection changes and confirm no list/map updates occur
 * Open the filter edit modal again and make various region selections. On closing modal confirm: 1) recommendations list updates and accurately filters neighborhoods by region 2) map styling and bounds update to filtered recommendations
 * Re-open modal and select all regions again
 * Close modal and confirm recommendations list and bounds reset to beginging
 * Open modal and select to filer to ECC-only neighborhoods
 * Confirm recommendations list and map styling/bounds updates and accurately filters neighborhoods to ECC only
 * Find a neighborhood that is "too far" in the list
 * Open its neighborhood details and confirm the "Your Trips" section no longer displays the route map
 * Click on a destination and confirm the google map transit directions automatically open in a new tab

Resolves #727 #733


.vscode/settings.json
```diff
@@ -1,20 +1,21 @@
...
+    "editor.formatOnSave": true,
 
-  // 3. Enable ESLint to auto-fix issues on save
-  "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "explicit"
-  },
+    // 3. Enable ESLint to auto-fix issues on save
+    "editor.codeActionsOnSave": {
+        "source.fixAll.eslint": "explicit",
+        "source.organizeImports": "explicit"
```
rachelekm, line 14: This seemed to capture formatting but the only line change is the addition of ` "source.organizeImports": "explicit"`

src/backend/api/tmp_data/old-data-town-to-regions.csv
```diff
@@ -0,0 +1,237 @@
...
+Wrentham,South Shore
+Scituate,South Shore
+Norwell,South Shore
+Norfolk,South Shore
+Millis,South Shore
+Marshfield,South Shore
+Mansfield,South Shore
+Hull,South Shore
+Hingham,South Shore
+Cohasset,South Shore
```
rachelekm, line 237: This CSV is from BHA and used to generate the map image in the modal, but I also found this useful to accurately test the regions filter is working. Assuming the updated neighborhood data updates is still not received in the next week or so it would be useful to have this on staging for demo purposes.

I didn't want to spend our limited time importing data from this csv (especially when it would be overwritten soon) so I had gemini create a quick script to run in the django shell. This felt the easiest to run in staging and later remove once we have the actual neighborhood data. Let me know what you think!

rachelekm: I also added a note in the data updates issue to remove this whole `tmp_data` dir following successful update.

src/frontend/src/reducers/modalsDisplay/modalsDisplaySlice.ts
```diff
@@ -0,0 +1,54 @@
...
+    setIsEditTripsOpen,
+    setIsEditFiltersOpen,
+} = modalsDisplaySlice.actions;
+
+export const selectIsNeighborhoodDetailsOpen = (state: RootState) =>
+    state.modalsDisplay.isNeighborhoodDetailsOpen;
+export const selectIsEditFiltersOpen = (state: RootState) =>
+    state.modalsDisplay.isEditFiltersOpen;
+export const selectIsEditTripsOpen = (state: RootState) =>
+    state.modalsDisplay.isEditTripsOpen;
```
rachelekm, line 52: I ended up pulling out modal display states into its own reducer since we need to update this state in neighborhood details and the compare page.

src/frontend/src/components/YourTrips/CommuteGroup.tsx
```diff
@@ -36,6 +37,11 @@ const CommuteGroup = ({
 
     useEffect(() => {
         setSelectedMapDestination(initialTrip.destination);
+        const minCommuteTime =
+            allNeighborhoodCommutes[initialTrip.neighborhoodZipcode]?.commutes[
+                initialTrip.destination.location.label
+            ].commuteMin;
+        setSelectedDestIsTooFar(minCommuteTime > 120);
     }, [initialTrip.destination]);
```
rachelekm, line 48: I noticed this while testing, but when the trip is "too far" it doesn't create any segments to map (because 120 minutes is the analysis max). We should treat the invalid trip routes like "car" and hide the map and direct to the google map directions on click. These changes implement that fix.
Screenshot 2025-10-13 at 1 04 04 PM

aaronxsu:
It looks like CI failed. Could you take a look and see what's going on?

rachelekm:
> It looks like CI failed. Could you take a look and see what's going on?

Thanks, @aaronxsu! It failed on a couple linting errors. Just addressed those and rebased along the way to fix the immediate merge conflicts. This is ready to take a look!

src/backend/api/tmp_data/old-data-town-to-regions.csv
aaronxsu: So, it sounds like, in the neighborhood data to be provided by BHA, it will include values in the `region` field, so running the script (added in this PR) consuming this temporary data will not be necessary at that point. The temp data and script are here for staging and demo purposes, and will be removed before the deployment.

Does my understanding sound right?

src/frontend/src/constants.ts
```diff
@@ -95,3 +95,10 @@ export const PROFILE_MAX_IMPORTANCE = MAX_IMPORTANCE - 1;
  */
 
 export const LOCAL_STORAGE_TOKEN_KEY = "apiToken";
+
+export const NEIGHBORHOOD_REGIONS = [
```
aaronxsu, line 106: I'd add a TODO here in case these region names change in the updated CSV they will provide us. And probably also link to this block in the data update issue in case we forget.

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
```diff
@@ -36,10 +38,24 @@ export const getRankedNeighborhoodLists =
...
-        const neighborhoodsList = neighborhoodsSortedWithRoutes(state);
+        // Get ranked list
+        let neighborhoodsList = neighborhoodsSortedWithRoutes(state);
         const neighborhoodNameByZipcode =
             selectNeighborhoodNameByZipcode(state);
 
+        // Apply filters
+        if (
+            state.neighborhoods.neighborhoods &&
+            Object.keys(state.neighborhoods.filters)
```
aaronxsu: I don't think I followed why we are checking `Object.keys(state.neighborhoods.filters)` here. Is it nullable? Did you see anything weird in development if not checking this?

src/frontend/src/reducers/neighborhoods/utils/filterNeighborhoodsList.ts
```diff
@@ -0,0 +1,65 @@
...
+// Filtered neighborhoods then grouped for list display
+const filterNeighborhoodList = (
+    neighborhoodsList: {
+        recommended: string[];
+        tooFar: string[];
+    },
+    neighborhoods: Neighborhoods,
+    filters: FiltersState
+) => {
+    if (!Object.keys(filters).length) {
```
aaronxsu: Doesn't `Object.keys(filters).length` always evaluate to true? Am I missing something here?

Maybe you are looking for something like this?
```
const { ecc, regions, textSearch } = filters;
const searchTerm = textSearch?.trim().toLowerCase();
if (!ecc && !regions?.length && !searchTerm) return neighborhoodsList;
```

src/frontend/src/reducers/neighborhoods/utils/filterNeighborhoodsList.ts
```diff
@@ -0,0 +1,65 @@
...
+        return neighborhoodsList;
+    }
+
+    const neighborhoodPropsByZipcode = neighborhoods.features.reduce(
```
aaronxsu: Seems like this mapping doesn't really change no matter the filters- sounds right? I'd move this out if this is the case.

src/frontend/src/reducers/neighborhoods/utils/filterNeighborhoodsList.ts
```diff
@@ -0,0 +1,65 @@
...
+        zipList.filter(zip => neighborhoodPropsByZipcode[zip].ecc);
+    const filterByRegions = (zipList: string[], regions: string[]) =>
+        zipList.filter(
+            zip =>
+                neighborhoodPropsByZipcode[zip].region &&
+                regions.includes(neighborhoodPropsByZipcode[zip].region)
+        );
+
+    if (filters.ecc) {
+        filteredRecommended = filterEccOnly(filteredRecommended);
```
aaronxsu: We might have an opportunity to combine the three filters - ECC, regions, and text search in one filter and go through the list in one pass for recommended and in one pass for top 10. Did you get a chance to explore that?

src/frontend/src/pages/Discover/ModalsWrapper.tsx
```diff
@@ -0,0 +1,13 @@
+import EditFiltersModal from "src/components/EditFiltersModal";
+import NeighborhoodDetail from "../NeighborhoodDetail";
+import EditTripsModal from "src/components/EditTripsModal";
+
+const ModalsWrapper = ({ isMobile }: { isMobile?: boolean }) => (
```
aaronxsu, line 5: I'd give a default for `isMobile` so if the prop is not present, it's `false`

src/frontend/src/components/EditTripsModal.tsx
```diff
@@ -0,0 +1,36 @@
+import { ModalOverlay, Modal } from "components/base/Modal/Modal";
```
aaronxsu, line 1: I'd reorder the import orders in this module.

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+    Dialog,
+    Heading,
+    Link,
+} from "react-aria-components";
+import { useTranslation } from "react-i18next";
+
+import {
+    selectIsEditFiltersOpen,
+    setIsEditFiltersOpen,
+} from "src/reducers/modalsDisplay/modalsDisplaySlice";
```
aaronxsu: I'd remove the `src/` bit from the import. This might be the case in other places too.

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+    const mode = useAppSelector(selectActiveMode);
+    const [filtersBuffer, setFiltersBuffer] = useState<FiltersState | null>(
+        null
+    );
+
+    const filters = useAppSelector(selectNeighborhoodFilters);
+
+    useEffect(() => {
+        setFiltersBuffer({ ...filters });
+    }, []);
```
aaronxsu: Should we listen to `filters` in the store?

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+    if (!filtersBuffer) {
+        return <></>;
+    }
+
+    const regionsFilterSelected = (isSelected: boolean, region: string) => {
+        const updatedRegionsFilter = filtersBuffer.regions
```
aaronxsu: At this point, can `filtersBuffer.regions` be falsy?

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+    };
+
+    const onOpenChange = (isOpen: boolean) => {
+        dispatch(setNeighborhoodFilters(filtersBuffer));
+        dispatch(setIsEditFiltersOpen(isOpen));
+        // Recalculate ranking on filters close
+        if (
+            isOpen == false &&
+            (filters.ecc !== filtersBuffer.ecc ||
+                filters.regions.length !== filtersBuffer.regions.length)
```
aaronxsu: Length might not be the most reliable criteria here. E.g. what if previously the filter was `["North Shore", "Metro West"]`, but now it is `["Boston and Greater Boston", "South Shore"]`. I'd consider using a loadash method to compare the array or converting the arrays to `Set` and use Set properties to compare them.

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+                                    setFiltersBuffer({
+                                        ...filtersBuffer,
+                                        ecc: isSelected,
+                                    })
+                                }
+                                description={`Select to only recommend neighborhoods that are Expanded Choice Communities (ECC)`}
+                                footer={
+                                    <Link
+                                        href={
+                                            import.meta.env
```
aaronxsu: I'd have it in a constant similar to https://github.com/azavea/echo-locator/blob/develop/src/frontend/src/pages/Auth/Login.tsx#L20

Or maybe we put it in the constants module, since now it's used in more than 1 place

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+                                    ? "/regions-map-mobile.png"
+                                    : "/regions-map-desktop.png"
+                            }
+                            customImageClassName="w-full"
+                            customCaption={
+                                <>
+                                    CC BY-SA 3.0. $
+                                    {t("filterModal.regionMapCaption")}{" "}
+                                    <Link
+                                        href="https://commons.wikimedia.org/wiki/File:Simple_Massachusetts_Vector.svg"
```
aaronxsu: If this URL won't change, it's ok to hardcode. Otherwise, I'd consider having an env var

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -0,0 +1,243 @@
...
+    };
+
+    return (
+        <ModalOverlay
+            isDismissable
+            isMobile
+            isOpen={modalOpen}
+            onOpenChange={onOpenChange}
+        >
+            <Modal size="medium" className="p-5 max-h-full overflow-y-scroll">
```
aaronxsu, line 110: It might be cleaner to create a style file for the class names in this module. Depending on your remaining budgeted time on this task, this is not a priority.

src/frontend/src/components/YourTrips/CommuteGroup.tsx
```diff
@@ -36,6 +37,11 @@ const CommuteGroup = ({
 
     useEffect(() => {
         setSelectedMapDestination(initialTrip.destination);
+        const minCommuteTime =
+            allNeighborhoodCommutes[initialTrip.neighborhoodZipcode]?.commutes[
+                initialTrip.destination.location.label
+            ].commuteMin;
+        setSelectedDestIsTooFar(minCommuteTime > 120);
```
aaronxsu: I think I've seen 120 somewhere else- we might reuse that constant. Otherwise, I'd suggest we create one.

src/frontend/src/components/YourTrips/CommuteGroup.tsx
aaronxsu: Good catch!

src/frontend/src/pages/Discover/Map/Map.tsx
```diff
@@ -37,18 +42,52 @@ const Map = ({ isMobile = true, mapDisplay = true }: Props) => {
...
+    const rankedNeighborhoodsLists = useAppSelector(
         selectRankedNeighborhoodsLists
     );
+    const { topTen, recommended } = rankedNeighborhoodsLists;
     const destinations = useAppSelector(selectUserDestinations);
     const activeDestination = useAppSelector(selectActiveDestination);
 
+    const filteredNeighborhoodBounds = useMemo(() => {
+        const filteredFeatures = neighborhoodBounds?.features.filter(f => {
+            const flatFilteredNeighborhoods = Object.values(
```
aaronxsu: I'd move this out of the filter function, as this won't change and we should not need to create it in each iteration.

aaronxsu (changes requested):
Overall, looking and working well. I left some suggestions in the comments, with one maybe related to a small bug when the region filter array has the same length and the filter won't seem to trigger.

In the following steps, the last step did not seem to work for me. Could you check it out? I might be missing something though.
>- Find a neighborhood that is "too far" in the list
>- Open its neighborhood details and confirm the "Your Trips" section no longer displays the route map
>- Click on a destination and confirm the google map transit directions automatically open in a new tab

src/backend/api/tmp_data/old-data-town-to-regions.csv
rachelekm: Yes, that's correct. The CSV template we gave them includes a column for the `region` field

rachelekm:
Thanks so much, @aaronxsu, great catches in this review! I initially started applying filters with the concept that filters state was empty and key-value pairs were added on filter selections. This grew to feel confusing from a development/type-safe-side and (and added more complexity for region filtering) so I eventually brought in the filters state as part of initializing the neighborhoods slice. However, I forgot a lot of the initial conditional checks for filters, causing issues since now those conditions would never be false-y 😅 I refactored the filtering logic (along with added some simplifications thanks to your suggestions) in [144c967](https://github.com/azavea/echo-locator/pull/738/commits/144c967876effa8dae51dc76f93b2adef69ee93b). This should take care of that buggy-ness!

I also addressed your other comments and will reference the commit of those fixes inline. This is ready for another review! Noting that I rebased on develop to tackle merge conflicts so you will need to force pull!


>In the following steps, the last step did not seem to work for me. Could you check it out? I might be missing something   though.
>
>Find a neighborhood that is "too far" in the list
>Open its neighborhood details and confirm the "Your Trips" section no longer displays the route map
>Click on a destination and confirm the google map transit directions automatically open in a new tab


☝️ This seems to still work for me, are you seeing straight-line routes? I have it set up where it only switches to "car mode" if the min commute time is greater than 120 minutes but maybe there are invalid routes if the max is over 120 minutes too?:

Screenshot 2025-10-15 at 5 38 30 PM

src/frontend/src/reducers/neighborhoods/neighborhoodsThunk.ts
rachelekm: Added an explanation in my summarizing comment but this is resolved in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/reducers/neighborhoods/utils/filterNeighborhoodsList.ts
rachelekm: Good catch this is resolved in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: Yup! This is resolved in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: Added an explanation in my summarizing comment but this is resolved in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: Great point 😅 this is resolved in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/pages/Discover/Map/Map.tsx
rachelekm: Good point, I ended up simplifying this quite a bit and was able to get rid of the filter function in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/constants.ts
rachelekm: Makes sense I added an inline comment in 3460532857133490b34154d939ed73543c75cb47 and updated the checklist in #643

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: This is updated in 501d91a80a9e1623cccc1c991a1362f7f03dfc10

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: This is updated in 501d91a80a9e1623cccc1c991a1362f7f03dfc10

src/frontend/src/components/YourTrips/CommuteGroup.tsx
rachelekm: Used that constant in 560adc85564537d73e4675b41eec2d625bd4e95e

src/frontend/src/components/EditTripsModal.tsx
rachelekm: Fixed in 530c52b9b6fd67fc4aef3e83ccee8c313889a192

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: Fixed in 530c52b9b6fd67fc4aef3e83ccee8c313889a192

src/frontend/src/pages/Discover/ModalsWrapper.tsx
rachelekm: Fixed in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/reducers/neighborhoods/utils/filterNeighborhoodsList.ts
rachelekm: Nice that's a great point, this was my quick-as-I-can implementation to get everything working. Combined these together into one filter and it's much more simple! That update is captured in 144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/reducers/neighborhoods/utils/filterNeighborhoodsList.ts
rachelekm: Added an explanation in my summarizing comment but this is resolved in https://github.com/azavea/echo-locator/commit/144c967876effa8dae51dc76f93b2adef69ee93b

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: Decided to skip this update given the remaining time

aaronxsu (approved):
Thanks for addressing the comments- the changes look great to me.

>☝️ This seems to still work for me, are you seeing straight-line routes? I have it set up where it only switches to "car mode" if the min commute time is greater than 120 minutes but maybe there are invalid routes if the max is over 120 minutes too?:

I was more thinking about the last step: "Click on a destination and confirm the google map transit directions automatically open in a new tab". I clicked on the destination in the accordion, but does the google map link open for you?

One other thing I discovered in the second round of testing, which could be an edge case or maybe I am missing something: I had the top 10 tour in progress, i.e. I see one of the top 10 neighborhood cards (not detail modal), and then I clicked on the filter button, applied some filters, and closed it => now I seem to remain on the old top 10 neighborhood without the filters (maybe?). I wonder if we should exist the top 10 tour if those buttons (on the sub-header) are clicked. What do you think?

rachelekm:
> 
> I was more thinking about the last step: "Click on a destination and confirm the google map transit directions automatically open in a new tab". I clicked on the destination in the accordion, but does the google map link open for you?

Oh I see now! This was working when I had multiple destinations because the google map link opens on accordion item "expansion". However if an item is already expanded (like in a destinations list of 1) the callback for "on expand change" will never be called. The react-aria Disclosure components (underneath Accordion) can't have click events set in their props, so I created a separate button styled like an Accordion item for users with only 1 destination set. This is now working in 74926817f7de1711c49d3a4126b45b9aac2112d9.


> 
> One other thing I discovered in the second round of testing, which could be an edge case or maybe I am missing something: I had the top 10 tour in progress, i.e. I see one of the top 10 neighborhood cards (not detail modal), and then I clicked on the filter button, applied some filters, and closed it => now I seem to remain on the old top 10 neighborhood without the filters (maybe?). I wonder if we should exist the top 10 tour if those buttons (on the sub-header) are clicked. What do you think?

Great catch, I cleaned this up and now not only does the top 10 tour reset appropriately but the map bounds adjust to the new filtered features as well:

https://github.com/user-attachments/assets/29b04a5d-41eb-47d6-813b-cdde23b08d4e

Thank you, @aaronxsu! Since this is approved and these last points addressed I go ahead and merge in on successful CI. For the sake of time/budget, if there is any followup let's open those in new cards to handle as part of later polish.

Issue #721, opened 2025-09-25
Add "Add new trip" modal


rachelekm:
Create "Add new trip" modal, following figma designs.

- Selecting "kind of place" option sets a new destination's purpose (note there can only be one selection per destination and purposes are not unique between user destinations)
- "Where is it" input should be address autocomplete. Create using react-aria autocomplete component (search/geocoding added in follow-up card, for now use hard-coded values from mapbox search API return example)
- "Add trip" button creates a new destination in user profile state and returns to trip wizard modal page on complete
- Content should be translatable

<img width="428" height="664" alt="Image" src="https://github.com/user-attachments/assets/e9fa30c4-1548-4a4f-a3aa-204151d3da60" />

PR #740, opened 2025-10-14
Add base Trips Wizard

### LINKED ISSUE #720: Create initial trip destination wizard page
Create initial trip destinations wizard page, following figma designs

- Step should appear following last step of user profile wizard on first login if empty destinations list in `userProfile` or on selecting edit (UPDATE: With addition of filters, can hold off on implementing latter until this we have trip filters modal)
- Display destinations as cards, if exist, with option to delete trip via "x"
- If no destinations, button to continue wizard is disabled. If destinations exist, button reads "finish"
- On finish user, update user profile:
       - If destination removed, update user profile and remove destination key entry in `networks.timesAndPathsData`
       - If destination added, update user profile and dispatch `getTimesAndPathsDataForPlace(newDestination)` (We don't want to refetch and parse this data if we don't have to, so only run for new destinations that aren't in `networks.timesAndPathsData`)
       - Set user `activeDestination` state as first destination of list on complete  
- Content in page should be translatable
- "Add another trip" button action not implemented in the card

Image

Image

rachelekm:
## Overview

Implements the Trips wizard as well as the Add Trips modal. Adds logic to add/remove destinations to user profile buffer (using sample location data).

### Checklist

- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/c0e79dae-21db-42f8-a417-982f833a7c48

## Testing Instructions

 * `scripts/server`
 * Login as a user with no destinations
 * Move through the user profile wizard to get to the last step
 * Confirm last step displays no destinations and the Finish button is disabled
 * Click button to add a new trip
 * Confirm Add Trip modal opens
 * Click outside modal or on the back button
 * Confirm taken back to trips wizard step and no changes to wizard progress
 * Click to add trips again
 * Confirm "Add trip" button disabled with no purpose selection or no address input
 * Select purpose option (confirm styling matches figma)
 * Begin to type address in search suggestion and use button to clear selection (confirm this works)
 * The sample autocomplete suggestions list from Mapbox includes "Michigan", type an address input with this search term
 * Confirm autocomplete dropdown opens
 * Select option and confirm input populated with full selection text (I left an inline comment here, but I had trouble with keyboard-accessible focusing for button/dropdown. For the sake of time left it as is and can open a tech-debt card)
 * Confirm on selection that "Add trip" button no enabled
 * Select Add Trip and confirm guided back to trip wizard step
 * Confirm new destination now added to step
 * Confirm "Finish" now enabled
 * Select Finish and confirm user profile successfully updates with new data (Specifically, creates new Destination)
 * Confirm paths and times data successfully fetched for all destinations (under-the-hood this is creating a destination using one sample coordinate, so they will all be the same "place")
 * Confirm rankings calculate as expected
 * Logout and re-login, confirm no longer directed to Profile page

Resolves #720 #721

### DIFF
--- a/src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
+++ b/src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
@@ -1,0 +1,99 @@
+import { useAsyncList, type AsyncListData } from "@react-stately/data";
+import {
+    Autocomplete as AriaAutocomplete,
+    Input,
+    Menu,
+    MenuItem,
+    SearchField,
+} from "react-aria-components";
+
+import TimesIcon from "assets/icons/times.svg?react";
+import Button from "components/base/Button/Button";
+import {
+    dropdownItemStyles,
+    dropdownPopoverStyles,
+} from "../Dropdown/Dropdown.styles";
+import autocompleteStyles from "./Autocomplete.styles";
+
+type Suggestion = { name: string; [key: string]: any };
+
+export function Autocomplete({
+    placeholder,
+    suggestions,
+    loadAsyncSuggestions,
+    onSuggestionCallback,
+}: {
+    placeholder: string;
+    suggestions?: Suggestion[];
+    loadAsyncSuggestions?: (
+        query: string,
+        signal: AbortSignal
+    ) => Promise<Suggestion[]>;
+    onSuggestionCallback?: (suggestion: Suggestion) => void;
+}) {
+    const { searchRoot, input: inputStyles } = autocompleteStyles();
+    const baseDropdownPopoverStyles = dropdownPopoverStyles();
+    const baseDropdownItemStyles = dropdownItemStyles();
+
+    let filteredSuggestions: AsyncListData<Suggestion> = useAsyncList({
+        async load({ signal, filterText }) {
+            if (loadAsyncSuggestions && filterText) {
+                const items = await loadAsyncSuggestions(filterText, signal);
+                return { items };
+            } else if (suggestions && filterText) {
+                const filtered = suggestions.filter(item =>
+                    item.name.toLowerCase().includes(filterText.toLowerCase())
+                );
+                return { items: filtered };
+            }
+            return { items: [] };
+        },
+        initialFilterText: "",
+    });
+
+    // Force close menu on suggestion selection
+    const menuIsOpen =
+        filteredSuggestions.items.length > 1 ||
+        (filteredSuggestions.items.length === 1 &&
+            filteredSuggestions.filterText !==
+                filteredSuggestions.items[0].name);
+
+    return (
+        <AriaAutocomplete
+            onInputChange={filteredSuggestions.setFilterText}
+            inputValue={filteredSuggestions.filterText}
+        >
+            <SearchField aria-label="Search" autoFocus className={searchRoot()}>
+                <Input placeholder={placeholder} className={inputStyles()} />
+                <Button
+                    variant="outline"
+                    size="small"
+                    leftIcon={
+                        <TimesIcon className="h-5 w-5 font-ligth fill-black" />
+                    }
+                    className="h-7 w-7 py-2 px-3 self-center mr-3"
+                />
+            </SearchField>
+            <Menu
+                items={menuIsOpen ? filteredSuggestions.items : []}
+                className={`${baseDropdownPopoverStyles} ${menuIsOpen ? "visible" : "invisible"}`}
+            >
+                {item => (
+                    <MenuItem
+                        id={item.name}
+                        onAction={() => {
+                            filteredSuggestions.setFilterText(item.name);
+                            if (onSuggestionCallback) {
+                                onSuggestionCallback(item);
+                            }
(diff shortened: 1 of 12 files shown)


src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+        },
+        initialFilterText: "",
+    });
+
+    // Force close menu on suggestion selection
+    const menuIsOpen =
+        filteredSuggestions.items.length > 1 ||
+        (filteredSuggestions.items.length === 1 &&
+            filteredSuggestions.filterText !==
+                filteredSuggestions.items[0].name);
```
rachelekm, line 66: I ran into lots of trouble getting the Autocomplete menu to open/close as expected as well as navigate from the keyboard. Specifically, the menu of suggestions is always open, even when no search input or after a selection. This felt confusing from a user perspective. This component is a lot more constrained, so this is the best hack I could do that seemed to work without spending too much time on it for our budget constraints.

rachelekm: I also could not get focus to work as expected, I don't have a fix for that here for the sake of time but will open a tech-debt card

src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+        signal: AbortSignal
+    ) => Promise<Suggestion[]>;
+    onSuggestionCallback?: (suggestion: Suggestion) => void;
+    onClearCallback?: () => void;
+}) {
+    const { searchRoot, input: inputStyles } = autocompleteStyles();
+    const baseDropdownPopoverStyles = dropdownPopoverStyles();
+    const baseDropdownItemStyles = dropdownItemStyles();
+
+    let filteredSuggestions: AsyncListData<Suggestion> = useAsyncList({
```
aaronxsu, line 45: This is an interesting first-use of `useAsyncList` in this project 👏🏼. I'd suggest adding some comments on what it does!

src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
aaronxsu: Got it! With the remaining time, I think this is fine, re: focus and conditions to show the menu. I'd summarize your comments here and turn them into code comment.

src/frontend/src/pages/Discover/Profile/AddTripModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+                    ...destination,
+                    purpose: purposeKey as PlaceKey,
+                });
+            } else {
+                setDestination({ ...destination, purpose: "" });
+            }
+        }
+    };
+
+    const purposesMap = Object.values(Place).map(value => ({
```
aaronxsu: Would it make sense to move it out of the component as this won't change per component props?

src/frontend/src/pages/Discover/Profile/StepTrips.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <Button
+                variant="outline"
+                size="large"
+                leftIcon={
+                    <PlusIcon className="h-[18px] w-[18px] fill-gray-600" />
+                }
+                className="justify-normal text-gray-700 text-lg font-bold"
+                onPress={() => setIsAddTripModalOpen(true)}
+            >
+                {t("userTrip.wizard.stepAddTrip.addATrip")}
```
aaronxsu: I'd consider conditionally display the copy here so that when there is no destination, it reads "Add a trip", and when there is more than 1, "Add another trip"

aaronxsu (commented):
This is looking great! I left some comments below. Additionally, when testing, I caught some edge cases that I'd like us to tackle:
1. As shown in the screenshot, the gap between the trip items seems larger than that in the design
2. The calculation of the primary destination seems a bit off as the two I added are both seen as primary. My suggestion: let's not assign primary in this modal step, but we take care of that in the finish button callback function: i.e. before submitting via the API, we mark the very first destination in the buffer as primary.
3. The `AddTripModal` state might need to be cleaned. For example: after I enter the first destination, click on "Add trip" button, it goes back to the main modal; then I click on "Add another trip", I am back to the add trip modal, but this time, the purpose is already selected (the same purpose as I selected in the previous trip) and the "Add trip" button is enabled (it should be disabled at this stage)
4. I'd suggest we update the highlight border color of the autocomplete text box. Currently it's yellow/orange. I'd probably use one of our theme colors or the blue color used when an element is highlighted.

What do you think?

Image

rachelekm:
Thanks, @aaronxsu! I fixed the above logic issues and added in that styling polish. This is all set for another look! Noting that I rebased with develop so you may need to force pull again.

aaronxsu (approved):
Thanks for addressing the comments and fixing the edge cases! This is now looking great and working very smoothly. Very nice work 👏🏼 

This is good to go after a changelog entry

PR #742, opened 2025-10-17
Add address autocomplete and handle invalid destinations

### LINKED ISSUE #608: Implement trip address autofill & geocoder
Implement address geocoder input of "Add new trip". The app previously used MapBox geocoding API with address autofill component, continue with this setup using the Conveyal Mapbox API key for this project.

- Implement Mapbox search API for autocomplete suggestions
- Implement Mapbox Geocoding API to forward geocode addresses and save trip coordinates to new destination

Image

rachelekm:
## Overview

Enables address search autocomplete following previous implementation, using Mapbox Geocoding API. I started first looking into the Mapbox Search Box API as an alternative but it didn't feel worth switching since we already have an implementation to refer to and know that immediate client needs fit within that API request free tier.

### Checklist

- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/b676e683-1806-4ec6-bfc0-a1e68ac87fdb

### Notes

The autocomplete suggestions are geographically constrained by the `BOUNDS` constant, however that bounding box still covers some area that we don't have indexes for in the network analyses on S3. On the current site, you can add a destination from anywhere, but it just shows no results. In our current implementation, we need all the path/time data for every destination upfront for commute ranges, grouped routes, and the compare page so if we have a destination with data it breaks the app. We didn't have any error handling in place for this scenario and its high impact, so I added some very basic guardrails in commit 6dfca888acfb67eb3de0ff0d08af58c6bcf3ed60.

In the event a user adds a destination we don't have analysis data for, a user is kicked back to the trips wizards (right now, the profile until we can implement editing trips). An error message will display and the invalid destination styling will be updated. A user cannot proceed to update their profile until the erroring destination is removed:
Screenshot 2025-10-16 at 6 58 41 PM

## Testing Instructions

 * `scripts/server`
 * Login as a user without any destinations
 * Move through the profile setup wizard and add a new trip
 * Begin to type in an address and confirm 1) Autocomplete populated with real addresses 2) Only Boston-area addresses are suggested
 * Select address option and confirm suggestions dropdown closes
 * Add trip and confirm destination updates in steps trip
 * Add more trips and confirm autocomplete updates as expected
 * Add a trip we know will fail (I've found the address "213 Scraggy Neck Rd" to be outside the bounds)
 * Select finish
 * Confirm only see loading screen for a while before see profile page again
 * Move through the wizard again and confirm trips step now displays error message
 * Remove erroring destinations and confirm message no longer displays and finish button now enabled
 * Select finish again and confirm now brought to discover page
 * Confirm destination markers on map match the points of the valid destinations added
 * Open up a couple neighborhoods and confirm routes generated for different destinations are as expected

Resolves #608

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
       - VITE_APP_GIT_COMMIT=${VITE_APP_GIT_COMMIT:-latest}
       - VITE_NETWORK_URL_ROOT=${VITE_NETWORK_URL_ROOT}
       - VITE_BHA_ECHO_PROGRAM_URL=${VITE_BHA_ECHO_PROGRAM_URL}
+      - VITE_MAPBOX_API_TOKEN=${MAPBOX_ACCESS_TOKEN}
     volumes:
       - ./src/frontend:/usr/local/src
       - $HOME/.aws:/root/.aws:ro
--- a/src/frontend/src/components/AddressAutocomplete.tsx
+++ b/src/frontend/src/components/AddressAutocomplete.tsx
@@ -1,0 +1,79 @@
+import { throttle } from "lodash";
+import { useCallback, useRef } from "react";
+import { useParams } from "react-router";
+
+import { getAddressSuggestions } from "api/addressSearch";
+import {
+    Autocomplete,
+    type Suggestion,
+} from "components/base/Autocomplete/Autocomplete";
+import type { Point } from "geojson";
+import { Language, type LanguageKey } from "src/enums";
+
+const RATE_LIMIT = 500;
+
+const AddressAutocomplete = ({
+    placeholder,
+    onClearCallback,
+    onSuggestionCallback,
+}: {
+    placeholder: string;
+    onClearCallback: () => void;
+    onSuggestionCallback: (suggestion: Suggestion) => void;
+}) => {
+    const { lang } = useParams<{ lang: LanguageKey | undefined }>();
+    const abortRef = useRef<AbortController | null>(null);
+    const cacheRef = useRef<Record<string, Suggestion[]>>({});
+
+    // When autocomplete is enabled in forward geocoding query,
+    // each user keystroke counts as one request to the Geocoding API.
+    // Implement strategies from old site to reduce requests.
+    // https://github.com/azavea/echo-locator/blob/master/taui/src/components/geocoder.js#L81-L105
+    const fetchSuggestions = useCallback(
+        async (query: string, signal: AbortSignal) => {
+            if (!query.trim()) return [];
+            // Return cached result if available
+            if (cacheRef.current[query]) {
+                return cacheRef.current[query];
+            }
+
+            // Abort previous request
+            if (abortRef.current) {
+                abortRef.current.abort();
+            }
+
+            const controller = new AbortController();
+            abortRef.current = controller;
+
+            const { features } = await getAddressSuggestions(
+                query,
+                signal,
+                lang ?? Language.EN
+            );
+            return features
+                .map(f => {
+                    if (!f.properties && f.geometry.type !== "Point")
+                        return null;
+                    return {
+                        name: f.properties?.full_address,
+                        geometry: f.geometry as Point,
+                    };
+                })
+                .filter(f => f !== null);
+        },
+        []
+    );
+
+    const throttledFetch = throttle(fetchSuggestions, RATE_LIMIT);
+
+    return (
+        <Autocomplete
+            placeholder={placeholder}
+            loadAsyncSuggestions={throttledFetch}
+            onClearCallback={onClearCallback}
+            onSuggestionCallback={onSuggestionCallback}
+        />
+    );
+};
+
+export default AddressAutocomplete;
--- a/src/frontend/src/pages/Discover/Profile/AddTripModal.tsx
+++ b/src/frontend/src/pages/Discover/Profile/AddTripModal.tsx
@@ -5,55 +5,12 @@ import { useTranslation } from "react-i18next";
 import type { Destination } from "reducers/userProfile/types";
 import { type PlaceKey } from "src/enums";
 
+import AddressAutocomplete from "components/AddressAutocomplete";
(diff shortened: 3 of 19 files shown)


src/frontend/src/reducers/networks/utils/fetchAndProcessDataByDestination.ts
```diff
@@ -58,7 +58,20 @@ export async function fetchAndProcessDataByDestination(
             const [pathsResponse, timesResponse] = await Promise.all([
                 pathsFetchPromise,
                 timesFetchPromise,
-            ]);
+            ]).catch(e => {
+                console.log(e);
```
aaronxsu: I'd change this to `console.error(e)`

src/frontend/src/pages/Discover/Profile/AddTripModal.tsx
```diff
@@ -99,33 +56,21 @@ const AddTripModal = ({
...
-
-        const location = {
-            label: label,
-            position: {
-                lat: EXAMPLE_COORDS_RES.latitude,
-                lon: EXAMPLE_COORDS_RES.longitude,
-            },
-        };
+    const handleLocationSelection = (selection: Suggestion) => {
+        if (selection.geometry) {
```
aaronxsu: What I would usually suggest for cases like this is a structure like the following for readability and reduced indentation. Up to you!
```
if (false) return;
// rest of logic goes here
```

docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
       - VITE_APP_GIT_COMMIT=${VITE_APP_GIT_COMMIT:-latest}
       - VITE_NETWORK_URL_ROOT=${VITE_NETWORK_URL_ROOT}
       - VITE_BHA_ECHO_PROGRAM_URL=${VITE_BHA_ECHO_PROGRAM_URL}
+      - VITE_MAPBOX_API_TOKEN=${MAPBOX_ACCESS_TOKEN}
```
aaronxsu, line 10: Just noting here that we would need to update the dotenv file for staging, and also it might be good to note this in the prod deployment issue, if it's not documented yet.

src/frontend/src/components/AddressAutocomplete.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+            // Abort previous request
+            if (abortRef.current) {
+                abortRef.current.abort();
+            }
+
+            const controller = new AbortController();
+            abortRef.current = controller;
+
+            return await getAddressSuggestions(
```
aaronxsu: I understand this block was likely adapted from the old code base, but did you get a chance to try the following pattern which follows closer to what we do now in this new version of the app?
```ts
const { data: { features } } = await getAddressSuggestions(...);
return features.map(...);
```

aaronxsu (commented):
Finished the first round of code review and left some comments. Could you please take a look at the failed CI?

docker-compose.yml
rachelekm: The entry for `MAPBOX_ACCESS_TOKEN` is already in the staging and prod `dotenv` file so we shouldn't need to update this one, but I did update the release to explicitly check/update values in the `dotenv`.

rachelekm:
Thanks, @aaronxsu! I addressed those initial comments as well as fixed the type error causing CI to fail. This is ready for another review! I also rebased with develop to knock out the merge conflicts so you might need to force pull

aaronxsu (changes requested):
Thanks for addressing the comments and fixing the CI errors. I got through the testing instructions and found the following cases:
1. When the checkbox is checked, the box border color changes, but the check mark does not show up. This might be a styling thing.

Image

2. After finishing the wizard with `213 Scraggy Neck Rd` as one of the destinations, I am back on the profile screen, which is expected. However, I think it might be helpful to update the wording on this page (similar to the error message you added on the trips step) and briefly state why I am still on the profile page.

3. In my testing, I set 213 Scraggy Neck Rd as the first destination, which submits as primary. However, after I proceed to remove it in the second round of the wizard and submit, I was not able to. I noticed the following error in the console.
```
Profile.tsx:72 Uncaught TypeError: Cannot assign to read only property 'primary' of object '#<Object>'
    at handleFinish (Profile.tsx:72:29)
```

Image

rachelekm:
Thanks, @aaronxsu, good catches!!

> 1. When the checkbox is checked, the box border color changes, but the check mark does not show up. This might be a styling thing.
> 

The default variant styling isn't working as I expected for checkboxes (I added a new variant as part of the previous filters work) so I explicitly set the variant style in [5b62123](https://github.com/azavea/echo-locator/pull/742/commits/5b62123baace196d8bdc07541d4510ce10a08626) and that seemed to do the trick!
Screenshot 2025-10-20 at 7 58 19 PM


> 2. After finishing the wizard with `213 Scraggy Neck Rd` as one of the destinations, I am back on the profile screen, which is expected. However, I think it might be helpful to update the wording on this page (similar to the error message you added on the trips step) and briefly state why I am still on the profile page.

I think the end goal would not to land back on the profile page but instead the edit trips wizard re-opens (as if you selected "edit your trips") and that messaging is right there. However I can't implement that until the edit work is in (I've got a draft PR #745 for it) and maybe for the sake of budget/less re-work we just stick with this flow back to profile anyways. I implemented your suggestion and added an additional message to the profile page in [0d7137e](https://github.com/azavea/echo-locator/pull/742/commits/0d7137e76deb17c8b91b8f4dbb49208fb4aec9ec), let me know what you think!


> 3. In my testing, I set 213 Scraggy Neck Rd as the first destination, which submits as primary. However, after I proceed to remove it in the second round of the wizard and submit, I was not able to. I noticed the following error in the console.

Nice catch! Luckily this was a quick fix (commit [2b0fb73](https://github.com/azavea/echo-locator/pull/742/commits/2b0fb73dda7355e72e22e9bee5f98bb8440e839b)) and this is now resolved.

aaronxsu (approved):
Thanks for addressing the issues. Good to go ~after a changelog entry~

Issue #739, opened 2025-10-13
Add ability to edit user profile


rachelekm:
Use the existing user profile wizard steps to update user profile. This includes opening the profile wizard from the filters modal as well as updating `hasVehicle` and `useCommuteRail` values in the edit trips modal. "Editing" user profile wizard should follow 3 steps in figma designs (but differs from initializing wizard in that is not followed by "Trips" wizard step). On finish update user profile in backend and update frontend `networks.activeMode` state.

On editing trips:
- If destination removed, update user profile and remove destination key entry in networks.timesAndPathsData
- If destination added, update user profile and dispatch getTimesAndPathsDataForPlace(newDestination) (We don't want to refetch and parse this data if we don't have to, so only run for new destinations that aren't in networks.timesAndPathsData)
- Set user activeDestination state as first destination of list on complete
- User cannot proceed with 0 destinations (should automatically be kicked back to Profile page)

PR #744, opened 2025-10-20
Add edit trip modal

### LINKED ISSUE #728: Add trips edit modal

rachelekm:
## Overview

Adds the trip editing modal for users to update:
1. active destination
2. the active mode used for routing an calculations, set by toggles for traffic conditions, is user has vehicle or is using commuter rail.

Currently "Edit your trips" does not open anything until #739 (any addition/removing of destinations expected to be handled in wizard).

### Checklist

- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/0ac92e8b-9f00-4ae5-a37f-652dc5386a45

## Testing Instructions

 * `scripts/server`
 * Login to django admin and confirm your user has multiple destinations
 * Login to app and confirm correct active destination (first in destinations list, if none selected) in redux state
 * Click on the edit trips button in the user profile subheader
 * Confirm edit trips modal opens
 * Confirm user profile state in redux is selected by default (eg. the active destination is expanded option in accordion, the correct traffic condition is toggled ("Peak" by default), the correct commute options are selected ("car" if user.hasVehicle and commuter checkbox selected if "user.useCommuterRail).
 * Select a variety of options and confirm user profile state updates correctly when closing modal
 * Confirm the `networks.activeMode` updates after fulfilled user profile with the correct option key (now taking into account "peak" and "off-peak")
 * Confirm ranking list re-generates
 * In neighborhood details confirm commute range and "Your Trips" maps updated to match user-selected options (e.g. car instead of transit)

Resolves #728

### DIFF
--- a/src/frontend/src/components/EditFiltersModal.tsx
+++ b/src/frontend/src/components/EditFiltersModal.tsx
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import {
     CheckboxGroup as AriaCheckboxGroup,
     Dialog,
@@ -57,19 +57,8 @@ const EditFiltersModal = ({ isMobile = false }) => {
     const modalOpen = useAppSelector(selectIsEditFiltersOpen);
     const bedrooms = useAppSelector(selectUserBedroomCount);
     const mode = useAppSelector(selectActiveMode);
-    const [filtersBuffer, setFiltersBuffer] = useState<FiltersState | null>(
-        null
-    );
-
     const filters = useAppSelector(selectNeighborhoodFilters);
-
-    useEffect(() => {
-        setFiltersBuffer({ ...filters });
-    }, [filters]);
-
-    if (!filtersBuffer) {
-        return <></>;
-    }
+    const [filtersBuffer, setFiltersBuffer] = useState<FiltersState>(filters);
 
     const regionsFilterSelected = (isSelected: boolean, region: string) => {
         const updatedRegionsFilter = [...filtersBuffer.regions];
--- a/src/frontend/src/components/EditTripsModal.tsx
+++ b/src/frontend/src/components/EditTripsModal.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import type { Key } from "react-aria-components";
 import { Dialog, Heading } from "react-aria-components";
 import { useTranslation } from "react-i18next";
@@ -23,12 +23,13 @@ import {
     ToggleButton,
     ToggleButtonGroup,
 } from "components/base/ToggleButton/ToggleButton";
-import { modalHeadingClassName } from "pages/Discover/ModalsWrapper";
 import { Accordion } from "./base/Accordion/Accordion";
 import { AccordionItem } from "./base/Accordion/AccordionItem";
 import CommuterRailCheckbox from "./CommuterRailCheckbox";
 import ModalCloseButton from "./ModalCloseButton";
 import TravelModeToggle from "./TravelModeToggle";
+
+const modalHeadingClassName = "text-lg font-bold text-gray-900";
 
 const EditTripsModal = () => {
     const { t } = useTranslation();
@@ -37,15 +38,7 @@ const EditTripsModal = () => {
     const profile = useAppSelector(selectUserProfile);
     const trafficConditions = useAppSelector(selectTrafficConditions);
     const [profileBuffer, setProfileBuffer] =
-        useState<UserProfileSliceState | null>(null);
-
-    useEffect(() => {
-        setProfileBuffer({ ...profile });
-    }, [profile]);
-
-    if (!profileBuffer) {
-        return <></>;
-    }
+        useState<UserProfileSliceState>(profile);
 
     // Update active destination
     const onSelectDestination = (keys: Iterable<string, void, undefined>) => {
@@ -182,8 +175,6 @@ const EditTripsModal = () => {
                         <h2 className={modalHeadingClassName}>
                             {t("editTripsModal.travelModeHeading")}
                         </h2>
-                        {/* Toggles in sync with logic in StepTravelMode
-                         */}
                         <TravelModeToggle
                             buffer={profileBuffer}
                             handleChange={onChangeHasVehicleToggle}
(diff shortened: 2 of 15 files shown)


src/frontend/src/pages/Discover/ModalsWrapper.tsx
```diff
@@ -1,6 +1,8 @@
 import EditFiltersModal from "src/components/EditFiltersModal";
-import NeighborhoodDetail from "../NeighborhoodDetail";
 import EditTripsModal from "src/components/EditTripsModal";
+import NeighborhoodDetail from "../NeighborhoodDetail";
+
+export const modalHeadingClassName = "text-lg font-bold text-gray-900";
```
aaronxsu: Wondering why this is defined here rather than where it's used. Lemme know if there is any concern I am missing here.

src/frontend/src/components/EditTripsModal.tsx
```diff
@@ -1,32 +1,198 @@
...
 
 const EditTripsModal = () => {
-    const dispatch = useDispatch();
+    const { t } = useTranslation();
+    const dispatch = useAppDispatch();
     const modalOpen = useAppSelector(selectIsEditTripsOpen);
+    const profile = useAppSelector(selectUserProfile);
+    const trafficConditions = useAppSelector(selectTrafficConditions);
+    const [profileBuffer, setProfileBuffer] =
+        useState<UserProfileSliceState | null>(null);
```
aaronxsu: For the buffer implementations in the recent PRs, I wonder, why not initializing the `profileBuffer` by `profile` directly. Did you run into performance issues or other state management issues?

src/frontend/src/components/EditTripsModal.tsx
```diff
@@ -1,32 +1,198 @@
...
+                            >
+                                {t("editTripsModal.offPeak")}
+                            </ToggleButton>
+                        </ToggleButtonGroup>
+                    </div>
+                    <div className="flex flex-col gap-4">
+                        <h2 className={modalHeadingClassName}>
+                            {t("editTripsModal.travelModeHeading")}
+                        </h2>
+                        {/* Toggles in sync with logic in StepTravelMode
```
aaronxsu: Is this a good thing or something we should change?

src/frontend/src/components/EditTripsModal.tsx
```diff
@@ -1,32 +1,198 @@
...
+                            {t("editTripsModal.dialogHeading")}
+                        </Heading>
+
+                        <p className="text-sm text-gray-600">
+                            {t("editTripsModal.dialogSubHeading")}
+                        </p>
+                        <ModalCloseButton onPress={() => onOpenChange(false)} />
+                    </div>
+                    <div className="flex flex-col gap-4">
+                        <Accordion
```
aaronxsu, line 113: I'd personally make these destination options as radio buttons that look like the vertical button group in the design, which would likely be closer to what it achieves semantically. You might have considered it too. But that would mean a new component to be created. Given the remaining time and budget, using an accordion works too.

aaronxsu (approved):
Looking great and working as advertised. I left some questions in the comments. I also found two small styling issues- you might have caught them as well, but just surfacing them here in case we forget

1. If the purpose is friends and family, in the mobile view, it looks like the following

Image

2. In the desktop view, when the address/label is long, it looks like the following

Image

src/frontend/src/pages/Discover/ModalsWrapper.tsx
rachelekm: I was sharing this with other modals but ultimately it's now only between 2 components and really not significant enough to warrant being pulled out. I took this out in [856b547](https://github.com/azavea/echo-locator/pull/744/commits/856b54751d57f9d22770e688a9af8590dfced79f)

src/frontend/src/components/EditTripsModal.tsx
rachelekm: Good call out, I was just moving fast and copying from the wizard. I updated this in [856b547](https://github.com/azavea/echo-locator/pull/744/commits/856b54751d57f9d22770e688a9af8590dfced79f) for both edit modals so its directly initialized and I didn't see any issues

src/frontend/src/components/EditTripsModal.tsx
rachelekm: I took out this comment since its confusing, just wanted a way to reference the clarifying comment in StepTravelMode

src/frontend/src/components/EditTripsModal.tsx
rachelekm: Yeah good to document this decision here, because I did think about that but given our budget/an already-styled accordion this felt like the best way forward. It'd be great to not only use that new radio group component here but also for the commute group component when a user's setup for car directions or a neighborhood is too far. I'll open up a tech debt issue for us and hopefully we can sneak it in at the end

rachelekm:
Thanks, @aaronxsu! I implemented your suggestions (all except for the accordion -> radio group). The edit button styling was partially addressed in a previous PR (https://github.com/azavea/echo-locator/pull/742/commits/8f0be27f75d4fd443656b7424195a470d35c5a28) but didn't fix the purpose overflow in mobile view. That's fixed in  [b70316f](https://github.com/azavea/echo-locator/pull/744/commits/b70316f6e0b5cafdc54b772b7571edd6726ad2d2).
Screenshot 2025-10-22 at 10 47 35 AM

Screenshot 2025-10-22 at 10 55 56 AM

I also rebased on develop and will go ahead and merge on successful CI.

src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -57,19 +57,8 @@ const EditFiltersModal = ({ isMobile = false }) => {
     const modalOpen = useAppSelector(selectIsEditFiltersOpen);
     const bedrooms = useAppSelector(selectUserBedroomCount);
     const mode = useAppSelector(selectActiveMode);
-    const [filtersBuffer, setFiltersBuffer] = useState<FiltersState | null>(
-        null
-    );
-
     const filters = useAppSelector(selectNeighborhoodFilters);
-
-    useEffect(() => {
```
aaronxsu, line 66: Great to see the initialization in the useState! Just noting here, that in some cases I worked on before, this useEffect might still be needed. We will test the workflows later when they are all in and then we could see if this is necessary or not. Even if we need to add it back, it should be quick and easy.

PR #745, opened 2025-10-20
Enable profile/trip wizards edit mode

### LINKED ISSUE #739: Add ability to edit user profile

rachelekm:
## Overview

Enables opening the profile and trips wizards separately for editing profile/destination information. Creates separate "user profile wizard" and "trips wizard" components to be opened on respective button calls through the app.

Also updates invalid destination handling so edits trip wizard re-opens on error.

### Checklist

- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/01a60f9a-8227-4fb1-a08e-3e08d8ac9ee8

## Testing Instructions

 * Login as a user without destinations
 * Move through profile setup wizard and confirm no regressions
 * Add an invalid destination as part of adding trips in last step (213 scraggy neck rd, for instance)
 * Confirm that edit trips wizard re-opens on invalid data for a user to fix
 * Remove invalid destination and confirm no regressions navigating to discover page
 * Open profile/filters edit modal and select button to edit user profile
 * Confirm wizard opens and only 3 steps in progress tracker
 * Confirm user profile updates on finishing modal and components' state updates as expected in app
 * Open edit trips modal and select button to "Edit your trips"
 * Confirm trips wizard opens without progress tracker (only 1 step)
 * Confirm can open modal to add a new trip
 * Add an invalid destination and a valid boston destination
 * Confirm on selecting "finish" that the trips edit wizard re-opens prompting user to remove invalid data
 * Remove invalid data and confirm user profile updates and new destinations have paths/times data
 * Confirm destination updates are reflected in components' state (in immediate trip accordion, in destination markers, in neighborhood details "your trips")
 * Click "edit your trips" button in neighborhood details "Your Trips" section
 * Confirm no issues adding/removing destinations in wizard from here

Resolves #739

### DIFF
--- a/src/frontend/src/components/EditWizard/EditWizard.tsx
+++ b/src/frontend/src/components/EditWizard/EditWizard.tsx
@@ -13,6 +13,7 @@ import Wizard from "components/Wizard/Wizard";
 import { updateUserProfile } from "reducers/userProfile/userProfileThunk";
 import { useAppDispatch, type RootState } from "store/store";
 
+import _ from "lodash";
 import type { BaseProps } from "../../pages/Discover/Profile/types";
 
 type EditWizardChildComponent<P = {}> = React.ReactElement<P & BaseProps>;
@@ -74,19 +75,20 @@ const EditWizard = ({
         if (!destinations.find(d => d.primary)) {
             destinations[0] = { ...destinations[0], primary: true };
         }
-        dispatch(
-            updateUserProfile({
-                ...profileBuffer,
-                voucherRooms: profileBuffer.rooms,
-                importanceAccessibility: parseInt(
-                    profileBuffer.importanceAccessibility
-                ),
-                importanceSchools: parseInt(profileBuffer.importanceSchools),
-                importanceViolentCrime: parseInt(
-                    profileBuffer.importanceViolentCrime
-                ),
-            })
-        );
+        const updatedProfile = {
+            ...profileBuffer,
+            voucherRooms: profileBuffer.rooms,
+            importanceAccessibility: parseInt(
+                profileBuffer.importanceAccessibility
+            ),
+            importanceSchools: parseInt(profileBuffer.importanceSchools),
+            importanceViolentCrime: parseInt(
+                profileBuffer.importanceViolentCrime
+            ),
+        };
+        if (!_.isEqual(updatedProfile, profile)) {
+            dispatch(updateUserProfile(updatedProfile));
+        }
         handleClose();
     };
 
--- a/src/frontend/src/components/EditWizard/EditWizard.tsx
+++ b/src/frontend/src/components/EditWizard/EditWizard.tsx
@@ -1,0 +1,127 @@
+import {
+    Children,
+    cloneElement,
+    isValidElement,
+    useEffect,
+    useState,
+    type ReactNode,
+} from "react";
+import { useTranslation } from "react-i18next";
+import { useSelector } from "react-redux";
+
+import Wizard from "components/Wizard/Wizard";
+import { updateUserProfile } from "reducers/userProfile/userProfileThunk";
+import { useAppDispatch, type RootState } from "store/store";
+
+import type { BaseProps } from "../../pages/Discover/Profile/types";
+
+type EditWizardChildComponent<P = {}> = React.ReactElement<P & BaseProps>;
+
+interface EditWizardProps {
+    isWizardOpen: boolean;
+    setWizardOpen: (isOpen: boolean) => void;
+    children: ReactNode;
+}
+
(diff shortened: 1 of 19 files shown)


src/frontend/src/reducers/userProfile/userProfileThunk.ts
```diff
@@ -9,5 +11,29 @@ export const getUserProfile = createAsyncThunk(
 
 export const updateUserProfile = createAsyncThunk(
     "userProfile/updateUserProfile",
-    async (profile: UserProfile) => await putUserProfile(profile)
+    async (profile: UserProfile, { dispatch, getState, rejectWithValue }) => {
+        const data = await putUserProfile(profile);
+        const state = getState() as RootState;
+        const appDispatch = dispatch as AppDispatch;
+        // Fetch times and paths data for any new destinations
```
aaronxsu, line 18: I am not super familiar with the algo, so I am wondering, for example, if I change something in my profile, not only just the destinations, e.g from 2 bedrooms to 3 bedrooms, will the times and paths data change? My understanding is that it should not, since times and path should only have things to do with location (i.e. destinations), but just wanted to double check.

src/frontend/src/reducers/userProfile/userProfileThunk.ts
```diff
@@ -9,5 +11,29 @@ export const getUserProfile = createAsyncThunk(
...
+        const newDestinations = [...data.destinations].filter(
```
aaronxsu: We could remove the spread here and filter on data.destinations directly, since filtering creates a new array and there is no side-effect in the filter function you defined here either.

src/frontend/src/components/EditWizard/EditWizard.tsx
```diff
@@ -0,0 +1,133 @@
...
+                    importanceViolentCrime: parseInt(
+                        profileBuffer.importanceViolentCrime
+                    ),
+                })
+            );
+        }
+        handleClose();
+    };
+
+    const childrenWithStepProps = Children.map(children, (child, index) => {
```
aaronxsu, line 98: I like how this is implemented! It's now abstracted only after we created all the steps and thus know all the required props. The inline comments are helpful too!

src/frontend/src/components/EditWizard/EditWizard.tsx
```diff
@@ -0,0 +1,133 @@
...
+    // redux store's user profile gets updated on fulfilled PUT in the reducer
+    const handleFinish = () => {
+        const destinations = profileBuffer.destinations;
+        // One destination must be primary, used to set
+        // userProfile.activeDestination to calculate score.
+        // Use first destination by default.
+        if (!destinations.find(d => d.primary)) {
+            destinations[0] = { ...destinations[0], primary: true };
+        }
+        if (!_.isEqual(profileBuffer, profile)) {
```
aaronxsu, line 78: I like this, to avoid excessive API calls and unnecessary re-calculations when there's nothing changed.

aaronxsu (approved):
This is working very smoothly. Very nice job getting all these wizard steps abstracted and orchestrated to work together with proper refactor. This is good to go after a changelog entry

src/frontend/src/reducers/userProfile/userProfileThunk.ts
rachelekm: Yup, that understanding is correct! The `timesAndRoutesData` state is every destination's travel time and path information and is unaffected by other user profile changes.

I initially followed our current conventions and dispatched the action to fetch new destination data via useEffect in the `Discover`. However, this was causing breaking changes since the `userProfile` thunk was fulfilling and updating selectors that rely on a destination's paths/times data before it actually finished fetching. I think ideally we'd clean up those useEffects and selectors so there's better guardrails, but including the data fetching as part of the user update thunk felt more controlled with the time/budget we have left.

rachelekm:
Thanks for the review!

Issue #757, opened 2025-10-24
Modals are not centered in Safari and iOS


philippschmitt:
The modals aren't centered on Safari on iOS and Mac. 

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/28225052-bde3-4c5a-ac56-9f0f997f6400" />

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/287c8bcb-001e-4cc6-a25f-502033038e1b" />

Issue #752, opened 2025-10-23
Text Search: Fix open dropdown list on switch from map view to list view


rachelekm:
The text search dropdown in automatically open if there's an initial value in either the map modal or list view. This dropdown should be closed on switching views:

https://github.com/user-attachments/assets/a17fac49-6368-4f92-aa99-ecf75d9b7a00

Issue #751, opened 2025-10-23
Text Search: Enable keyboard navigation for clear and dropdown suggestion selection


rachelekm:
We currently used react-aria's `ComboBox` (initially `Autocomplete`, which was more constrained) in combination with `SearchField` to create the address and neighborhood autocomplete components. Our implementation follows react-aria doc's example for [handling async loading of autocomplete suggestions](https://react-spectrum.adobe.com/react-aria/Autocomplete.html#async-loading). However, this example did not work as expected out of the box and we're not able to focus on the "clear" button or keyboard navigate autocomplete suggestions. The goal of this issue is to refactor our base autocomplete component so that it support keyboard clear and dropdown select.

rachelekm:
Closing in favor of duplicate #791

PR #743, opened 2025-10-20
Add neighborhood search

rachelekm:
## Overview

Adds ability for user to filter by neighborhoods. Search input available within modal (opened via search button control) on map page. On mobile, search input also available in list view.

Search include neighborhood name autocomplete. Neighborhood suggestions should already be filtered by regions/ecc options to prevent selecting an already-filtered neighborhood. On neighborhood selection, map bounds update to fit filtered list of neighborhoods. On clearing text search, map bounds reset to default bounds.

### Checklist

- [x] `fixup!` commits have been squashed
- [x] `CHANGELOG.md` updated with summary of features or fixes, following
      [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [ ] `README.md` updated if necessary to reflect the changes
- [x] Run `./scripts/format` to lint, format, and fix the application source code.
- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/c2b130c8-dd49-4fca-b08a-56c9fd9e0f55

## Testing Instructions UPDATED following filter refactor

 * `scripts/server`
 * Login and confirm no regressions displaying discover page top 10 tour, neighborhood bounds, and list
 * Confirm search button appears as top-right map control under zoom controls
 * Click search button and confirm search modal appears
 * Begin typing and confirm autocomplete list displays with all potential neighborhood suggestions
 * Make neighborhood selection and close modal
 * Confirm map zooms to neighborhood feature and is styled with outline
 * Click around different neighborhoods/open top 10 tour and confirm on exiting still brought back to text search feature
 * Re-open search and type for a neighborhood that shares a name with multiple features ("Cambridge")
 * Confirm similar UX to above but all neighborhood features with the shared name are styled
 * Re-open search and type for a neighborhood without making a neighborhood selection (e.g. "Camb" or "Boston")
 * Confirm similar UX to above and all neighborhood names with that substring are included in the bounds and styled
 * Confirm text search does not affect top 10 tour or recommendation coloring
 * Open list view and confirm list ONLY shows neighborhoods with the last text search selection
 * Confirm list search input in sync with map search modal input
 * Clear list search input and confirm list updates to only show neighborhoods that match text search
 * Clear search and confirm map and list reset
 * Open the filters modal and unselect a couple regions to filter
 * Close and confirm all neighborhoods are still clickable on the map but now recommendations/top 10 filtered to selected regions / ecc.
 * Confirm bounds fit to recommended neighborhoods
[...]

PR #766, opened 2025-10-30
Fix various modal & autocomplete UX issues

### LINKED ISSUE #750: Text Search: Enable modal close+search on key Enter
The text search modal on the map view currently only applies the search on modal close (either via close button or clicking outside). We should also enable a user to search by pressing Enter, closing the modal and applying the search.

rachelekm:
## Overview

1. Fixes all modal alignment so that centered on Safari and iOS

2. Fixes autocomplete UX bugs:
   - Dispatches text search on exiting modals vs explicitly clicking outside of input and then exiting modal
   - Prevents autocomplete dropdown suggestions from remaining open following selection or search between mobile map and list view displays. Broadly cleans up autocomplete dropdown open logic
   - Allows keyboard navigation to text search clear button Note: I still can't setup keyboard navigation for the dropdown options. The react-aria combobox component has a lot of internal logic that *should* setup navigation from the get-go but I think is getting thrown off since we're using it within modals  (under the hood it expects a popover ref for the dropdown) and that's overwriting my explicit keydown callbacks. I'm leaving this for now with our remaining budget and we have an issue open to address (#751 ) 

### Checklist

- [x] CI passes after rebase

### Demo

Modals centered in Safari and iOS:
Screenshot 2025-10-30 at 10 48 25 AM

Text search UX cleanup:

https://github.com/user-attachments/assets/604af3a0-6570-40fa-a807-f36c1135d1f0

## Testing Instructions

 * `scripts/server` and login
 * On mobile in the map display page, open up the text seach modal
 * Add some text input for a fuzzy search and exit the search modal
 * Confirm search performs as expected and navigate to list display
 * Confirm text search updated in list display search input and no dropdown is open
 * Clear text search and confirm no regressions resetting that filter
 * Input text search and make a neighborhood selection
 * Confirm the dropdown closes and the search performs as expected
 * Navigate to the map display and open the search modal
 * Confirm the dropdown is not open
 * Confirm no regressions with text search on desktop view
 * Open the app in a Safari browser
 * Confirm modals are centered as expected, specifically with login input modal, neighborhood details, and the search modal
 * Open the app in a Firefox browswer
 * Open responsive design mode and select different devices, specifically various iOS devices
 * Confirm modals are centered as expected in each device type

Resolves #750 #752 #757

### DIFF
--- a/src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
+++ b/src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
@@ -1,11 +1,11 @@
 import { useAsyncList, type AsyncListData } from "@react-stately/data";
 import type { Point } from "geojson";
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import {
     ComboBox as AriaCombobox,
     Input,
-    Menu,
-    MenuItem,
+    ListBox,
+    ListBoxItem,
     SearchField,
 } from "react-aria-components";
 
@@ -74,15 +74,9 @@ export function Autocomplete({
         initialFilterText: value ?? "",
     });
 
-    // Force close menu on suggestion selection
-    useEffect(() => {
-        setIsMenuOpen(
-            !!filteredSuggestions.items.length &&
-                !filteredSuggestions.items.find(
-                    i => i.name === filteredSuggestions.filterText
-                )
-        );
-    }, [filteredSuggestions.items, filteredSuggestions.filterText]);
+    const inputRef = useRef<HTMLInputElement>(null);
+    const listRef = useRef<HTMLDivElement>(null);
+    const clearButtonRef = useRef<HTMLButtonElement>(null);
 
     // Handles filter text set from other parent search,
     // to keep the map search and list search values in sync
@@ -91,7 +85,6 @@ export function Autocomplete({
     }, [value]);
 
     const handleTextSearch = () => {
-        setIsMenuOpen(false);
         if (filteredSuggestions.filterText) {
             const search = allowFuzzySearch
                 ? { name: filteredSuggestions.filterText }
@@ -119,6 +112,8 @@ export function Autocomplete({
                 inputValue={filteredSuggestions.filterText}
                 allowsCustomValue
                 aria-label="Search"
+                onOpenChange={isOpen => setIsMenuOpen(isOpen)}
+                onBlur={handleTextSearch}
             >
                 <SearchField
                     aria-label="Search Input"
@@ -130,11 +125,13 @@ export function Autocomplete({
                     <Input
                         placeholder={placeholder}
                         className={inputStyles()}
+                        ref={inputRef}
                     />
                     {filteredSuggestions.filterText && (
                         <Button
                             variant="outline"
                             size="small"
+                            ref={clearButtonRef}
                             leftIcon={
                                 <TimesIcon className="h-5 w-5 font-ligth fill-black" />
                             }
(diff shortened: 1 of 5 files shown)


src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
```diff
@@ -130,40 +130,51 @@ export function Autocomplete({
...
                             variant="outline"
                             size="small"
+                            ref={clearButtonRef}
                             leftIcon={
                                 <TimesIcon className="h-5 w-5 font-ligth fill-black" />
                             }
                             className="h-7 w-7 py-2 px-3 self-center mr-3"
                             aria-label="Clear search"
+                            onKeyDown={e => {
+                                if (e.key === "Enter") {
```
aaronxsu: I don't exactly recall, but I do wonder if these key strings are the same across different desktop OS.

aaronxsu (approved):
Tested and worked as advertised!

I noticed the following style thing, which we might consider fixing if time permits. In mobile neighborhood list view, the section below the text search seems to be pushed down when there are text search suggestions in the dropdown.

https://github.com/user-attachments/assets/dba055e0-bb86-41f8-8efe-301a77cb3fb3

src/frontend/src/components/base/Autocomplete/Autocomplete.tsx
rachelekm: Oh great point, it might be worth just taking these onKeyDown props out before merge. It only partially solves #751 and is likely to cause more confusion to have them exist than not, from a development side and also if it's working differently across devices.

aaronxsu: Sounds like a good plan to me

rachelekm:
> I noticed the following style thing, which we might consider fixing if time permits. In mobile neighborhood list view, the section below the text search seems to be pushed down when there are text search suggestions in the dropdown.

I saw this as well! tl;dr I'm struggling to work with this react-aria component and could see even a small style tweak spiraling. I even think it'd be better to have a larger refactor for it down that road vs continue to hack at it.

I'm pretty sure that style bug is because that dropdown isn't in an overlay or popover. I think this might be the same underlying cause preventing out-of-the-box keyboard navigation for this component. The initial react-spectrum `Autocomplete` component we were using did not have the ability to pass in props to allow fuzzy search. It's built with the the react-aria component `Combobox` under the hood (which does allow this customization) so I implemented that directly in #743. However, `Combobox` state expects references to a `Popover` component and a `Listbox` component for opening a dropdown and selecting options. I tried out quickly swapping those components in place of the `SearchField` and `Menu` (as well as applying the expected props from `Combobox` state to the component in place since we can [also import the context states](https://react-spectrum.adobe.com/react-aria/ComboBox.html#custom-children)) but then this caused some strange UI side effects for me like the inability to click on anything on that page. I think it's because of some inherit state their `Popover` component might have that we might need to work around, especially from within another modal. One quick alternative could be adding in the [Autocomplete component from HeroUI. ](https://www.heroui.com/docs/components/autocomplete)That lib is react-aria under-the-hood (and I'm pretty sure they also created tailwind variants?) and allows individual installation of components. But I'll add these thoughts into a future tech-debt issue.

aaronxsu:
Got it. Thanks for expanding on those trial and errors. In this case, I'da say this is a trade-off for us to deal with in a tech debt issue, as you suggested. For the sake of time, let's keep the existing implementation then!

rachelekm:
Thanks, @aaronxsu! I dropped the `onKeyDown` props in the latest rebase and added a changelog entry. I'll merge on successful CI build.

PR #794, opened 2026-02-26
Prevent users from unselecting all regions in filters

### LINKED ISSUE #787: Prevent unselecting all regions in filter modal
Add guardrails in filter component logic to always have at least one region selected

Image

rachelekm:
## Overview

Add guardrails to prevent users from unselecting all region options in the Profile modal.

### Checklist

- [x] CI passes after rebase

### Demo

https://github.com/user-attachments/assets/ca64b6b3-e47e-468d-a2c8-4dd70bd88800

## Testing Instructions

 * `scripts/server`
 * Login and open the Profile modal
 * Under the Filter Regions sections attempt to unselect all region checkboxes and confirm cannot unselect last
 * Exit modal and confirm region filter works as expected
 * Open modal again and select more regions, confirm can now unselect last checkbox
 * Confirm regions filter continues to work with variety of region selections / no scenario allows you to unselect all checkboxes

Resolves #787

### DIFF
--- a/src/frontend/src/components/EditFiltersModal.tsx
+++ b/src/frontend/src/components/EditFiltersModal.tsx
@@ -62,22 +62,21 @@ const EditFiltersModal = ({ isMobile = false }) => {
     const [filtersBuffer, setFiltersBuffer] = useState<FiltersState>(filters);
 
     const regionsFilterSelected = (isSelected: boolean, region: string) => {
-        const updatedRegionsFilter = [...filtersBuffer.regions];
-        if (isSelected) {
-            updatedRegionsFilter.push(region);
-        } else {
-            const regionIndex = updatedRegionsFilter.indexOf(region);
-            if (regionIndex > -1) {
-                updatedRegionsFilter.splice(regionIndex, 1);
+        setFiltersBuffer(prevBufferState => {
+            const updatedRegionsFilter = new Set(prevBufferState.regions);
+            if (isSelected) {
+                updatedRegionsFilter.add(region);
+            } else {
+                updatedRegionsFilter.delete(region);
             }
-        }
-        // Prevent unselecting all regions
-        if (updatedRegionsFilter.length === 0) {
-            return;
-        }
-        setFiltersBuffer({
-            ...filtersBuffer,
-            regions: updatedRegionsFilter,
+            // Prevent unselecting all regions
+            if (updatedRegionsFilter.size === 0) {
+                return prevBufferState;
+            }
+            return {
+                ...prevBufferState,
+                regions: [...updatedRegionsFilter],
+            };
         });
     };
 
@@ -176,6 +175,11 @@ const EditFiltersModal = ({ isMobile = false }) => {
                                     key={index}
                                     size="small"
                                     variant="ghost"
+                                    isDisabled={
+                                        // prevent un-selecting last region
+                                        filtersBuffer.regions.length === 1 &&
+                                        filtersBuffer.regions.includes(name)
+                                    }
                                 >
                                     <div className="flex flex-row gap-3">
                                         <div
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/EditFiltersModal.tsx
```diff
@@ -62,18 +62,24 @@ const EditFiltersModal = ({ isMobile = false }) => {
...
     const regionsFilterSelected = (isSelected: boolean, region: string) => {
-        const updatedRegionsFilter = [...filtersBuffer.regions];
-        if (isSelected) {
-            updatedRegionsFilter.push(region);
-        } else {
-            const regionIndex = updatedRegionsFilter.indexOf(region);
-            if (regionIndex > -1) {
-                updatedRegionsFilter.splice(regionIndex, 1);
+        setFiltersBuffer(prevBufferState => {
+            const updatedRegionsFilter = [...prevBufferState.regions];
```
aaronxsu: I'd consider using `Set` to simplify the logic and make this function more efficient. Consider trying this out and see if it works:
```ts
$ git diff
diff --git a/src/frontend/src/components/EditFiltersModal.tsx b/src/frontend/src/components/EditFiltersModal.tsx
index 612587d..94c57f3 100644
--- a/src/frontend/src/components/EditFiltersModal.tsx
+++ b/src/frontend/src/components/EditFiltersModal.tsx
@@ -63,22 +63,19 @@ const EditFiltersModal = ({ isMobile = false }) => {
 
     const regionsFilterSelected = (isSelected: boolean, region: string) => {
         setFiltersBuffer(prevBufferState => {
-            const updatedRegionsFilter = [...prevBufferState.regions];
-            if (isSelected && !updatedRegionsFilter.includes(region)) {
-                updatedRegionsFilter.push(region);
+            const updatedRegionsFilter = new Set(prevBufferState.regions);
+            if (isSelected) {
+                updatedRegionsFilter.add(region);
             } else {
-                const regionIndex = updatedRegionsFilter.indexOf(region);
-                if (regionIndex > -1) {
-                    updatedRegionsFilter.splice(regionIndex, 1);
-                }
+                updatedRegionsFilter.delete(region);
             }
             // Prevent unselecting all regions
-            if (updatedRegionsFilter.length === 0) {
+            if (updatedRegionsFilter.size === 0) {
                 return prevBufferState;
             }
             return {
                 ...prevBufferState,
-                regions: updatedRegionsFilter,
+                regions: [...updatedRegionsFilter],
             };
         });
     };
```

aaronxsu (commented):
I left a comment for a potential implementation improvement to consider.

Two other questions:
1. Should we also visually disable checkbox click by changing the cursor style if there is only one checkbox is active?
2. Should we update copy to convey that user needs to at least select a region?

I understand it's a 1 pt card. If either of them could make things too complex, we could defer- I'd like to suggest to try them out before we move on though.

src/frontend/src/components/EditFiltersModal.tsx
rachelekm: Nice cleanup! I implemented this in [a54f6c8](https://github.com/azavea/echo-locator/pull/794/commits/a54f6c83c19e33bbf0151e77c47829ad8c02ae31)

rachelekm:
> 1. Should we also visually disable checkbox click by changing the cursor style if there is only one checkbox is active?

That's a good idea, I disabled the checkbox if it's the last region selected. What do you think?

https://github.com/user-attachments/assets/b29eacfb-4b37-41ff-aa02-50d312693ea5


> 2. Should we update copy to convey that user needs to at least select a region?

For now I'll say let's pass updating the copy. Mostly since BHA already did a copy pass that we don't want to prolong before sending off to translations. Additionally it feels more intuitive now that a user is prevented from unselecting the last region.

This is ready for a second look!

aaronxsu (approved):
Working and looking great. Glad that you liked the `Set` solution. Good to go 👍🏼

PR #801, opened 2026-03-04
Add neighborhood csv submission instructions

### LINKED ISSUE #788: Investigate discrepancies in recommendation logic
The recommendation logic (ie the algorithm) is a bit of a black box in that we don't have a lot of context on the weighting applied to different criteria or understanding of the logic to apply scores/sort based off of it. There's a general gut feeling the algorithm was off even before site revision and with new neighborhood data + transit data + code migration its especially difficult to tell how to adjust it. We should try and understand the algorithm more / provide some constants before we start playing with the new weights. This issue is to:
1. Bring forward resulting neighborhood scores and their criteria-specific scoring for each neighborhood for 1-2 given destinations
       - These values are already produced as part of the algorithm we can print them to the console on run or gather via a script
2. Repeat for old version of site to compare
3. Identify any differences in criteria-based scores, transit-level scores, or resulting scores
4. Document process/findings
5. Decide on next steps to adjust weights (or code, as needed) with clients

rachelekm:
## Overview

Adds a README to the `neighborhood_data` dir for more specific neighborhood data template instructions. Intended to be used as reference for us as well as client when sending over csv template to prevent loss of context between data updates.

See https://github.com/azavea/echo-locator/issues/788#issuecomment-3999736958 for background.

### Checklist

- [x] CI passes after rebase

## Testing Instructions

 * Review .txt file and ensure grammatically correct and easy to read
 * Review table rows against expected column types and inline comments in `neighborhood_data/generate_neighborhood_json` script to confirm parity

Partially resolves #788


aaronxsu (commented):
I might be missing something. I have a few questions and a suggestion:

1. The `generate_neighborhood_json.py` script seems to expect some [image related columns](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/generate_neighborhood_json.py#L64-L66), but they are not in the readme. I wonder if those columns are added by `fetch_images.py` in the processing pipeline and are thus not required in the CSV. Is this the case?
2. The `generate_neighborhood_json.py` script removes x, y, lat, lon from the column list. The README documents lat and lon but not x and y. I presume they are added by `add_zcta_centroids.py`. I'd like to verify if this understanding is correct.
3. The `generate_neighborhood_json.py` script [casts type of `zipcode`](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/generate_neighborhood_json.py#L101) to string, but the readme has this field as `Number`. I wonder if this could be that Boston zip code can be like `02108` and thus we needed it to be `str` so that the leading 0 won't get lost. If this is the case, then we might want to change the data type from `Number` to `Text` for `zipcode` in the readme.
4. Suggestion: the `README.md` at the root of the repo has a great section for Data. I'd like to suggest we add a sentence to point to this new readme text file that this should be referenced as the CSV schema. What do you think?

rachelekm:
Thanks for the review, @aaronxsu! Answering your questions below:
> 
> 1. The `generate_neighborhood_json.py` script seems to expect some [image related columns](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/generate_neighborhood_json.py#L64-L66), but they are not in the readme. I wonder if those columns are added by `fetch_images.py` in the processing pipeline and are thus not required in the CSV. Is this the case?

Yes, those extra image metadata columns are generated in `fetch_images`, [here](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/fetch_images.py#L38-L46), and written to an intermediate output file `neighborhood_centroids_descriptions.csv`. This file is then read and processed in the `generate_neighborhood_json` script. So those metadata columns are expected, but not in the initial CSV and only following `fetch_images`.


> 2. The `generate_neighborhood_json.py` script removes x, y, lat, lon from the column list. The README documents lat and lon but not x and y. I presume they are added by `add_zcta_centroids.py`. I'd like to verify if this understanding is correct.

Yup similar to the above, `x` and ``y are added by the `add_zcta_centroids` script to an enriched, intermediate output file, `neighborhood_centroids_descriptions.csv`, [here](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/add_zcta_centroids.py#L80-L81). The `x` and `y` values are just used to build a `Neighborhood` geometry coordinates so they're tossed once used in `generate_neighborhood_json` along with `lon and `lat`.

> 3. The `generate_neighborhood_json.py` script [casts type of `zipcode`](https://github.com/azavea/echo-locator/blob/develop/neighborhood_data/generate_neighborhood_json.py#L101) to string, but the readme has this field as `Number`. I wonder if this could be that Boston zip code can be like `02108` and thus we needed it to be `str` so that the leading 0 won't get lost. If this is the case, then we might want to change the data type from `Number` to `Text` for `zipcode` in the readme.

That's a good point, number felt intuitive for the BHA-side to interpret but we do want to preserve those starting `0` values (and it gets cast to a string anyways). I updated the README value type for zipcode in [25b8e3f](https://github.com/azavea/echo-locator/pull/801/commits/25b8e3f63915148b9addfee832e901106c0f60c3).

> 4. Suggestion: the `README.md` at the root of the repo has a great section for Data. I'd like to suggest we add a sentence to point to this new readme text file that this should be referenced as the CSV schema. What do you think?

That makes sense! Just added that context in[ 25b8e3f](https://github.com/azavea/echo-locator/pull/801/commits/25b8e3f63915148b9addfee832e901106c0f60c3), let me know what you think!

aaronxsu (approved):
Thanks for the replies to my questions and for the changes. They make sense to me.

This PR looks good and is good to go!

rachelekm:
Thanks for the review!

PR #806, opened 2026-03-06
Allow un-routable neighborhoods to be recommended

### LINKED ISSUE #805: Allow un-routable neighborhoods to be recommended
Recommended neighborhoods have been exclusively routable neighborhoods, but this is not intuitive looking at the map and in more nuanced travel scenarios. We want to update the sorting selector so now a neighborhood with a 2+ hour commute or walkable can also be recommended. This decision is following issue #788, the investigation and client pairing to improve the algorithm. Specifically, see [comment here](https://github.com/azavea/echo-locator/issues/788#issuecomment-4014316420).
This issue is to:
1. Update the recommended neighborhoods sorting selector so that all neighborhoods are added to the recommended list
      - Confirm the above update correctly colors neighborhood bounds on the map and the list is as expected
      - Confirm filters still move neighborhoods to "not a match" if applicable
      - Confirm no regressions elsewhere, ie ranked sorting within recommendations still works, top ten still works
2. Take note of any missing neighborhood bounds data via screenshots to send to client

Any small copy edits from the client pairing session will be addressed as part of translations work.

rachelekm:
## Overview

We are no longer using a neighborhoods "routability" as a constraint to be recommended. A neighborhood that has a 2+ hour commute by car or transit can now be recommended as well as neighborhoods that may not have transit routing but are close enough to walk.

- All neighborhoods are now added to the recommended list by default and still sorted by score
- If a user select any filter settings, ECC or region, then neighborhoods that don't fit that configuration are moved to the "not a match" list
- There are no changes to the legend text and the "not a match" list description will be updated in follow up work.

### Checklist

- [x] CI passes after rebase

### Demo
Neighborhood close to destination but previously un-routable and un-recommended. With updates now recommended:
Screenshot 2026-03-06 at 5 58 34 PM

## Testing Instructions

 * `scripts/server`
 * Login and set travel configurations to transit
 * Set destination to 700 Boylston St
 * Confirm all neighborhoods are now recommended
 * Confirm Boston - Back Bay one of the top neighborhoods (if commute time is very important)
 * Adjust importance settings and confirm neighborhoods list updates
 * Adjust variable filters and confirm neighborhoods appropriately move to the "not a match" list

Resolves #805

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Refactor networks data state to support place comparisons [#699](https://github.com/azavea/echo-locator/pull/699)
 - Refactor user authentication APIs [#718](https://github.com/azavea/echo-locator/pull/718)
 - Implement units design changes [#776](https://github.com/azavea/echo-locator/pull/776)
+- Allow un-routable neighborhoods to be recommended [#806](https://github.com/azavea/echo-locator/pull/806)
 
 ### Fixed
 
--- a/src/frontend/src/reducers/neighborhoods/selectors/neighborhoodsSortedWithRoutes.ts
+++ b/src/frontend/src/reducers/neighborhoods/selectors/neighborhoodsSortedWithRoutes.ts
@@ -10,7 +10,6 @@ import pullAt from "lodash/pullAt";
 import {
     BOOST_DOWNTOWN_RESULT_PLACE,
     DOWNTOWN_AREAS,
-    MAX_TRAVEL_TIME,
     RESULTS_WITH_DOWNTOWN,
 } from "../../../constants";
 import { createSelector } from "@reduxjs/toolkit";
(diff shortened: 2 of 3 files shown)


aaronxsu (approved):
This looks good and is working as advertised.

Unrelated to this PR and might not be the case on staging or prod in the future: I noticed some neighborhood images' URLs might be outdated locally. We might be running the scripts again on staging and production so this will not be relevant? Or maybe we create a tech debt card to fix this so that if the URL is broken, we show a placeholder etc? Anyway, this does not block merging this PR- just putting this on your radar!

Image

rachelekm:
Thanks, @aaronxsu!

> Unrelated to this PR and might not be the case on staging or prod in the future: I noticed some neighborhood images' URLs might be outdated locally. We might be running the scripts again on staging and production so this will not be relevant? Or maybe we create a tech debt card to fix this so that if the URL is broken, we show a placeholder etc? Anyway, this does not block merging this PR- just putting this on your radar!

Good catch, I didn't see this! I won't open up an issue for it for now since it looks like this example is working on staging (screenshot below) and we will need to re-run data updates soon anyways. I wonder if these URLs are referencing the old thumbnail-sized images that are no longer on S3? We created new full-sized images as part of the frontend rewrite and any vestigial thumbnails should've been cleaned up in latest dev data so hopefully this just resolves once we update with new data.



Screenshot 2026-03-11 at 3 24 14 PM

